PR #2, opened 2020-04-16
Set up a default Node.js CI GitHub action.

axmb:
This is the default nodejs.yml config, with some small tweaks/customization for our monorepo. Should run on sending a PR and on pushing to `master`.

Can always tweak this in the future, but should prevent most breakages for now.


axmb:
Merge is totally goofed (PEBKAC). Closing/creating a separate PR.

PR #3, opened 2020-04-16
Setting up a base GitHub action for our Node apps.

axmb:
This is the default nodejs.yml config, with some small tweaks/customization for our monorepo. Should run on sending a PR and on pushing to `master`.

Can always tweak this in the future, but should prevent most breakages for now.


axmb:
[Set up some basic CI in our monorepo](https://trello.com/c/zvna0QXW/61-set-up-some-basic-ci-in-our-monorepo)

PR #5, opened 2020-04-17
Configure linting (eslint) and integrate prettier for auto-formatting.

axmb:
Getting the linter/formatter ready to rock. Not applying in this PR; don't want to confuse the issue; I'll apply fixes separately.

Mostly just using the baseline recommended config for TS. Made a couple judgement calls on some common rules. Feel free to turn this thread into a holy war on quotes/tabs/etc.

### DIFF
--- a/data-serving/data-service/.prettierrc.js
+++ b/data-serving/data-service/.prettierrc.js
@@ -1,0 +1,7 @@
+module.exports = {
+  semi: true,
+  trailingComma: "all",
+  singleQuote: true,
+  printWidth: 120,
+  tabWidth: 4
+};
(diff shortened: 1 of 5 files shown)


data-serving/data-service/.prettierrc.js
```diff
@@ -0,0 +1,7 @@
+module.exports = {
+  semi: true,
+  trailingComma: "all",
+  singleQuote: true,
+  printWidth: 120,
```
attwad, line 5: but, but, our dear and treasured 80 col limit??!!?

https://google.github.io/styleguide/jsguide.html#formatting-column-limit

just kidding, 120 lgtm!

axmb: Ah, good catch! I thought I ought to change this, but forgot to circle back.

Prettier recommends 80, too -- so we're officially correct.

https://prettier.io/docs/en/options.html#print-width

PR #11, opened 2020-04-20
Add config api routes and chose a different default port for curation service

attwad:
…with data service service

### DIFF
--- a/verification/curator-service/src/controllers/sources.ts
+++ b/verification/curator-service/src/controllers/sources.ts
@@ -1,0 +1,41 @@
+import { Request, Response } from 'express';
+
+/**
+ * List the sources.
+ */
+export const list = (req: Request, res: Response) => {
+    res.sendStatus(501);
+    return;
+};
+
+/**
+ * Get a single source.
+ */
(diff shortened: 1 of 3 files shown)


attwad:
Ah damn, I forgot to push before submitting my other pending changes (I should have created another branch...), do you mind reviewing this again so I can merge both? Thank you

verification/curator-service/src/controllers/sources.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+export const create = (req: Request, res: Response) => {
+    res.sendStatus(501);
+    return;
+};
+
+/**
+ * Delete a single source.
+ */
+export const del = (req: Request, res: Response) => {
```
axmb, line 38: I like the naming convention for controllers (just the method name, since we're importing the entire module). but hyper-nit here: can we just spell out delete (it's the same length as update)?

attwad: No:
'delete' cannot be called on an identifier in strict mode.ts(1102)
it's an actual keyword :(

attwad:

PR #12, opened 2020-04-20
Base create method for case data

axmb:
Sets up body parsing and express validation, so separating this into an initial PR from the remainder of the mutation methods.


axmb:
[Define API to CRUD MongoDB case data collection](https://trello.com/c/k7yBDBGI/46-define-api-to-crud-mongodb-case-data-collection)

PR #15, opened 2020-04-20
Wire up dataserver to MongoDB instance

khmoran:
Not ready for submit -- wanted to start a discussion on how we want to organize things and specify values for different rigs/environments

### DIFF
--- a/data-serving/data-service/README.md
+++ b/data-serving/data-service/README.md
@@ -1,0 +1,8 @@
+# Starting a dev server
+1. Run a local MongoDB instance
+   `mongod`
+2. Run the local dataserver
+   `npm run dev`
+
+# Import data into local MongoDB instance
+`mongoimport --db dev --collection covid19 --file data-serving/samples/line-list.json`
--- a/data-serving/data-service/README.md
+++ b/data-serving/data-service/README.md
@@ -5,4 +5,5 @@
    `npm run dev`
 
 # Import data into local MongoDB instance
-`mongoimport --db dev --collection covid19 --file data-serving/samples/line-list.json`
+From the data-service root, run:
+`npm run import-data`
--- a/data-serving/data-service/src/common/db-client.ts
+++ b/data-serving/data-service/src/common/db-client.ts
@@ -2,7 +2,7 @@ import { MongoClient, Db } from 'mongodb';
 
 /** A thin wrapper around [[`MongoClient`]]. */
 export default class DbClient {
-    client: MongoClient;
+    private client: MongoClient;
 
     constructor(connectionString: string) {
         this.client = new MongoClient(connectionString, {
--- a/data-serving/data-service/src/common/db-client.ts
+++ b/data-serving/data-service/src/common/db-client.ts
@@ -1,20 +1,24 @@
-import { MongoClient } from 'mongodb';
+import { MongoClient, Db } from 'mongodb';
 
-class DbClient {
+/** A thin wrapper around [[`MongoClient`]]. */
+export default class DbClient {
     client: MongoClient;
 
-    constructor() {
-        this.client = new MongoClient('mongodb://localhost:27017', { useUnifiedTopology: true });
+    constructor(connectionString: string) {
+        this.client = new MongoClient(connectionString, {
+            useUnifiedTopology: true,
+        });
     }
 
     public async connect() {
         await this.client.connect();
-        return this.client;
+    }
+
+    public db(dbName: string): Db {
+        return this.client.db(dbName);
     }
 
     public disconnect() {
         this.client.close();
     }
 }
-
-export = new DbClient();
--- a/data-serving/data-service/src/common/db-client.ts
+++ b/data-serving/data-service/src/common/db-client.ts
@@ -10,15 +10,15 @@ export default class DbClient {
         });
     }
 
-    public async connect() {
+    async connect() {
         await this.client.connect();
     }
 
-    public db(dbName: string): Db {
+    db(dbName: string): Db {
         return this.client.db(dbName);
     }
 
-    public disconnect() {
+    disconnect() {
         this.client.close();
     }
 }
--- a/data-serving/data-service/src/common/db-client.ts
+++ b/data-serving/data-service/src/common/db-client.ts
@@ -1,0 +1,20 @@
+import { MongoClient } from 'mongodb';
+
+class DbClient {
+    client: MongoClient;
+
+    constructor() {
+        this.client = new MongoClient('mongodb://localhost:27017', { useUnifiedTopology: true });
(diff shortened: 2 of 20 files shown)


data-serving/data-service/src/common/db-client.ts
```diff
@@ -0,0 +1,20 @@
+import { MongoClient } from 'mongodb';
+
+class DbClient {
+    client: MongoClient;
+
+    constructor() {
+        this.client = new MongoClient('mongodb://localhost:27017', { useUnifiedTopology: true });
```
attwad, line 7: is 27017 a hardcoded mongo thing?

attwad: usually you'd pass the mongo url (which could also contain username and password) in an env variable.

data-serving/data-service/src/server.ts
```diff
@@ -1,9 +1,22 @@
 import app from './index';
+import DbClient from './common/db-client';
+
 
 // Start Express server.
-const server = app.listen(app.get('port'), () => {
+const server = app.listen(app.get('port'), async () => {
     console.log('App running at http://localhost:%d.', app.get('port'));
```
attwad, line 7: localhost:%d is bound to become stale once we start exporting that server outside, can we stick to "listening on port xxx" instead?

axmb:
The canonical solution to this appears to be the node `dotenv` [package](https://www.npmjs.com/package/dotenv).

If we think that LG, while you do merge things, I can send a change setting this up, and you can place the initial DB URI (et al.) in there.

data-serving/data-service/src/common/db-client.ts
khmoran: yep, hardcoding this here is temporary -- i'll use the dotenv package once ab merges it

data-serving/data-service/src/server.ts
khmoran: yeah this is outside the scope of this change but happy to update in a followup for sure

data-serving/data-service/README.md
```diff
@@ -0,0 +1,8 @@
+# Starting a dev server
+1. Run a local MongoDB instance
+   `mongod`
```
axmb, line 3: Perhaps add a `npm script` for this to `package.json`?

E.g.,

`local-db: mongod && mongoimport --db dev --collection covid19 --file data-serving/samples/line-list.json`

data-serving/data-service/src/common/db-client.ts
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    public async connect() {
```
axmb, line 10: Does `npm run lint` get mad about these not having types?

data-serving/data-service/src/common/db-client.ts
```diff
@@ -0,0 +1,20 @@
+import { MongoClient } from 'mongodb';
```
axmb, line 1: This is the right place to start. We should probably take a look at [mongoose](https://www.npmjs.com/package/mongoose), which is a super popular object-management library built on top of (replacing) mongodb. I haven't looked at the trade-offs, now that native MongoDB validators are a thing, but I'd be interested in your thoughts.

data-serving/data-service/src/common/db-client.ts
khmoran: weirdly no but some types aren't a bad idea -- now only the new db() function will return something so added a type here

data-serving/data-service/src/common/db-client.ts
khmoran: yeah I started looking into it and found opinions on both sides so I decided to defer on that decision, but definitely agree that we need to look into it

(here's one piece I started reading earlier: https://medium.com/better-programming/mongoose-vs-the-mongodb-node-driver-7dc59f00a5dd)

data-serving/data-service/README.md
khmoran: added one for importing the data, but mongod stays running in the fg... is there a nice way to script: (1) starting mongod, (2) in a separate process, running mongoimport, (3) killing mongod? or backgrounding mongod & then killing after

data-serving/data-service/src/server.ts
attwad: Sure fine to do in another change, thanks :)

data-serving/data-service/src/server.ts
```diff
@@ -1,9 +1,39 @@
...
+        console.log('  Connected to the database!');
+
+        const db = client.db(process.env.DB_NAME);
+        const docs = await db
+            .collection(process.env.DB_COLLECTION)
+            .find()
+            .toArray();
+        console.log(`    Found ${docs.length} records`);
+    } catch (e) {
+        console.log('  Failed to connect to the database. :(', e);
```
attwad, line 35: use console.error

khmoran: oops thanks!

data-serving/data-service/src/server.ts
khmoran: ah I see you were really talking about just changing the message here instead of using an env variable -- just doing in this PR since it's easy peasy. Thanks!

data-serving/data-service/src/common/db-client.ts
```diff
@@ -0,0 +1,24 @@
...
+export default class DbClient {
+    client: MongoClient;
+
+    constructor(connectionString: string) {
+        this.client = new MongoClient(connectionString, {
+            useUnifiedTopology: true,
+        });
+    }
+
+    public async connect() {
```
axmb, line 13: Uh oh. Time to decide if we'll use explicit `public` (since it's default in TS). I vote implicit!

data-serving/data-service/src/controllers/case.ts
```diff
@@ -40,3 +40,21 @@ export const create = async (req: Request, res: Response) => {
 
     res.send(`Triggered create case with age: ${req.body.age}`);
 };
+
+/**
```
axmb, line 44: Er, merge looks a little weird?

data-serving/data-service/README.md
khmoran: sg, thanks!

data-serving/data-service/src/common/db-client.ts
khmoran: gtk, agreed we should omit if it's default

data-serving/data-service/src/controllers/case.ts
khmoran: I DON'T EVEN KNOW WHY THIS FILE IS EDITED PLEASE SEND HELP

PR #16, opened 2020-04-20
Adding base update/delete methods to case data API.

axmb:


axmb:
[Define API to CRUD MongoDB case data collection](https://trello.com/c/k7yBDBGI/46-define-api-to-crud-mongodb-case-data-collection)

PR #17, opened 2020-04-20
dotenv set up for both services

axmb:
Documentation [here](https://www.npmjs.com/package/dotenv). By convention, `.env` contains secrets and is .gitignored/strongly discouraged from being uploaded. We could share a copy in drive for development, perhaps.

As an example:

```
alex@anblox:~/.../data-serving/data-service$ cat .env && echo
PORT=3002
alex@anblox:~/.../data-serving/data-service$ npm run start

> data-service@1.0.0 start [...]/data-serving/data-service
> node dist/server.js

App running at http://localhost:3002.
  Press CTRL-C to stop
```

### DIFF
--- a/verification/curator-service/src/index.ts
+++ b/verification/curator-service/src/index.ts
@@ -1,10 +1,13 @@
 import express from 'express';
+import dotenv from 'dotenv';
 
 // Controllers (route handlers).
 import * as homeController from './controllers/home';
 import * as sourcesController from './controllers/sources';
 
 const app = express();
+
+dotenv.config();
 
 // Express configuration.
 app.set('port', process.env.PORT || 3001);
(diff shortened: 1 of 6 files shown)


verification/curator-service/src/index.ts
```diff
@@ -1,11 +1,14 @@
...
 // Controllers (route handlers).
 import * as homeController from './controllers/home';
 import * as sourcesController from './controllers/sources';
 
 const app = express();
 
+dotenv.config();
+
 // Express configuration.
 app.set('port', process.env.PORT || 3001);
```
attwad, line 13: should we change that?

axmb: This line was forwards-compatible. This is still what you'd want using dotenv.

attwad: Oh cool :)

axmb:
> nice. should we add some documentation on this so folks know it's a requirement/what fields they need to specify?
> 
> (mostly for people outside core dev who, as you said, can share a config privately)

I think there's a larger task here to organize service-/repo-level documentation; there are a handful of pieces that ought to be documented. I'll make that task/assign myself.

PR #21, opened 2020-04-21
added json schema validator for curator sources db

attwad:


attwad:
I'm dropping this in favor of mongoose schema validation for now, we can revisit this later if needed.

PR #22, opened 2020-04-21
Validate env with envalid

axmb:


data-serving/data-service/src/util/validate-env.ts
```diff
@@ -1,7 +1,16 @@
...
+    PORT: number;
+}> &
+    CleanEnv & {
+        readonly [varName: string]: string | undefined;
+        // eslint-disable-next-line indent
+    } {
+    return cleanEnv(process.env, {
         DB_CONNECTION_STRING: str({
             desc: 'MongoDB URI provided to MongoClient.',
             devDefault: 'mongodb://localhost/',
```
khmoran, line 16: do we want the mongod port here? mongodb://localhost:27017

axmb: It's an optional url part, [and defaults to 27017](https://mongodb.github.io/node-mongodb-native/driver-articles/mongoclient.html#basic-parts-of-the-url). Doesn't hurt to specify, though -- I'll add.

PR #23, opened 2020-04-21
plug mongodb into curator service, implement list

attwad:
Tested on a local mongo db, going to /api/sources/ gave me:

`[{"_id":"5e9f0e9cccd99c8b1401fec0","name":"test-source"}]`

which I am really proud of.

I am not really proud of how the tests are setup though, I just don't understand jest/supertest/whatever we're doing here enough to understand how to keep the `MemStore` into the test and not have it leaked into the index.ts. Any help here will be greatly appreciated.

@Kelly I did things a bit differently that how you did in the dataserver (using different implementations of a standard Store interface (mongo or inmemory)), I feel it's cleaner but it might not be "typescripty", I am not sure. Comments appreciated.

### DIFF
--- a/verification/curator-service/src/storage/store.ts
+++ b/verification/curator-service/src/storage/store.ts
@@ -1,0 +1,26 @@
+import { Db } from 'mongodb';
+
+const SOURCES_COLLECTION = 'sources';
+
+export interface Store {
+    listSources(): Promise<Array<object>>;
+}
+
+export class MongoStore implements Store {
+    constructor(private readonly db: Db) {}
+    async listSources(): Promise<Array<object>> {
+        const col = this.db.collection(SOURCES_COLLECTION);
+        return await col.find().toArray();
+    }
+}
+
+export class MemStore implements Store {
+    sources: Array<object> = [];
+
+    async listSources(): Promise<Array<object>> {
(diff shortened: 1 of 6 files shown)


verification/curator-service/src/storage/store.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export class MongoStore implements Store {
+    constructor(private readonly db: Db) {}
+    async listSources(): Promise<Array<object>> {
+        const col = this.db.collection(SOURCES_COLLECTION);
+        return await col.find().toArray();
+    }
+}
+
+export class MemStore implements Store {
```
axmb, line 17: For testing, there's a jest-mongodb package that simplifies directing MongoDB to an in-memory server ([jest docs](https://jestjs.io/docs/en/mongodb)). I'm not sure this set up is worth having to conditionally configure the implementation in our app, and writing the in-memory methods for each method/source. Mongoose offers this same object-orientation; we'd just operate on the model schemas in lieu of defining .*Store classes ([docs](https://www.npmjs.com/package/mongoose#accessing-a-model)).

My impression is that the default jest/mongoose set up will be simpler, but I'm interested in your take on the above.

PR #24, opened 2020-04-21
Implement an object-managed connection to MongoDB in data-service; use it in case API; facilitate testing

axmb:
Here's an approach for us to consider re: managed app access to MongoDB and testability.

Stuff to look at:

- Slightly simpler connection set up in `index.ts` with [mongoose](https://www.npmjs.com/package/mongoose).
- Validated data model in `model/case.ts`.
- With mongoose magic, this gives us an object oriented MongoDB API for each model defined. Implemented create/list in `controllers/case.ts`.
- Testing is mostly unchanged. Using [jest-mongodb](https://www.npmjs.com/package/@shelf/jest-mongodb) to set up an in-memory store (re: `test/case.test.ts`), which requires a few lines of boilerplate. I didn't bother with this, but if that's too ugly/slow for folks [we can abstract it](https://jestjs.io/docs/en/configuration.html#globalsetup-string).

This isn't perfect -- for instance, I'm sure we could catch mongoose ValidationErrors to surface as 422s, in lieu of duplicating some logic via express-validators. But sending it as a starting point to discuss the approach.


data-serving/data-service/jest-mongodb-config.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
```
attwad, line 1: can you comment on what is that file used for please?

data-serving/data-service/src/controllers/case.ts
```diff
@@ -16,30 +17,43 @@ export const get = (req: Request, res: Response): void => {
  * Handles HTTP GET /cases.
  */
 export const list = (req: Request, res: Response): void => {
-    res.send('Triggered list cases.');
+    Case.find({}, (err, c) => {
```
attwad, line 20: seems like this is using callbacks, any chance moongoose supports both callbacks and promises like the mongodb nodejs driver? If not, no big deal, just asking.

data-serving/data-service/test/case.test.ts
```diff
@@ -1,5 +1,30 @@
...
 import app from '../src/index';
+import mongoose from 'mongoose';
+import { Case } from '../src/model/case';
+
+beforeAll(async () => {
+    await mongoose.connect(
+        // This is provided by jest-mongodb.
+        // The `else testurl` is to appease Typescript.
+        process.env.MONGO_URL || 'testurl',
+        { useNewUrlParser: true, useCreateIndex: true },
```
attwad, line 11: no useUnifiedTopology: true here? and why do we need useCreateIndex in tests?

data-serving/data-service/test/case.test.ts
```diff
@@ -1,5 +1,30 @@
...
+        (err) => {
+            if (err) {
+                console.error(err);
+                process.exit(1);
```
attwad, line 15: Looking at https://jest-bot.github.io/jest/docs/setup-teardown.html#one-time-setup makes me think that if beforeAll returns promise (could just be return mongoose.connect() in this case) then you don't need to check for the result and tests would fail if that promise isn't resolved? I might be wrong but please double check as it would simplify the setup.

data-serving/data-service/src/controllers/case.ts
attwad: FYI I did that in my PR it feels cleaner imo than callbacks

https://masteringjs.io/tutorials/mongoose/promise

data-serving/data-service/src/index.ts
```diff
@@ -36,13 +36,11 @@ app.use('/api', apiRouter);
                and collection ${env.DB_COLLECTION}`,
```
attwad, line 36: DB_COLLECTION doesn't seem used anymore?

data-serving/data-service/test/case.test.ts
attwad: FWIW I din't need those in tests, I just used the same connection params as in index.ts

data-serving/data-service/test/case.test.ts
attwad: yeah I confirmed in my PR that the test can simply be:

```
beforeAll(() => {
    return mongoose.connect(
        // This is provided by jest-mongodb.
        // The `else testurl` is to appease Typescript.
        process.env.MONGO_URL || 'testurl',
        { useNewUrlParser: true, useUnifiedTopology: true },
    );
});

afterAll(() => {
    return mongoose.disconnect();
});
```

data-serving/data-service/src/index.ts
attwad: mongoose derives collection names from the first argument in the schema definition, so this is not needed anymore

data-serving/data-service/src/controllers/case.ts
```diff
@@ -16,30 +17,43 @@ export const get = (req: Request, res: Response): void => {
...
  */
 export const create = async (req: Request, res: Response): Promise<void> => {
-    await check('age', 'Age must be a valid number')
-        .not()
-        .isEmpty()
-        .isNumeric()
+    await check('outcome', `Outcome must be one of: ${validOutcomes}`)
+        .isIn(validOutcomes)
+        .run(req);
+    await check('date', 'Date must be avalid ISO 8601 date.')
```
khmoran, line 37: super nit: typo

data-serving/data-service/src/controllers/case.ts
```diff
@@ -16,30 +17,43 @@ export const get = (req: Request, res: Response): void => {
...
- *
- * For now, just attempts to parse an "age" field and return it in the response.
- * Returns 422 if "age" isn't present in the request body.
  */
 export const create = async (req: Request, res: Response): Promise<void> => {
-    await check('age', 'Age must be a valid number')
-        .not()
-        .isEmpty()
-        .isNumeric()
+    await check('outcome', `Outcome must be one of: ${validOutcomes}`)
```
khmoran, line 34: might be nice to (eventually) move this validation logic into the the model, in some sort of static #fromJSON method, so all of the validation is in one place

data-serving/data-service/src/model/case.ts
```diff
@@ -0,0 +1,43 @@
+import mongoose from 'mongoose';
+
+export const validOutcomes = ['pending', 'recovered', 'death'];
+
+const demographicsSchema = new mongoose.Schema({
```
khmoran, line 5: does mongoose have some sort of built-in way to parse a model object from a schema or do we have to roll our own validation? seems like a bummer to define the schema so nicely here but then still have to manually check each field in the request

data-serving/data-service/src/index.ts
khmoran: how does that work with having different collections with the same fields, e.g. one for covid19 and one for ebola?

data-serving/data-service/src/controllers/case.ts
axmb: SG, updated.

data-serving/data-service/test/case.test.ts
axmb: SG, updated.

data-serving/data-service/test/case.test.ts
axmb: Nice find. Updated, thanks.

data-serving/data-service/src/index.ts
axmb: Yeah -- I held off on removing the now-obviated params until we settled on an approach. I'll remove this.

data-serving/data-service/src/controllers/case.ts
axmb: Ah, thank you!

data-serving/data-service/src/controllers/case.ts
axmb: +1 -- we'll have to mess around with this a bit.

data-serving/data-service/src/model/case.ts
axmb: It might be tractable to simply remove document-specific validation from the controller, and to catch ValidationErrors from mongoose. We should include that in the above investigation.

data-serving/data-service/src/index.ts
```diff
@@ -32,17 +32,14 @@ app.use('/api', apiRouter);
...
         console.log(
             `  Connecting to instance ${env.DB_CONNECTION_STRING}
-               with db ${env.DB_NAME}
-               and collection ${env.DB_COLLECTION}`,
+               with db ${env.DB_NAME}`,
         );
 
-        const client = new DbClient(env.DB_CONNECTION_STRING);
-        await client.connect();
+        await mongoose.connect(`${env.DB_CONNECTION_STRING}/${env.DB_NAME}`, {
```
attwad, line 38: do you want to follow the same approach as I did and merge those two envs vars into one? I don't see the benefit of splitting them.

axmb: SG, updated.

axmb:
[Implement the data server CRUD API](https://trello.com/c/xGzQbBcO/47-implement-the-data-server-crud-api)

PR #25, opened 2020-04-22
Connect curator service to mongodb using mongoose

attwad:
Small changes from ab's original PR https://github.com/open-covid-data/healthmap-gdo-temp/pull/24:

- using one env variable, not sure why the need for conenction url + db name was there, a single string seems sufficient
- this is not using the environment validator that the data service is using (to be reconciled in a subsequent PR)
- using promises instead of callbacks when using mongoose find()

### DIFF
--- a/verification/curator-service/src/controllers/sources.ts
+++ b/verification/curator-service/src/controllers/sources.ts
@@ -8,7 +8,6 @@ import { Source } from '../model/source';
 export const list = async (req: Request, res: Response): Promise<void> => {
     const docs = await Source.find({});
     res.send(docs);
-    return;
 };
 
 /**
--- a/verification/curator-service/src/controllers/sources.ts
+++ b/verification/curator-service/src/controllers/sources.ts
@@ -1,10 +1,13 @@
 import { Request, Response } from 'express';
 
+import { Source } from '../model/source';
+
 /**
  * List the sources.
  */
-export const list = (req: Request, res: Response): void => {
-    res.sendStatus(501);
+export const list = async (req: Request, res: Response): Promise<void> => {
+    const docs = await Source.find({});
+    res.send(docs);
     return;
 };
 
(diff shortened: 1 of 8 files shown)


attwad:
[Implement curator service CRUD API for sources and runs](https://trello.com/c/cfofHZ4M/78-implement-curator-service-crud-api-for-sources-and-runs)

verification/curator-service/test/sources.test.ts
```diff
@@ -1,6 +1,27 @@
 import request from 'supertest';
 import app from '../src/index';
 
+import mongoose from 'mongoose';
+
+beforeAll(async () => {
+    await mongoose.connect(
+        // This is provided by jest-mongodb.
+        // The `else testurl` is to appease Typescript.
+        process.env.MONGO_URL || 'testurl',
```
khmoran, line 10: we could optionally use an assertion here (and in index.ts), but that might be too heavy-handed here

https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions

attwad: I'm planning to plug in the env validator that's already plugged into the data validator service, this will ensure the env is valid and this test will be moot, I hope it's okay if I just leave it as is for now.

khmoran: yep sg!

verification/curator-service/src/controllers/sources.ts
```diff
@@ -1,10 +1,13 @@
...
+
 /**
  * List the sources.
  */
-export const list = (req: Request, res: Response): void => {
-    res.sendStatus(501);
+export const list = async (req: Request, res: Response): Promise<void> => {
+    const docs = await Source.find({});
+    res.send(docs);
     return;
```
axmb, line 11: nit: after updating to await/async, I didn't include `return;` in my void controller methods. I don't have a preference, but would be good to standardize (I wonder if there's a linter option).

attwad: removed the return statement as well

PR #26, opened 2020-04-22
Fix validation on provided case max date.

axmb:


axmb:
FYI, @khmoran @attwad

data-serving/data-service/src/model/case.ts
```diff
@@ -18,7 +18,7 @@ const caseSchema = new mongoose.Schema({
     date: {
         type: Date,
         min: '2019-11-01',
-        max: '2020-04-21',
+        max: Date.now,
```
attwad, line 21: How is that better? Shouldn't max be a later date like in 2030 or something?

attwad: Ah nm got it

PR #27, opened 2020-04-22
Wire the remainder of the data-service case API to DB

axmb:


data-serving/data-service/src/controllers/case.ts
```diff
@@ -7,8 +7,13 @@ import { Case, validOutcomes } from '../model/case';
  *
  * Handles HTTP GET /cases/:id.
  */
-export const get = (req: Request, res: Response): void => {
-    res.send(`Triggered get case with ID ${req.params.id}.`);
+export const get = async (req: Request, res: Response): Promise<void> => {
+    const c = await Case.findById(req.params.id);
+    if (!c) {
+        res.status(404).send('Case with provided ID not found.');
```
attwad, line 13: nit but do we want to echo the id in the response to help debug issues?

data-serving/data-service/test/case.test.ts
```diff
@@ -42,23 +59,58 @@ describe('POST', () => {
...
     });
 });
 
 describe('PUT', () => {
-    it('update should return 200 OK', (done) => {
-        request(app).put('/api/cases/id').expect(200, done);
+    it('update present item should return 200 OK', async () => {
+        const id = '53cb6b9b4f4ddef1ad47f943';
+        const c = new Case({ _id: id, outcome: 'pending', date: '2020-04-21' });
+        await c.save();
```
attwad, line 80: nit but save() actually returns the saved document with its id I believe
https://mongoosejs.com/docs/api/document.html#document_Document-save

perhaps it'd be best to use that returned ID instead of faking one ourselves?

axmb: SG, updated.

data-serving/data-service/src/controllers/case.ts
axmb: SG, updated.

axmb:
[Implement the data server CRUD API](https://trello.com/c/xGzQbBcO/47-implement-the-data-server-crud-api)

data-serving/data-service/src/controllers/case.ts
```diff
@@ -52,15 +57,27 @@ export const create = async (req: Request, res: Response): Promise<void> => {
  *
  * Handles HTTP PUT /cases/:id.
  */
-export const update = (req: Request, res: Response): void => {
-    res.send(`Triggered update case with ID ${req.params.id}.`);
+export const update = async (req: Request, res: Response): Promise<void> => {
+    const c = await Case.findByIdAndUpdate(req.params.id, req.body, {
+        new: true,
```
attwad, line 62: you need runValidators: true here as well

PR #29, opened 2020-04-22
Implement CRUD api for sources

attwad:
A few changes from your PR since I looked at it:

- check for validation in update calls
- added useFindAndModify: false when connecting to the DB to remove deprecation warnings

### DIFF
--- a/verification/curator-service/src/controllers/sources.ts
+++ b/verification/curator-service/src/controllers/sources.ts
@@ -7,15 +7,19 @@ import { Source } from '../model/source';
  */
 export const list = async (req: Request, res: Response): Promise<void> => {
     const docs = await Source.find({});
-    res.send(docs);
+    res.json(docs);
 };
 
 /**
  * Get a single source.
  */
-export const get = (req: Request, res: Response): void => {
-    res.sendStatus(501);
-    return;
+export const get = async (req: Request, res: Response): Promise<void> => {
+    const doc = await Source.findById(req.params.id);
+    if (!doc) {
+        res.sendStatus(404);
+        return;
+    }
+    res.json(doc);
 };
 
 /**
--- a/verification/curator-service/src/controllers/sources.ts
+++ b/verification/curator-service/src/controllers/sources.ts
@@ -1,5 +1,4 @@
 import { Request, Response } from 'express';
-
 import { Source } from '../model/source';
 
 /**
--- a/verification/curator-service/src/controllers/sources.ts
+++ b/verification/curator-service/src/controllers/sources.ts
@@ -16,7 +16,9 @@ export const list = async (req: Request, res: Response): Promise<void> => {
 export const get = async (req: Request, res: Response): Promise<void> => {
     const doc = await Source.findById(req.params.id);
     if (!doc) {
-        res.sendStatus(404);
+        res.status(404).json(
+            `source with id ${req.params.id} could not be found`,
+        );
         return;
     }
     res.json(doc);
@@ -33,7 +35,9 @@ export const update = async (req: Request, res: Response): Promise<void> => {
             runValidators: true,
         });
         if (!source) {
-            res.sendStatus(404);
+            res.status(404).json(
+                `source with id ${req.params.id} could not be found`,
+            );
             return;
         }
         res.json(source);
--- a/verification/curator-service/src/controllers/sources.ts
+++ b/verification/curator-service/src/controllers/sources.ts
@@ -26,29 +26,47 @@ export const get = async (req: Request, res: Response): Promise<void> => {
  * Update a single source.
  */
 export const update = async (req: Request, res: Response): Promise<void> => {
-    const source = await Source.findByIdAndUpdate(req.params.id, req.body, {
-        // Return the udpated object.
-        new: true,
-    });
-    if (!source) {
-        res.sendStatus(404);
+    try {
+        const source = await Source.findByIdAndUpdate(req.params.id, req.body, {
+            // Return the udpated object.
+            new: true,
+            runValidators: true,
+        });
+        if (!source) {
+            res.sendStatus(404);
+            return;
+        }
+        res.json(source);
+    } catch (err) {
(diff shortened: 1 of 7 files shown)


attwad:
[Implement curator service CRUD API for sources and runs](https://trello.com/c/cfofHZ4M/78-implement-curator-service-crud-api-for-sources-and-runs)

verification/curator-service/src/index.ts
```diff
@@ -1,12 +1,14 @@
-import express from 'express';
+import bodyParser from 'body-parser';
```
axmb, line 1: Wonder if there's a linter option/plugin for managing import order.

verification/curator-service/src/controllers/sources.ts
```diff
@@ -7,37 +7,71 @@ import { Source } from '../model/source';
...
 /**
  * Update a single source.
  */
-export const update = (req: Request, res: Response): void => {
-    res.sendStatus(501);
-    return;
+export const update = async (req: Request, res: Response): Promise<void> => {
+    try {
+        const source = await Source.findByIdAndUpdate(req.params.id, req.body, {
+            // Return the udpated object.
```
axmb, line 31: nit: sp

verification/curator-service/test/home.test.ts
```diff
@@ -1,6 +1,21 @@
 import request from 'supertest';
 import app from '../src/index';
 
+import mongoose from 'mongoose';
+
+beforeAll(() => {
+    return mongoose.connect(
```
axmb, line 7: Is this necessary, in this file?

My current understanding is that MongoDB can live with a bum connection to localhost until a request is made. As with the other data-service tests, if we set up dotenv with a reasonable `devDefault` for the DB connection URI, we won't need to do this.

I was recently looking into the way jest-mongodb works, since it appears somewhat nuanced. When a test with jest-mongodb is run, it seems the primary connection switches from whatever's specified by the default app config (e.g. a localhost address) to the custom address used for the jest instance. This to say that if the test doesn't actually use MongoDB, it shouldn't matter.

verification/curator-service/test/sources.test.ts
```diff
@@ -16,29 +17,103 @@ afterAll(() => {
...
+
 describe('GET', () => {
-    it('list should return 200', (done) => {
-        request(app).get('/api/sources').expect(200, done);
+    it('list should return 200', async () => {
+        const source = new Source({
+            name: 'test-source',
+            origin: { url: 'http://foo.bar' },
+        });
+        const saved = await source.save();
```
axmb, line 30: mongoose updates the saved document (e.g. you can simply use the `source` object)

verification/curator-service/src/controllers/sources.ts
```diff
@@ -7,37 +7,71 @@ import { Source } from '../model/source';
...
+            new: true,
+            runValidators: true,
+        });
+        if (!source) {
+            res.sendStatus(404);
+            return;
+        }
+        res.json(source);
+    } catch (err) {
+        if (err.name === 'ValidationError') {
```
axmb, line 41: Nice!

verification/curator-service/src/controllers/sources.ts
```diff
@@ -7,37 +7,71 @@ import { Source } from '../model/source';
...
 /**
  * Create a single source.
  */
-export const create = (req: Request, res: Response): void => {
-    res.sendStatus(501);
-    return;
+export const create = async (req: Request, res: Response): Promise<void> => {
+    try {
+        const source = await Source.create(req.body);
+        res.location(`/sources/${source.id}`);
```
axmb, line 56: We should be returning the created object; is that done automagically via middleware by virtue of setting this?

I'd assume the opposite, that location were the potential middleware automagic.

verification/curator-service/src/controllers/sources.ts
```diff
@@ -7,37 +7,71 @@ import { Source } from '../model/source';
...
 /**
  * Get a single source.
  */
-export const get = (req: Request, res: Response): void => {
-    res.sendStatus(501);
-    return;
+export const get = async (req: Request, res: Response): Promise<void> => {
+    const doc = await Source.findById(req.params.id);
+    if (!doc) {
+        res.sendStatus(404);
```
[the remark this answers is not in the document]
attwad: Ah! do what say, not what I do :-p thanks for noticing. Done.

verification/curator-service/src/index.ts
attwad: Ah, that was done by vs code, not sure if there's a linter plugin but I can look around for it...

verification/curator-service/test/home.test.ts
attwad: OK I wasn't really keen on letting mongoose hang on a bum connection during tests, I seemed to notice slightly speedier tests when I added that but let's go for less code, less is better :)

verification/curator-service/src/controllers/sources.ts
attwad: Hmm that's kind of an unofficial REST convention, returning the location of the created object instead of the whole object.
I don't mind returning a 200 and the created object as json if you prefer, the location header containing the url is how I've done that in the past that's all.
**I think the point is that clients shouldn't have to build url paths, they should just navigate with urls given by the server (here follow the location header value).**

- https://restfulapi.net/http-methods/#post
- stackoverflow is more nuanced, there seem to be 3 school of thoughts (use location, return full object, do both)

verification/curator-service/test/sources.test.ts
attwad: nice, thanks!

verification/curator-service/src/controllers/sources.ts
attwad: I promise I tried, but I can't figure out where to put a space here ?? Please let me know what you meant :)

attwad:
Merging this now to move on with a few other things this morning, let's discuss the unclosed comments here later on.

PR #31, opened 2020-04-23
 Render summary from data-service API on the curator-service home page.

axmb:
Couple things:

- Not in love with the solution to type HTTP client responses (e.g. duplicating the interface definition). It's probably worth having a published type for clients to ingest that API data (at least for Node clients), but I'm satisfied with this as a starting point.
- I'll catch up with Allyson on front end plans tomorrow. For now, just getting this on the home page as a placeholder.
- Vagrant set up would be great. ;-) To test locally, I manually ran a local MongoDB and both Node apps. Not terrible, but could be better.

### DIFF
--- a/verification/curator-service/src/controllers/home.ts
+++ b/verification/curator-service/src/controllers/home.ts
@@ -8,14 +8,13 @@ interface Case {
 }
 
 function getLatestCaseDate(cases: [Case]): Date {
-    cases.sort((case1, case2) => {
-        const time1 = case1.date.getTime();
-        const time2 = case2.date.getTime();
-        if (time1 > time2) return 1;
-        if (time1 < time2) return -1;
-        return 0;
-    });
-    return cases[cases.length - 1].date;
+    return cases.reduce((case1, case2) =>
+        // These casts are required.
+        // JSON deserialization doesn't add functions to the object.
+        new Date(case1.date).getTime() > new Date(case2.date).getTime()
+            ? case1
+            : case2,
+    ).date;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


verification/curator-service/src/controllers/home.ts
```diff
@@ -1,10 +1,37 @@
 import { Request, Response } from 'express';
+import axios from 'axios';
+
+interface Case {
+    _id: string;
+    outcome: string;
+    date: Date;
+}
+
+function getLatestCaseDate(cases: [Case]): Date {
```
attwad, line 10: I think you can use Cases.find() and provide a {'sort': {'date': -1} param to do what you want

verification/curator-service/src/controllers/home.ts
```diff
@@ -1,10 +1,37 @@
 import { Request, Response } from 'express';
+import axios from 'axios';
+
+interface Case {
```
attwad, line 4: yeah that duplication is certainly not great once those types will start to grow and we'll have to keep them in sync :( cases are already defined in model/case.ts so let's find a way to export what we need from there. (in a later PR, this is fine for now)

attwad:
Also I started looking at docker compose for the dev stack, I have it running on my machine fine but I need to figure out the github action stuff, expect a PR later today (https://github.com/open-covid-data/healthmap-gdo-temp/pull/32)

verification/curator-service/src/controllers/home.ts
axmb: As with all problems in JS, `reduce` is the simplest I can make this.

verification/curator-service/src/controllers/home.ts
axmb: +1 -- we should discuss more

axmb:
[Implement the curation case data routes and wire it to the data server](https://trello.com/c/xrkBAgQR/49-implement-the-curation-case-data-routes-and-wire-it-to-the-data-server)

PR #35, opened 2020-04-23
Update sample line list data per schema changes

khmoran:
Sorry for the large diff here. Starting with this PR, I'm going to format this json with `python -m json.tool line-list.json` for consistency, unless anyone has another preference.

### DIFF
--- a/data-serving/samples/line-list.json
+++ b/data-serving/samples/line-list.json
@@ -2,7 +2,67 @@
     "_id": {
         "$oid": "5e99f21a1c9d440000ceb088"
     },
+    "demographics": {
+        "age": {
+            "years": 55,
+            "range": {
+                "start": 50,
+                "end": 59
+            }
+        },
+        "sex": "female",
+        "species": "homo sapien"
+    },
+    "location": {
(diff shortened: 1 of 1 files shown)


attwad:
ab and I are using vscode which automatically formats json, I hope our formatters are not going to differ :)

khmoran:
> vscode

do you know what the config is? (indents, json specification, etc) i can try to emulate it

attwad:
I'm afraid the only config in vscode is to enable/disable it:

"json.format.enable": true

khmoran:
> I'm afraid the only config in vscode is to enable/disable it:
> 
> "json.format.enable": true

if you want to give it a shot in vscode and show me the diff, I'll try to work backwards from that. or we can deal with this later :)

attwad:
don't worry about it, I did format it and saw a bunch of changes the biggest one being that you seem to order keys lphabetically whereas vscode doesn't. not a deal breaker I'd say, I don't think we'll be changing that file often.

khmoran:
> Looks good. You might find it easier, while creating the validator, to start minimal and add fields bit-by-bit, but up to you.

yeah tbh I think a lot is going to change so not a bad idea

PR #36, opened 2020-04-23
Curator service dotenv

axmb:


verification/curator-service/src/util/validate-env.ts
```diff
@@ -0,0 +1,23 @@
+import { CleanEnv, cleanEnv, str, port, url } from 'envalid';
+
+export default function validateEnv(): Readonly<{
+    DATASERVIER_API_URL: string;
```
allysonjp715, line 4: CURATORSERVIER ?

verification/curator-service/src/util/validate-env.ts
```diff
@@ -0,0 +1,23 @@
...
+    DB_CONNECTION_STRING: string;
+    PORT: number;
+}> &
+    CleanEnv & {
+        readonly [varName: string]: string | undefined;
+        // eslint-disable-next-line indent
+    } {
+    return cleanEnv(process.env, {
+        DATASERVIER_API_URL: url({
+            desc: 'URL at which to make requests to the data-service API',
```
allysonjp715, line 14: curator-service API

verification/curator-service/src/util/validate-env.ts
axmb: Argh, typo re: servier, thanks for the catch.

This is the URL we're using in the HTTP client, so it's for making requests to the data service.

PR #37, opened 2020-04-23
Initial mongodb validator for line-list case data

khmoran:
This validator is missing a bunch of things, but it's a solid first attempt that at least validates everything's types, required fields, and some of the basic values.

It's still missing things like:

- Requiring one field or another (for source data)
- Actual dictionary values
- Maybe a better way of doing nested validation? (i.e. the location validation is redundant between location and travel_history.location)

### DIFF
--- a/data-serving/data-service/schemas/cases.schema.json
+++ b/data-serving/data-service/schemas/cases.schema.json
@@ -1,0 +1,358 @@
+{
+  "$jsonSchema": {
+    "required": [
+      "_id",
+      "location",
+      "source",
+      "metadata"
+    ],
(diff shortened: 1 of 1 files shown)


khmoran:
Btw this is for use in importing existing data, not in the server -- lmk if you want it to live somewhere else, like in a data-import dir or something

khmoran:
> I'm assuming the action failures are unrelated?

As far as I can tel? @attwad wdyt?

attwad:
Yeah it seems like github still had issues running actions, I reran them manually now and they passed. LGTM.

PR #38, opened 2020-04-24
Move curator backend code to a dedicated dir, in preparation for creating a separate /ui/ dir.

axmb:
Two decisions up for discussion, here:

1. Moving this to a new sub-dir of `curator-service/`, in lieu of leaving it and making a new `verification/curator-ui/` dir.
2. Naming the sub-dirs `api/` and `ui/`.

Chose to do (1) because the backend and UI logically constitute what we've been referring to as the "curator-service." Chose (2) because naming is hard and I wasn't feeling creative (I'd also considered `backend` and `server` -- no real preference).

Intentionally not renaming the docker service; if we decide to create a separate image/app for the frontend, we can name it `curator-ui`.


axmb:
> Would any of the config type files be worth keeping out of the api dir? I'm thinking: eslint, prettierrc, jest.config, tslint, etc. I thought those were doing linting, style fixes, etc that would be useful to have on the UI side, but maybe I'm wrong on what those do :)

This is a great idea -- though I'd have to do a bit of research on how to share these. I think there's some level of assumption that these are in the base Node app dir. Let's revisit as we're setting up the ui dir/app.

PR #39, opened 2020-04-24
Completed schemas for sources

attwad:
I can deal with the merge conflict if you want to submit your PR that moves stuff around first, go ahead.


attwad:

verification/curator-service/src/model/source.ts
```diff
@@ -15,17 +15,68 @@ const originSchema = new mongoose.Schema({
...
+        required: 'Enter a field extraction regexp',
+    },
+});
+
+interface Parsing {
+    fields: Array<Field>;
+}
+
+const parsingSchema = new mongoose.Schema({
+    fields: {
```
axmb, line 39: nit: can this just be `fields: [fieldSchema],`?

verification/curator-service/src/model/source.ts
```diff
@@ -15,17 +15,68 @@ const originSchema = new mongoose.Schema({
...
+    active: boolean;
+    scheduleExpression: string;
+    parsing: Parsing;
+}
+
+const automationSchema = new mongoose.Schema({
+    name: {
+        name: {
+            type: String,
+            required: 'Enter an automation rule name',
```
axmb, line 56: Re: required: this is the "name of the image in docker registry" -- what if they're using the `parsing` field, instead?

verification/curator-service/src/model/source.ts
```diff
@@ -15,17 +15,68 @@ const originSchema = new mongoose.Schema({
...
+        },
+        tag: String,
+        active: Boolean,
+        scheduleExpression: String,
```
axmb, line 60: Something to consider down the line:

https://mongoosejs.com/docs/validation.html#custom-validators

verification/curator-service/src/model/source.ts
attwad: indeed, thanks

verification/curator-service/src/model/source.ts
attwad: Yeah I saw that, I'm waiting to see how we want to optionalize/make mandatory various manual vs automatic configs before adding more validation.
I left a TODO for a regexp validator like "cron(" or "rate("

PR #41, opened 2020-04-24
use regexps in query id params

attwad:
This avoids doing a mongo fetch when an ID is invalid.
Our tests were already checking that we'd return a 404 in that case so nothing changed in tests.

Object IDs are lowercase hexadecimal strings of 24 characters long.
Example: 507f1f77bcf86cd799439011

https://docs.mongodb.com/manual/reference/method/ObjectId/


attwad:

attwad:
the 404 is debatable, if you say you only handle routes that look like X but you get a route that looks like Y, it's okay to throw a 404 imo.

attwad:
I rebased and commits history is a mess... sorry... looks good though so I'll proceed with the merge into master.

PR #42, opened 2020-04-24
add basic model validation tests

attwad:


verification/curator-service/api/src/model/source.ts
```diff
@@ -49,10 +49,7 @@ interface Automation {
 
 const automationSchema = new mongoose.Schema({
     name: {
-        name: {
```
attwad, line 52: removed the required part here as it was a forgotten comment from Alex in a previous PR, now done.

PR #44, opened 2020-04-24
Skeleton of a line-list data conversion python script

khmoran:
This script right now only:

- reads in a local copy of the line-list csv
- changes the ID field to _id
- writes it to a json file

But I wanted to get eyes on it since I'm not a python person, so I'd like to know about any structural issues/areas for improvement early on. Thanks!


data-serving/scripts/convert_data.py
```diff
@@ -0,0 +1,55 @@
+# Lint as: python3
+'''Script to convert CSV line-list data into json compliant with the MongoDB schema.'''
+
+import argparse
+import csv
+import logging
+import json
+
+parser = argparse.ArgumentParser(
```
axmb, line 9: Purely nit/IMO -- I'd stuff this in main() (or a helper function called there). I think it's better not to have "naked" (no class/function) logic.

data-serving/scripts/convert_data.py
```diff
@@ -0,0 +1,55 @@
...
+    print('Converting data to new schema')
+    convert(cases)
+
+    print('Writing results to', args.json_path)
+    write_json(cases, args.json_path)
+
+    print('Great success! 🎉')
+
+
+def read_csv(csv_path):
```
axmb, line 32: If you're loving Typescript, worth considering [typing](https://docs.python.org/3/library/typing.html). Caveat: not widely used, and limited benefits for a one-off script. Just offering.

data-serving/scripts/convert_data.py
```diff
@@ -0,0 +1,55 @@
...
+    cases = []
+    with open(csv_path, newline='') as csvfile:
+        csvreader = csv.DictReader(csvfile)
+        logging.debug('fields: %s', csvreader.fieldnames)
+        for record in csvreader:
```
axmb, line 37: Using for loops in this file is probably fine. If it gets slow (there's a fair amount of total data), I'd switch to reading the csv in with pandas and operating on it as a dataframe, or using the `.values` therefrom to work with numpy. The vectorized functions will be ~thousands of times faster than for loops.

data-serving/scripts/convert_data.py
khmoran: oops idk why i did that -- fixed!

data-serving/scripts/convert_data.py
khmoran: why the heck not!

data-serving/scripts/convert_data.py
khmoran: much faster -- and cleaner! thanks!

attwad:
If we stick to this script it is a good start, just a few things to consider:

- you might want to use "file" types as argumentsto functions instead of string paths, it will make testing easier, I'm not sure if pandas support writing from those though.

- why python? How do you ensure that the json written by this script conforms to our mongoose validators written in typescript? If you write this script using nodejs you could simply re-use those.

- Is the idea to have another script that will then read from the json file and send it to mongo db? Why not doing this as part of this script? Or will this just be mongoimport? (that implies we ditched the idea of importing via the HTTP API, fine by me I just want to make sure we're all agreeing on that)

data-serving/scripts/convert_data.py
```diff
@@ -0,0 +1,51 @@
...
+    converted_cases = convert(original_cases)
+
+    print('Writing results to', args.json_path)
+    write_json(converted_cases, args.json_path)
+
+    print('Great success! 🎉')
+
+
+def read_csv(csv_path: str) -> DataFrame:
+    return pd.read_csv(csv_path, header=0, low_memory=False)
```
attwad, line 37: encoding='utf-8'

always, you'll thank me later

data-serving/scripts/convert_data.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def convert(original_cases: DataFrame) -> DataFrame:
+    logging.info('Converting %d cases', len(original_cases))
+    # TODO: Much, much more conversion.
+    return original_cases.rename(columns={"ID": "_id"})
+
+
+def write_json(cases: DataFrame, json_path: str) -> None:
+    cases.to_json(json_path, orient='records', indent=2)
```
attwad, line 47: do we need this to be pretty indented really? not sure what mongoimport expects feel free to ignore if it's supported.

khmoran: that's for my sanity when I inspect the output :) mongoimport doesn't care either way AFAICT

data-serving/scripts/convert_data.py
khmoran: done, thanks!

khmoran:
> * you might want to use "file" types as argumentsto functions instead of string paths, it will make testing easier, I'm not sure if pandas support writing from those though.

done, thanks!

> * why python? How do you ensure that the json written by this script conforms to our mongoose validators written in typescript? If you write this script using nodejs you could simply re-use those.

I think the short answer is, I can't ensure it conforms to our mongoose validators -- I was planning on using the schema that I checked in.

My rationale for using python instead of nodejs is that this is a "one-off" script (definitely will run it more than once, but hopefully it's still a temporary thing). I didn't want to add a bunch of custom stuff to our server to be able to handle this CSV format. I could probably write a standalone nodejs script, not integrated into our server, that could use the mongoose validators but I thought python would be easier for the purposes of converting the data. WDYT?

> * Is the idea to have another script that will then read from the json file and send it to mongo db? Why not doing this as part of this script? Or will this just be mongoimport? (that implies we ditched the idea of importing via the HTTP API, fine by me I just want to make sure we're all agreeing on that)

I originally planned to have this script import the data into mongodb as well, and I might still -- but I'm not quite sure if that's preferable to mongoimport yet, so I was leaving that to decide later :) I think we did set aside the idea of importing via the API since it's probably just more work.

data-serving/scripts/convert_data.py
```diff
@@ -0,0 +1,53 @@
+# Lint as: python3
```
attwad, line 1: I thought that was a google linter thing? I've never seen that outside google

data-serving/scripts/convert_data.py
attwad: no?

data-serving/scripts/convert_data.py
khmoran: oops, removed

khmoran:

PR #45, opened 2020-04-27
Initial React frontend that queries curator-service API (proxy) and display line list data

axmb:
Apologies that this is a relatively meaty PR. Tooling on this on a Sunday and wanted to make it through the full "feature."

- Moves the curator-service dataserver API call from the home route to a new `/api/cases` route. Doing this, in lieu of calling the dataserver directly from the frontend, because (1) it's cleaner and (2) keeping the frontend lean is pertinent while we concretize the structure (e.g., like whether we're using NextJS). I'm using the `proxy` param in `package.json` to do this without too much heavy lifting.
- Sets up a `components/` dir for the ui app, and moves the existing App root component there.
- Adds a few more React components [that create an initial UI](https://imgur.com/a/m2uxydu).
- Make the UI look not-horrible by using a couple `react-bootstrap` components.
- Component unit tests.

Two things to mention:

- Haven't added the frontend to our docker-compose, so for now I've run all of the apps individually (can do this after).
- Using the default `create-react-app` linting and formatting, so it isn't consistent with our other code. We should probably `eject` to generate the manual config files and set the ui app up analogous to our other apps.

Just to make the review less onerous:

@attwad -- PTAL at `api/`
@allysonjp715 -- PTAL at `ui/`


verification/curator-service/api/src/controllers/cases.ts
```diff
@@ -0,0 +1,24 @@
+import { Request, Response } from 'express';
+import axios from 'axios';
+
+export interface Case {
```
attwad, line 4: should we start a common directory or something where we share those interfaces?
Or should we keep those separate despite the duplication? I can see pros and cons to both...

verification/curator-service/ui/package.json
```diff
@@ -10,7 +11,10 @@
     "@types/node": "^12.12.37",
     "@types/react": "^16.9.34",
     "@types/react-dom": "^16.9.6",
+    "axios": "^0.19.2",
+    "bootstrap": "^4.4.1",
     "react": "^16.13.1",
+    "react-bootstrap": "^1.0.1",
```
attwad, line 17: This is a bold move ;-) Let's see what our designers are going to make us build, I'd have suspected react-material or something alike.

https://material-ui.com/

verification/curator-service/ui/src/components/LinelistRow.tsx
```diff
@@ -0,0 +1,19 @@
+import React from "react";
+import { Case } from "../../../api/src/controllers/cases";
+
+interface RowProps {
+    case: Case
+}
+
+export default class LinelistRow extends React.Component<RowProps, {}> {
```
attwad, line 8: Nit, I'd name this LinelistCaseRow in case we ever want to show something else in the table (separator, summary rows etc).
(This was a pattern I saw in the official react toturial I did this weekend)

(yes I'm ignoring your review guidelines to only look at /api don't blame me I'm curious :) )

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -0,0 +1,67 @@
...
+        }
+    }
+
+    render() {
+        const { errorMessage, isLoaded, linelist } = this.state;
+        if (errorMessage.length > 0) {
+            return <div>Error: {errorMessage}</div>;
+        } else if (!isLoaded) {
+            return <div>Loading...</div>;
+        } else {
```
attwad, line 45: else { not necessary

verification/curator-service/api/src/controllers/cases.ts
axmb: Yeah, I can see both arguments. I don't know whether we should publish the Mongoose document type, but I do think it'd be reasonable for the data-service to publish a "public" dir/package that contains the type, in the least.

Seems like a common problem in Node land; I'll do some research.

verification/curator-service/ui/package.json
axmb: Ha, yeah -- I imagine this will change. :-)

verification/curator-service/ui/src/components/LinelistRow.tsx
axmb: SG, I considered the same myself.

Feel free to review anything! I appreciate it; just a suggestion because I felt bad re: large PR.

verification/curator-service/ui/src/components/LinelistTable.tsx
axmb: SG, un-else'd.

axmb:
[Create a minimal curation frontend to CRUD data](https://trello.com/c/IoplvzNn/50-create-a-minimal-curation-frontend-to-crud-data)

PR #46, opened 2020-04-27
reenabled curl fetches of docker stack in github workflow

attwad:
Last time I merged the CL calling /api/cases from the curator service which made the tests fail and I couldn't figure out why, I was just tired I guess.
Now calling the /api/sources and /api/cases directly instead of / to make it explicit and easier to debug.


attwad:
[docker-compose setup for dev environment](https://trello.com/c/55tz7mmc/41-docker-compose-setup-for-dev-environment)

PR #47, opened 2020-04-27
added pagination support to /api/sources

attwad:


verification/curator-service/api/src/controllers/sources.ts
```diff
@@ -3,10 +3,36 @@ import { Source } from '../model/source';
...
+    if (page < 1) {
+        res.status(422).json('page must be > 0');
+        return;
+    }
+    if (limit < 1) {
+        res.status(422).json('limit must be > 0');
+        return;
+    }
+    const docs = await Source.find({})
+        .skip(limit * page - limit)
```
axmb, line 22: hyper-nit: limit * (page - 1) is probably clearer

axmb, line 22: hyper-nit: limit * (page - 1) is slightly clearer IMO

attwad: done

attwad:
let's see how we can generalize that, but it's quite simple and will only affect /list endpoints so let's see if we really need to do it.

PR #48, opened 2020-04-27
added UI to the dev stack

attwad:
- Got rid of {Case} imports from parent directory as it's forbidden by docker for good reasons.

- stopped using proxy config and instead used environment variables to pass the api endpoint to the react app (https://create-react-app.dev/docs/adding-custom-environment-variables/). Using cors in express server to support UI calling the api server.


attwad:
[docker-compose setup for dev environment](https://trello.com/c/55tz7mmc/41-docker-compose-setup-for-dev-environment)

PR #49, opened 2020-04-27
Add fuller case schema to dataserver.

khmoran:
- Added schema for cases, including all fields' datatypes and some validation
- Changed enums to enums proper
- Changed service layer to auto-magic parse the full body of the request instead of specifying individual fields
- Profit


khmoran:
Screenshot of curator UI with API change reflected: 
image

data-serving/data-service/src/model/case.ts
```diff
@@ -1,43 +1,289 @@
...
+    Work,
+}
+
+const dateValidator = {
+    type: Date,
+    min: '2019-11-01',
+    max: Date.now,
+};
+
+const rangeSchema = new mongoose.Schema({
```
axmb, line 32: Little confusing to have this as "rangeSchema" when there's several other *[Rr]anges in this file.

data-serving/data-service/src/model/case.ts
```diff
@@ -1,43 +1,289 @@
...
+    url: String,
+    other: String,
+});
+
+const pathogenSchema = new mongoose.Schema({
+    name: {
+        type: String,
+        required: 'Enter a pathogen name',
+    },
+    source: sourceSchema,
```
axmb, line 112: What does this represent in the context of a pathogen, again?

data-serving/data-service/src/model/case.ts
```diff
@@ -1,43 +1,289 @@
...
+        max: 200,
+    },
+    maximum: {
         type: Number,
         min: 0,
         max: 200,
     },
+});
+
+const ageSchema = new mongoose.Schema({
```
axmb, line 45: Sort of a higher-level comment: are there any trade-offs of defining every schema explicitly vs. using a nested schema?

https://mongoosejs.com/docs/subdocs.html#altsyntaxarrays
https://mongoosejs.com/docs/subdocs.html#altsyntaxsingle

data-serving/data-service/test/case.test.ts
```diff
@@ -49,35 +49,48 @@ describe('POST', () => {
     it('create without valid date should return 422', (done) => {
```
axmb, line 49: I'd probably just drop this test for now. There's a handful of other stuff that now has validation (we'll have to figure out the testing strategy there); date is no longer required.

data-serving/data-service/test/case.test.ts
```diff
@@ -49,35 +49,48 @@ describe('POST', () => {
...
         const res = await request(app)
             .post('/api/cases')
-            .send({ date: inputDate, outcome: inputOutcome })
+            .send({
+                eventSequence: { confirmed: inputDate },
+                outcome: inputOutcome,
+            })
             .expect('Content-Type', /json/)
             .expect(200);
+        console.log('\n\n' + res.body.toString());
```
axmb, line 78: My precious console lines...

data-serving/data-service/src/model/case.ts
```diff
@@ -1,43 +1,289 @@
...
+    other: {
+        type: [String],
+        uniqueItems: true,
+    },
+});
+
+const metadataSchema = new mongoose.Schema({
+    moderatorId: {
+        type: String,
+        required: 'Enter a moderator id',
```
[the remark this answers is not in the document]
khmoran: generally, they're fields with >90% completeness in the existing data; in this case (moderatorId), I think it just makes sense that new data coming in should have its origin recorded, since we can control that in our system. Confirmed with the epi folks that this initial set makes sense

data-serving/data-service/src/model/case.ts
```diff
@@ -1,43 +1,289 @@
...
         max: 200,
     },
+});
+
+const ageSchema = new mongoose.Schema({
+    years: {
+        type: Number,
+        min: -0.75,
+        max: 200,
+    },
```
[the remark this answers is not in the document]
khmoran: I assumed that these would be normalized ranges (e.g. 50-59), and we'd either only have those, when that's the best granularity available, or the actual years, from which we'd impute the range for ease of analysis.

But this might be an incorrect assumption... wdyt?

data-serving/data-service/test/case.test.ts
khmoran: EEK you found me out. still having trouble getting breakpoints to work with vscode + npm run test

data-serving/data-service/test/case.test.ts
khmoran: oh shoot this date was meant to be required by i accidentally dropped that... adding it back. should we keep the test in that case?

data-serving/data-service/src/model/case.ts
khmoran: No personal preference -- just didn't know where to draw the line, so I did them all explicitly :) Do you have a preference?

data-serving/data-service/src/model/case.ts
khmoran: Where the sequence data can be found, I believe. That's coming from this comment: https://docs.google.com/document/d/1fMPiwohsN2vDyeW7TafPLvAUij4bi49KjB4pTV7kJco/edit?disco=AAAAJcTSnIQ

Looking back at it, it looks like this should be named SequenceSource perhaps

data-serving/data-service/src/model/case.ts
khmoran: fair, will change to ageRangeSchema

data-serving/data-service/src/model/case.ts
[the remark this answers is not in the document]
axmb: Ah. Something to follow up with Stephen on.

If it's normalized ranges, I'm not sure this representation is right, though. Can we do something like a number enum set with the decade (0, 10, 20, 30, [...])?

data-serving/data-service/src/model/case.ts
axmb: Nope. Just curious if you're aware of any material trade-offs.

data-serving/data-service/test/case.test.ts
axmb: SG

data-serving/data-service/src/model/case.ts
```diff
@@ -1,43 +1,293 @@
...
-const caseSchema = new mongoose.Schema({
-    date: {
-        type: Date,
-        min: '2019-11-01',
-        max: Date.now,
-        required: 'Enter a date.',
+const eventSequenceSchema = new mongoose.Schema({
+    onsetSymptoms: dateValidator,
+    confirmed: {
+        ...dateValidator,
```
axmb, line 69: Ellipsis intended?

data-serving/data-service/src/model/case.ts
```diff
@@ -1,43 +1,293 @@
...
+        ...{
+            required: 'Enter a confirmation date',
```
axmb, line 71: This still isn't required in the Case body -- is it? I think this makes the field required in eventSequenceSchema, but it's OK for eventSequenceSchema to be absent from the Case body.

data-serving/data-service/src/model/case.ts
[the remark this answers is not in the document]
khmoran: SG -- will follow up with Stephen on this!

data-serving/data-service/src/model/case.ts
khmoran: oops, good catch, fixed

data-serving/data-service/src/model/case.ts
khmoran: yes, spread operator for combining the two objs

khmoran:

PR #51, opened 2020-04-27
Update line list validator and sample json to use camelCase

khmoran:


data-serving/samples/line-list.json
```diff
@@ -4,7 +4,7 @@
     },
     "demographics": {
         "age": {
-            "years": 55,
+            "years": 55.0,
```
axmb, line 7: Needed?

khmoran: if importing into mongodb, it assumes it's an int and fails validation unless I REALLY encourage it to be a double

khmoran:

PR #52, opened 2020-04-27
Paginate the case data table (and a couple other things)

axmb:
Paginate the case data table (to make the UI workable with a non-trivial amount of data).

Pagination is far better supported with `material-ui` rather than `bootstrap`, so this change makes that migration. Unfortunately the default material table isn't great (no row striping, low-contrast header), so some quick/light theming and stylization is included. Intended to be temporary, but makes the information easier to read, for now.

Comparison of changes:
bs_to_material

### DIFF
--- a/verification/curator-service/ui/src/components/LinelistCaseRow.tsx
+++ b/verification/curator-service/ui/src/components/LinelistCaseRow.tsx
@@ -1,6 +1,9 @@
 import React from "react";
+import TableCell from '@material-ui/core/TableCell';
+import TableRow from '@material-ui/core/TableRow';
 
 interface RowProps {
+    background: string
     case: Case
 }
 
@@ -11,14 +14,18 @@ export interface Case {
 }
 
 export default class LinelistCaseRow extends React.Component<RowProps, {}> {
+
     render() {
         const c = this.props.case;
+        const bg = this.props.background;
         return (
-            <tr>
-                <td>{c._id}</td>
-                <td>{new Date(c.date).toDateString()}</td>
-                <td>{c.outcome}</td>
-            </tr>
+            <TableRow key={c._id} style={{ background: bg }}>
+                <TableCell component="th" scope="row">
+                    {c._id}
+                </TableCell>
+                <TableCell>{new Date(c.date).toDateString()}</TableCell>
+                <TableCell>{c.outcome}</TableCell>
+            </TableRow >
         );
     }
 }
--- a/verification/curator-service/ui/src/components/LinelistTable.tsx
+++ b/verification/curator-service/ui/src/components/LinelistTable.tsx
@@ -1,12 +1,21 @@
 import React from "react";
 import axios from 'axios';
 import LinelistCaseRow, { Case } from "./LinelistCaseRow";
-import Table from 'react-bootstrap/Table';
+import Table from '@material-ui/core/Table';
+import TableBody from '@material-ui/core/TableBody';
+import TableCell from '@material-ui/core/TableCell';
+import TableContainer from '@material-ui/core/TableContainer';
+import TableHead from '@material-ui/core/TableHead';
+import TablePagination from '@material-ui/core/TablePagination';
+import TableRow from '@material-ui/core/TableRow';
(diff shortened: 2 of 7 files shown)


axmb:
[Create a minimal curation frontend to display data](https://trello.com/c/IoplvzNn/50-create-a-minimal-curation-frontend-to-display-data)

verification/curator-service/ui/src/components/LinelistCaseRow.tsx
```diff
@@ -11,14 +14,18 @@ export interface Case {
...
-            <tr>
-                <td>{c._id}</td>
-                <td>{new Date(c.date).toDateString()}</td>
-                <td>{c.outcome}</td>
-            </tr>
+            <TableRow key={c._id} style={{ background: bg }}>
+                <TableCell component="th" scope="row">
+                    {c._id}
+                </TableCell>
+                <TableCell>{new Date(c.date).toDateString()}</TableCell>
```
khmoran, line 26: use the new helper you created? guess you'll h ave to merge anyway :) sorry!

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -1,12 +1,21 @@
...
 import axios from 'axios';
 import LinelistCaseRow, { Case } from "./LinelistCaseRow";
-import Table from 'react-bootstrap/Table';
+import Table from '@material-ui/core/Table';
+import TableBody from '@material-ui/core/TableBody';
+import TableCell from '@material-ui/core/TableCell';
+import TableContainer from '@material-ui/core/TableContainer';
+import TableHead from '@material-ui/core/TableHead';
+import TablePagination from '@material-ui/core/TablePagination';
+import TableRow from '@material-ui/core/TableRow';
```
khmoran, line 10: nit: do we care about sorting imports?

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -35,31 +46,54 @@ export default class LinelistTable extends React.Component<{}, TableState> {
...
+                        <TableHead>
+                            <TableRow style={{ background: '#cfd8dc' }}>
+                                <TableCell>ID</TableCell>
+                                <TableCell>Date</TableCell>
+                                <TableCell>Outcome</TableCell>
+                            </TableRow>
+                        </TableHead>
+                        <TableBody >
+                            {(rowsPerPage > 0
+                                ? linelist.slice(page * rowsPerPage, page * rowsPerPage + rowsPerPage)
```
khmoran, line 77: presumably at some point we'll push the pagination to the API layer and only fetch the rows in the current page?

verification/curator-service/ui/src/components/LinelistCaseRow.tsx
axmb: Yeah, I'll merge tomorrow after the demo (since there's a 90% chance the merge'll get goofed).

verification/curator-service/ui/src/components/LinelistTable.tsx
axmb: I think we do, but I'm not sure. I'll fix this and reinvigorate my search for an auto import sorter.

verification/curator-service/ui/src/components/LinelistTable.tsx
axmb: We'll definitely change this approach at some point. Something like: send an initial request on mount with just the first `rowsPerPage`, then an async request loading the next n or remainder. This "pagination" is mostly about the display, but will have to be extended when we get to that (if it's still around).

verification/curator-service/ui/src/components/LinelistTable.tsx
attwad: https://marketplace.visualstudio.com/items?itemName=amatiasq.sort-imports ?

axmb: Ideally it'd be on the eslint side, but this is a good find.

PR #54, opened 2020-04-28
changed db to covid19 and added global test script

attwad:
Also added a dev script to run all tests instead of having to change directory and run npm test in all of them.

### DIFF
--- a/data-serving/data-service/package.json
+++ b/data-serving/data-service/package.json
@@ -11,12 +11,13 @@
   "scripts": {
     "build": "tsc",
     "dev": "nodemon",
-    "import-data": "mongoimport --db dev --collection cases --file ../samples/line-list.json",
+    "import-data": "mongoimport --db covid19 --collection cases --file ../samples/line-list.json",
     "lint": "tsc --noEmit && eslint '*/**/*.{js,ts,tsx}' --quiet --fix",
     "start": "node dist/server.js",
     "test": "jest --clearCache && jest --forceExit --coverage --verbose --detectOpenHandles",
     "test-watch": "jest --clearCache && jest --verbose --detectOpenHandles --watch",
-    "test-watch-all": "jest --clearCache && jest --coverage --verbose --detectOpenHandles --watchAll"
+    "test-watch-all": "jest --clearCache && jest --coverage --verbose --detectOpenHandles --watchAll",
+    "test-silent": "jest --forceExit --silent --detectOpenHandles"
   },
   "keywords": [],
   "author": "",
--- a/dev/test_all.sh
+++ b/dev/test_all.sh
@@ -1,0 +1,5 @@
+#!/bin/bash
+set -e
+npm --prefix=../verification/curator-service/api/ run-script test-silent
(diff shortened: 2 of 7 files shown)


dev/test_all.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+set -e
+npm --prefix=../verification/curator-service/api/ run-script test-silent
```
khmoran, line 3: why silent?

data-serving/data-service/package.json
```diff
@@ -11,12 +11,13 @@
   "scripts": {
     "build": "tsc",
     "dev": "nodemon",
-    "import-data": "mongoimport --db dev --collection cases --file ../samples/line-list.json",
+    "import-data": "mongoimport --db covid19 --collection cases --file ../samples/line-list.json",
```
khmoran, line 14: how are we going to handle dev/prod instances down the road? obviously this is fine locally but when we're deploying to both prod and non-prod envs in AWS, will we name our dbs in Atlas something like covid19-dev and covid19-prod?

khmoran:
Is anyone using the db in Atlas right now? I can create new ones to reflect this change

dev/test_all.sh
attwad: This script is aimed at being just a quick && simple check that I didn't break too many things, I wanted the output to be minimal (no per-case log) but that still prints errors if there are any. Let me know if you want me to drop the silent, it just suited my usecase.

data-serving/data-service/package.json
attwad: We could keep the same db name between dev and prod, the "instances" will be different, that script currently connects to the default local DB and can't upload to atlas for example.

Or we could separate them yes and have covid19-dev || dev-covid-19, I have no strong preference.

attwad:
I haven't used atlas so far, feel free to rename/recreate the DB

dev/test_all.sh
khmoran: gotcha, sg, the rest of us can always add a more verbose one if we want it :D

data-serving/data-service/package.json
khmoran: hmm... separate instances would probably be cleaner (don't have to change the db names across envs) -- maybe different sharded clusters for each? https://docs.atlas.mongodb.com/reference/glossary/#term-sharded-cluster

anyway, we can decide later once we start setting up CI/CD

verification/curator-service/api/src/util/validate-env.ts
```diff
@@ -16,7 +16,7 @@ export default function validateEnv(): Readonly<{
         }),
         DB_CONNECTION_STRING: str({
             desc: 'MongoDB URI provided to MongoClient.',
-            devDefault: 'mongodb://localhost:27017/dev',
```
axmb, line 19: There's probably an analogous default to update in data-service.

attwad: good catch, done.

PR #55, opened 2020-04-28
[data converter] Convert events and demographics fields

khmoran:
Running on 10% of the data set, only getting a single parsing error (incorrect data format)

Here's an example output for a fairly robust row:

```
{
    "demographics": {
      "age": {
        "range": {
          "min": 70.0,
          "max": 79.0
        }
      },
      "sex": "Male"
    },
    "events": [
      {
        "name": "onsetSymptoms",
        "date": {
          "range": {
            "min": {
              "$date": "2020-02-23T00:00:00Z"
            },
            "max": {
              "$date": "2020-02-23T00:00:00Z"
            }
          }
        }
      },
      {
        "name": "confirmed",
        "date": {
          "range": {
            "min": {
              "$date": "2020-03-06T00:00:00Z"
            },
            "max": {
              "$date": "2020-03-06T00:00:00Z"
            }
          }
        }
      }
    ]
  },
```

### DIFF
--- a/data-serving/scripts/convert_data.py
+++ b/data-serving/scripts/convert_data.py
@@ -6,6 +6,7 @@ import logging
 import json
 import pandas as pd
 import sys
+from constants import CSV_ID_FIELD, OUTPUT_COLUMNS
 from converters import convert_demographics, convert_events
 from pandas import DataFrame
 
@@ -22,7 +23,6 @@ def main():
     parser.add_argument('--sample_rate', default=1.0, type=float)
 
     args = parser.parse_args()
-    logging.info("Args: %s", args)
 
     print('Reading data from', args.infile.name)
     original_cases = read_csv(args.infile)
@@ -47,18 +47,13 @@ def read_csv(infile: str) -> DataFrame:
 
 
 def convert(cases: DataFrame) -> DataFrame:
-    logging.info('Converting %d cases', len(cases))
-
-    # [[_id]]
-    cases = cases.rename(columns={'ID': '_id'})
-
     # [[demographics]]
     cases['demographics'] = cases.apply(
-        lambda x: convert_demographics(x['_id'], x['age'], x['sex']), axis=1)
+        lambda x: convert_demographics(x[CSV_ID_FIELD], x['age'], x['sex']), axis=1)
 
     # [[events]]
     cases['events'] = cases.apply(
-        lambda x: convert_events(x['_id'], {
+        lambda x: convert_events(x[CSV_ID_FIELD], {
             'onsetSymptoms': x['date_onset_symptoms'],
             'admissionHospital': x['date_admission_hospital'],
             'confirmed': x['date_confirmation'],
@@ -66,10 +61,7 @@ def convert(cases: DataFrame) -> DataFrame:
         }, x['outcome']), axis=1)
 
     # Filter out deprecated columns
-    cases = cases.filter(
-        items=['demographics', 'events'])
-
-    return cases
+    return cases.filter(items=OUTPUT_COLUMNS)
 
 
 def write_json(cases: DataFrame, outfile: str) -> None:
--- a/data-serving/scripts/convert_data.py
+++ b/data-serving/scripts/convert_data.py
@@ -2,7 +2,10 @@
 
 import argparse
 import csv
+import datetime
 import logging
+import math
+import numbers
 import json
 import pandas as pd
 import sys
@@ -19,16 +22,24 @@ def main():
     parser.add_argument('--infile', type=argparse.FileType('r'))
     parser.add_argument('--outfile', nargs='?', type=argparse.FileType('w', encoding='UTF-8'),
                         default=sys.stdout)
+    parser.add_argument('--sample_rate', default=1.0, type=float)
+
     args = parser.parse_args()
     logging.info("Args: %s", args)
 
-    print('Reading data from', args.infile)
+    print('Reading data from', args.infile.name)
     original_cases = read_csv(args.infile)
+
+    if args.sample_rate < 1.0:
+        original_size = original_cases.size
+        original_cases = original_cases.sample(frac=args.sample_rate)
+        print(
+            f'Downsampling to {args.sample_rate} cases from {original_size} to {original_cases.size} rows')
 
     print('Converting data to new schema')
     converted_cases = convert(original_cases)
 
-    print('Writing results to', args.outfile)
+    print('Writing results to', args.outfile.name)
     write_json(converted_cases, args.outfile)
 
     print('Great success! 🎉')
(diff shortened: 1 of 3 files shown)


data-serving/scripts/converters.py
```diff
@@ -0,0 +1,121 @@
...
+
+VALID_SEXES = ['female', 'male', 'other']
+
+range_ct = 0
+single_ct = 1
+
+
+def convert_range(value: str, parse_fn: Callable[[str], str], format_fn: Callable[[Any], Any]):
+    '''
+    Converts either a string representation of a single value or a range of values into a range object. The values can be of any types using the custom parse and format functions.
```
axmb, line 19: Nit: linebreak this?

Not sure what editor you're using, but also check out `autopep8` if you haven't added a formatter yet.

data-serving/scripts/converters.py
```diff
@@ -0,0 +1,121 @@
...
+        value_range = [i for i in value_range if i]
+
+        # Handle open ranges (i.e. missing min or max).
+        range_min = parse_fn(value_range[0])
+        if len(value_range) == 1:
+            return {
+                'range': {
+                    'min': format_fn(range_min)
+                }
+            } if value.startswith('-') else {
```
axmb, line 39: Wouldn't starting with - indicate a max?

data-serving/scripts/converters.py
```diff
@@ -0,0 +1,121 @@
...
+
+def convert_range(value: str, parse_fn: Callable[[str], str], format_fn: Callable[[Any], Any]):
+    '''
+    Converts either a string representation of a single value or a range of values into a range object. The values can be of any types using the custom parse and format functions.
+    '''
+    if pd.isna(value):
+        return None
+
+    # First check if the value is represented as a range.
+    if type(value) is str and value.find('-') >= 0:
```
axmb, line 25: Are any other delimiters used (e.g. `:`, `..`)?

data-serving/scripts/converters.py
```diff
@@ -0,0 +1,121 @@
...
+        return {
+            'name': name,
+            'date': date_range
+        }
+    except (TypeError, ValueError):
+        logging.warning(
+            '[%s] [event[name=.%s]] invalid value %s', id, name, date_str)
+
+
+def convert_events(id: str, dates: str, outcome: str) -> List[Dict[str, Any]]:
```
axmb, line 86: I thought these were tuples? E.g. we'd be returning `[ {name: 'eventName1', date: $date1}, ...]`? I must be missing a layer.

data-serving/scripts/convert_data.py
```diff
@@ -38,14 +46,35 @@ def read_csv(infile: str) -> DataFrame:
...
+        lambda x: convert_events(x['_id'], {
+            'onsetSymptoms': x['date_onset_symptoms'],
+            'admissionHospital': x['date_admission_hospital'],
+            'confirmed': x['date_confirmation'],
+            'deathOrDischarge': x['date_death_or_discharge']
+        }, x['outcome']), axis=1)
+
+    # Filter out deprecated columns
+    cases = cases.filter(
+        items=['demographics', 'events'])
```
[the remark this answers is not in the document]
khmoran: you're a mind reader! added this in my next PR but will port to this one

data-serving/scripts/convert_data.py
```diff
@@ -38,14 +46,35 @@ def read_csv(infile: str) -> DataFrame:
...
+    cases['events'] = cases.apply(
+        lambda x: convert_events(x['_id'], {
+            'onsetSymptoms': x['date_onset_symptoms'],
+            'admissionHospital': x['date_admission_hospital'],
+            'confirmed': x['date_confirmation'],
+            'deathOrDischarge': x['date_death_or_discharge']
+        }, x['outcome']), axis=1)
+
+    # Filter out deprecated columns
+    cases = cases.filter(
```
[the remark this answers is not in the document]
khmoran: it is necessary, but I can just change cases = to return anyway

data-serving/scripts/convert_data.py
```diff
@@ -38,14 +46,35 @@ def read_csv(infile: str) -> DataFrame:
     return pd.read_csv(infile, header=0, low_memory=False, encoding='utf-8')
```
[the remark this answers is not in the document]
khmoran: it yelled at me to add it... apparently it doesn't do much but at least it stops yelling at me https://stackoverflow.com/questions/24251219/pandas-read-csv-low-memory-and-dtype-options

data-serving/scripts/convert_data.py
```diff
@@ -38,14 +46,35 @@ def read_csv(infile: str) -> DataFrame:
...
 
 
-def convert(original_cases: DataFrame) -> DataFrame:
-    logging.info('Converting %d cases', len(original_cases))
-    # TODO: Much, much more conversion.
-    return original_cases.rename(columns={"ID": "_id"})
+def convert(cases: DataFrame) -> DataFrame:
+    logging.info('Converting %d cases', len(cases))
```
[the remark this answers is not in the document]
khmoran: hmm i see it imported in both files?

the distinction was that print statements were supposed to just be FYIs for the status of the run, vs. logging was meant to be for writing parsing errors to a file so they could be fixed/accounted for, so tbh we don't need to use it in this file as much as in converters.py. i've removed logging calls in here so it only outputs parser warnings

does that make sense?

data-serving/scripts/convert_data.py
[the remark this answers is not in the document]
khmoran: ah, yes, it does. _id is going to be auto-generated by mongodb on import instead -- the ids used in the original data don't work as ObjectIds (not valid hexes). in the next PR i'm sending, it will be archived in importedCase, though.

removing the line where we rename it from ID to _id since that's useless now

data-serving/scripts/converters.py
```diff
@@ -0,0 +1,121 @@
...
+import math
+import numbers
+import pandas as pd
+from typing import Any, Callable, Dict, List
+
+
+VALID_SEXES = ['female', 'male', 'other']
+
+range_ct = 0
+single_ct = 1
```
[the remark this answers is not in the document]
khmoran: another shame find

data-serving/scripts/converters.py
khmoran: so vscode is using autopep8 to format on save, but it doesn't seem to do comments right? it also *looks* line it's wrapping in vscode but if i less it, it clearly isn't

data-serving/scripts/converters.py
khmoran: ...oops, yes

data-serving/scripts/converters.py
khmoran: not that i've seen yet, for these fields. I think : is used in other fields i haven't mapped yet, for lists (like symptoms)

i've gotten surprisingly few parser errors on the whole dataset -- a few invalid dates, a few ages with strings in them ("9 months")

data-serving/scripts/converters.py
khmoran: not sure I follow -- wouldn't that example you have there be a list of dictionaries (objects)?

data-serving/scripts/convert_data.py
```diff
@@ -38,14 +46,35 @@ def read_csv(infile: str) -> DataFrame:
...
+    # Filter out deprecated columns
+    cases = cases.filter(
+        items=['demographics', 'events'])
+
+    return cases
 
 
 def write_json(cases: DataFrame, outfile: str) -> None:
-    cases.to_json(outfile, orient='records', indent=2)
+    json.dump([row.dropna().to_dict()
```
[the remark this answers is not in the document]
khmoran: the row.dropna() is to remove any of the top-level JSON fields that are empty, like `'demographics': {}`
 AFAICT, DataFrame.to_json doesn't support that option

data-serving/scripts/convert_data.py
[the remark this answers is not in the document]
axmb: Yep, makes total sense.

data-serving/scripts/converters.py
axmb: Interesting, thanks!

data-serving/scripts/converters.py
axmb: I think I'm mixing metaphors (and by metaphors I mean punctuation). I meant to ask: is it better to represent that data as a dict, a tuple of [str, date], or a defined class/object? In the grand scheme of things, if the dict approach is working for you thus far, it's probably good enough.

khmoran: oh, gotcha. I think I chose dict because it's the most json-y, so it's easier for me to scan and visualize the output -- but I'm open to other approaches!

khmoran:

PR #56, opened 2020-04-29
Add react-router to map /new to a WIP page

attwad:
image

### DIFF
--- a/verification/curator-service/ui/package.json
+++ b/verification/curator-service/ui/package.json
@@ -43,6 +43,5 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
-  },
-  "devDependencies": {}
+  }
 }
--- a/verification/curator-service/ui/package.json
+++ b/verification/curator-service/ui/package.json
@@ -15,6 +15,7 @@
     "axios": "^0.19.2",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
+    "react-router-dom": "^5.1.2",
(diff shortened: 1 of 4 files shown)


verification/curator-service/ui/package.json
```diff
@@ -41,5 +42,8 @@
       "last 1 firefox version",
       "last 1 safari version"
     ]
+  },
+  "devDependencies": {
```
axmb, line 46: Nit, but for now it's probably worth just using `dependencies`. There's not a meaningful distinction if our intention is to produce/serve a static bundle. See related discussion on CRA:

https://github.com/facebook/create-react-app/issues/4342

attwad: fair enough, thanks for pointing this out. Is this a react only thing? For non ui packages we've used dev deps already, should we move everything over?

axmb: It's not a React thing -- the distinction between devDeps and deps is simply not meaningful if the intention of your app is to compile into a static bundle rather than serve requests. It does make a difference if you're running a full-blown Node server.

For instance, we don't want to include @types in our production serving binaries, because it's useless bloat. So we use devDeps to indicate they aren't included in prod env. But our React code will never "serve" in prod, it'll just get compiled to a static bundle.

attwad: ack, thanks for the clarification.

attwad:

PR #58, opened 2020-04-30
add basic form to submit new case

attwad:
No client-side validation yet, I can add that later once we "finalize" the schema.
Submitting just prints the form content as json to the console.

Uses formik to make react forms less painful and some material components when available (all except date picker).

image


attwad:

PR #59, opened 2020-04-30
tweak topbar

attwad:
- change empty list icon into a home button until we have a proper navbar on the left
- add a "Add a new case" button in the bar that navigates to /new

image


attwad:

PR #60, opened 2020-04-30
remove curator /api/cases, have the UI directly talk to data /api/cases

attwad:


data-serving/data-service/test/case.test.ts
```diff
@@ -43,7 +43,7 @@ describe('GET', () => {
     });
     it('list should return 200 OK', (done) => {
         request(app)
-            .get('/api/cases')
+            .get('/api/cases/')
```
axmb, line 46: Silly question perhaps: is there something about CORS that requires this? Convention for HTTP APIs is not to apply a terminating slash.

attwad: no, just that we were not consistent with that between the curator api that I removed and the existing data service (that requires a slash).
Let's keep the consistency for now and remove the trailing slash if we need/feel it's better.

PR #61, opened 2020-04-30
Add autopep8 and and ruler properties to vscode config

khmoran:


dev/.vscode/settings.json
```diff
@@ -30,5 +30,13 @@
   "typescript.referencesCodeLens.enabled": true,
   "telemetry.enableTelemetry": false,
   "telemetry.enableCrashReporter": false,
-  "window.zoomLevel": 0
-}
\ No newline at end of file
+  "window.zoomLevel": 0,
+  "python.formatting.provider": "autopep8",
+  "python.linting.pep8Args": [
+    "--max-line-length=100"
```
axmb, line 36: We use 80 for the eslint config -- think it's worth doing the same for python?

khmoran: yeah why not. I don't know why I decided to split it down the middle between 80 and 120, to an unhappy compromise

PR #62, opened 2020-04-30
Submit form to create new cases

attwad:
This contains just a bit more than the minimum  set of fields currently required, I could submit new cases successfully this way even though it's not super pretty.

image


attwad:

PR #64, opened 2020-05-01
Update Case object in UI to reflect new schema

allysonjp715:


verification/curator-service/ui/src/components/LinelistCaseRow.test.tsx
```diff
@@ -15,11 +15,21 @@ it('renders provided case', () => {
...
+                {
+                  name: 'confirmed',
+                  date: {
+                    range: {
+                      start: date,
+                    },
+                  },
+                }
+              ],
+              importedCase: {
```
attwad, line 28: huh, ideally we shouldn't be filling anything in imported cases from this UI, this is only for stuff that comes out of the existing CSV. Not sure why outcome was dropped, we should check with @khmoran .

allysonjp715: Turns out outcome was in the schema (https://github.com/open-covid-data/healthmap-gdo-temp/blob/master/data-serving/data-service/src/model/case.ts) it just wasn't in the sample data. Added to the sample data and reverted to using the string outcome.

attwad: outcome is in the mongoose schema but not in the mongodb schema, @khmoran can you please tell us which one is correct?

https://github.com/open-covid-data/healthmap-gdo-temp/blob/master/data-serving/data-service/schemas/cases.schema.json

khmoran: oops sorry y'all -- the mongodb schema + example case are up-to-date

was working on a PR for the dataserver to make it backward compatible with the fields the UI was using, but I should have done that atomically. from now on will make sure I don't break you!

it's on my list to separate the storage format from the dataserver API format

data-serving/samples/line-list.json
```diff
@@ -168,6 +168,7 @@
         "id": "def456",
         "url": "https://www.colorado.gov/pacific/cdphe/news/10-new-presumptive-positive-cases-colorado-cdphe-confirms-limited-community-spread-covid-19"
     },
+    "outcome": "Pending",
```
khmoran, line 171: outcome was intentionally removed and folded into events: https://docs.google.com/document/d/1fMPiwohsN2vDyeW7TafPLvAUij4bi49KjB4pTV7kJco/edit#heading=h.5qv1o3y0h3l1

(it was removed from the mongodb schema too: https://github.com/open-covid-data/healthmap-gdo-temp/pull/63/files)

you could probably find the event they had as the outcome with something like this:

`this.events.find((e: Event) => e.date == null)`

khmoran: sorry, that code snippet will only work for real data (where 'outcome' has no date) -- in the sample data, I did add a date to the outcome

allysonjp715: Reverted this file change

verification/curator-service/ui/src/components/LinelistCaseRow.test.tsx
allysonjp715: Spoke with Kelly and decided to keep using the importedCase. The schema is going to change a bit and we don't want to keep updating the UI to match the changing schema. So for now we'll rely on the imported case.

PR #65, opened 2020-05-01
Update mongoose schema to match latest mongodb schema

khmoran:
Sorry for the huge diff here. A lot of it is from the test changes.

- Updates mongoose schema
- Adds validation to match design doc
- Adds basic model tests
- Remove detailed model validation tests from controller test

Things that I don't like about this/ that still need to be done:

- Remove dependency in controller test on model specifics (mock out model.save?)
- Right now the samples/line-list.json file still needs to exist, even though it's effectively the same as the fully-specified test file, because mongoose and mongoimport require different formats. Can we dedupe these somehow? Generate one from the other?
- Change API tests to assert that fields that the UI is using are present in response/are parsed by request
- Separate API request/response types from db types


data-serving/data-service/src/model/case.ts
```diff
@@ -114,91 +105,126 @@ const pathogenSchema = new mongoose.Schema({
...
+                        required: 'Enter a name for the event',
+                    },
+                    date: dateRangeSchema,
+                },
+            ],
+            validate: {
+                validator: function (events: [Event]) {
+                    return events.some((e) => e.name == 'confirmed');
+                },
+                message: 'Must include a case confirmed event',
```
axmb, line 180: nit: would be clearer to say 'an event with name equal to "case"'.

data-serving/data-service/src/model/case.ts
```diff
@@ -114,91 +105,126 @@ const pathogenSchema = new mongoose.Schema({
...
-    location: locationSchema,
-    metadata: metadataSchema,
-    notes: String,
-    outbreakSpecifics: outbreakSpecificsSchema,
-    outcome: {
-        type: String,
-        enum: Object.keys(Outcome),
-        required: 'Enter an outcome.',
+    {
+        useNestedStrict: true,
```
axmb, line 205: We don't have any nested `strict` settings.

data-serving/data-service/src/model/case.ts
```diff
@@ -210,14 +236,15 @@ interface Demographics {
...
     imputed: [string];
-    other: [string];
 }
 
-interface EventSequence {
-    onsetSymptoms: Date;
-    confirmed: Date;
-    hospitalAdmission: Date;
-    deathOrDischarge: Date;
+interface EventDate {
```
axmb, line 241: I'm sure I'm missing something obvious -- why define an interface for this?

data-serving/data-service/src/model/case.ts
```diff
@@ -259,34 +288,34 @@ interface Source {
 interface Symptoms {
     provided: [string];
     imputed: [string];
-    other: [string];
 }
 
 interface DateRange {
```
axmb, line 293: `Range<Date>`?

data-serving/data-service/src/model/case.ts
```diff
@@ -259,34 +288,34 @@ interface Source {
...
     start: Date;
     end: Date;
 }
 
+interface TravelDates {
```
axmb, line 298: Same comment as above re: interface necessity.

data-serving/data-service/test/case.test.ts
```diff
@@ -36,110 +33,77 @@ describe('GET', () => {
...
+    it('one absent item should return 404 NOT FOUND', () => {
+        return request(app)
             .get('/api/cases/53cb6b9b4f4ddef1ad47f943')
-            .expect(404, done);
+            .expect(404);
     });
-    it('list should return 200 OK', (done) => {
-        request(app)
+    it('list should return 200 OK', () => {
+        return request(app)
```
axmb, line 42: Just curious, why switch to returning Promises in this file? I'm guessing this'll cause the test framework to wait for each case to resolve, whereas using callbacks won't (IIUC, this is why the callback style is more popular with supertest).

data-serving/data-service/src/model/case.ts
khmoran: I assumed we'd want to use classes where possible, and not objects (when we actually start doing things with models we read in.)

Otherwise it would be:

interface Event {
    name: string;
    date: {
      range: EventDate;
   }
}

Maybe that's OK, but seems a little weird.

The alternative is to not have a nested object with only 'range' in it. I've been going back and forth on this and would love your input. I've included it in an overabundance of caution in case we ever wanted to go back and add other properties to Event.date (other than range... like if we ever want to put a singular date back in). Maybe that's silly?

The alternative is to just have a compressed Event.dateRange of type Range<Date>. WDYT?

data-serving/data-service/src/model/case.ts
khmoran: done, thanks! (event with name "confirmed")

data-serving/data-service/src/model/case.ts
khmoran: hmm I guess I read the documentation wrong -- since it defaults to strict, even if this is 'false', everything will be treated as strict (which is what we want). thanks!

data-serving/data-service/src/model/case.ts
khmoran: oops, yep, thanks

data-serving/data-service/src/model/case.ts
khmoran: see convo above -- whatever we decide for Event.date, I'll apply here too

data-serving/data-service/test/case.test.ts
khmoran: I personally think it's cleaner and easier to trace the flow -- I generally strongly prefer promises to callbacks. From what I've read, it also makes error messages clearer.

But this might be just my prior test framework preference -- if this isn't a very supertest-friendly, I can change it.

Can you explain a little more about waiting for the cases to resolve? Don't we generally call done() at the some point at which we'd be returning the promise?

axmb: Yeah, I generally prefer promises to callbacks as well. Looking back, the article I'm cribbing is for supertest with an older version of Mocha (whereas we're using (the current version of) Jest). There shouldn't be a material difference; this structure SGTM.

data-serving/data-service/src/model/case.ts
axmb: I think it's a little too cautious for now. This is one of the reasons we selected a document DB -- embrace the YAGNI! I had your stated alternative in mind.

khmoran: OK, done. 

We'll need to update the mongodb example case json, the mongodb schema json, and the python converter in another PR to reflect this change

khmoran:
@attwad want to flag that this now throws validation errors on POST from the creator UI -- are you OK with that or do you want me to add some conversion into the controller for that in this PR?

attwad:
omg you broke the frontend?! unforgivable! :pitchforks: :angrymob:
my rule is: did it have a test? if not then please don't feel bad breaking it :)
Nobody's using it so far, just Allyson and I playing with it, we should add an integration test once we feel confortable about it, maybe at the end of this sprint.
I'll just merge this now and fix the frontend today, thanks for flagging.

khmoran:

PR #68, opened 2020-05-04
Paginate /api/cases results

attwad:


data-serving/data-service/src/index.ts
```diff
@@ -26,7 +26,7 @@ app.use(bodyParser.urlencoded({ extended: true }));
 app.get('/', homeController.get);
 const apiRouter = express.Router();
 apiRouter.get('/cases/:id([a-z0-9]{24})', caseController.get);
-apiRouter.get('/cases/', caseController.list);
+apiRouter.get('/cases', caseController.list);
 apiRouter.post('/cases/', caseController.create);
```
axmb, line 30: nit: little incongruous with `create` still documented as `/cases/`.

attwad: ok I will send another PR to remove all trailing slashes when appropriate :)

PR #69, opened 2020-05-04
add total count to pagination of curator service sources api

attwad:


attwad:
[Implement curator service CRUD API for automated source scraping runs](https://trello.com/c/LOOSItyi/80-implement-curator-service-crud-api-for-automated-source-scraping-runs)

verification/curator-service/api/src/controllers/sources.ts
```diff
@@ -19,21 +19,30 @@ export const list = async (req: Request, res: Response): Promise<void> => {
...
-        .limit(limit + 1);
-    // If we have more items than limit, add a response param
-    // indicating that there is more to fetch on the next page.
-    if (docs.length == limit + 1) {
-        docs.splice(limit);
-        res.json({
-            sources: docs,
-            nextPage: page + 1,
-        });
+    try {
```
axmb, line 22: This is identical to the List method in the cases API. Do we want a common utility (on <mongoose.Model>) for this? Or even the entirety of our CRUD? Or is the layer of indirection not worth the LoC?

attwad: Hmm yeah some common library would be nice but we would have to *really* make that a library, by that I mean that the data service code shouldn't import stuff that's outside of its src dir or node_modules, so we'd have to make it an npm package and "install" it.
Seems a bit overkill tbh so for now I'm okay with the duplication.

PR #71, opened 2020-05-04
Update mongodb schema & example json to use range format in mongoose

khmoran:
Also updates the UI to parse these fields

I have a dream to create an API so I don't have to change schema + UI in the same PR... maybe later this week!

### DIFF
--- a/data-serving/data-service/schemas/cases.schema.json
+++ b/data-serving/data-service/schemas/cases.schema.json
@@ -16,25 +16,19 @@
         "bsonType": "object",
         "additionalProperties": false,
         "properties": {
-          "age": {
+          "ageRange": {
             "bsonType": "object",
             "additionalProperties": false,
             "properties": {
-              "range": {
-                "bsonType": "object",
-                "additionalProperties": false,
-                "properties": {
-                  "start": {
-                    "bsonType": "double",
-                    "minimum": -1,
-                    "maximum": 200
-                  },
-                  "end": {
-                    "bsonType": "double",
-                    "minimum": -1,
-                    "maximum": 200
-                  }
-                }
+              "start": {
+                "bsonType": "double",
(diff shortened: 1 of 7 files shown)


data-serving/data-service/schemas/cases.schema.json
```diff
@@ -16,25 +16,19 @@
...
-                  "end": {
-                    "bsonType": "double",
-                    "minimum": -1,
-                    "maximum": 200
-                  }
-                }
+              "start": {
+                "bsonType": "double",
+                "minimum": -1,
+                "maximum": 200
```
attwad, line 26: turtles can live up to almost 300 years!! ;-)
Kidding but I saw "species" before somewhere so if we're talking non-humans we can raise that limit a bit.

verification/curator-service/ui/src/components/NewCaseForm.tsx
```diff
@@ -51,7 +49,7 @@ export default function NewCaseForm() {
                             revisionMetadata: {
                                 id: 0,
                                 moderator: "TODO",
-                                date: "",
+                                date: Date.now(),
```
attwad, line 52: This will make the test fail as the test expects the exact data.

The server should fill the revision object, not the client. Which is why I left it empty for now.

data-serving/data-service/schemas/cases.schema.json
khmoran: done, will fix later in the mongoose validator :)

verification/curator-service/ui/src/components/NewCaseForm.tsx
khmoran: gotcha, will make sure this happens on the dataserver in a future PR

khmoran:

PR #72, opened 2020-05-04
Update converter to export to latest data format

khmoran:
Changes to converter include:
- range representation changes
- the fact that location id should be a string (even though historic data has ints)

Also updates schema to reflect what the imported data will look like (as opposed to new data entered via the server), i.e. the imported data:
- won't have ids (mongodb will autogenerate on import);
- may not have some of the required fields, since almost no column has 100% coverage
- may have many more fields in importedCase

Tested with mongoimport:
`417 document(s) imported successfully. 24 document(s) failed to import.`

### DIFF
--- a/data-serving/data-service/schemas/cases.schema.json
+++ b/data-serving/data-service/schemas/cases.schema.json
@@ -210,10 +210,7 @@
               }
             },
             "purpose": {
-              "enum": [
-                "family",
-                "conference"
-              ]
+              "bsonType": "string"
             },
             "additionalInformation": {
               "bsonType": "string"
--- a/data-serving/data-service/schemas/cases.schema.json
+++ b/data-serving/data-service/schemas/cases.schema.json
@@ -2,12 +2,6 @@
   "$jsonSchema": {
     "bsonType": "object",
     "additionalProperties": false,
-    "required": [
-      "_id",
-      "location",
-      "source",
-      "revisionMetadata"
-    ],
     "properties": {
       "_id": {
         "bsonType": "objectId"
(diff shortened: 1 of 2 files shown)


data-serving/data-service/schemas/cases.schema.json
```diff
@@ -2,12 +2,6 @@
   "$jsonSchema": {
```
axmb, line 2: Do we have any notion of "this is the stuff we're currently importing"? Is it basically look at what's in vs. out of importedCases or the python file?

data-serving/scripts/converters.py
```diff
@@ -41,32 +41,27 @@ def convert_range(
         range_min = parse_fn(value_range[0])
         if len(value_range) == 1:
             return {
-                'range': {
-                    'max': format_fn(range_min)
-                }
+                'start': format_fn(range_min)
```
axmb, line 44: Isn't max == end, not start?

Associated nit: slightly confusing to call this local value `range_min`.

data-serving/scripts/converters.py
```diff
@@ -143,22 +141,22 @@ def convert_location(id: str, location_id: float, country: str, adminL1: str,
 
     try:
         if pd.notna(location_id):
-            location['id'] = float(location_id)
+            location['id'] = str(int(location_id))
```
axmb, line 144: What does the int() conversion do, vs. just str()?

data-serving/scripts/converters.py
khmoran: what is wrong with me

also renamed the local vars

data-serving/scripts/converters.py
khmoran: actually not 100% sure what to do with this.

my thinking was: the existing data has integers for location ids. I'm not sure that will be the case going forward if we adopt a new geocoding system, so I want to keep the field flexible as a string, but I figured I should still throw an error if the existing data doesn't conform to the id system being used.

maybe it's best to omit this field for now and just have it in the originalCase object

khmoran:
> Can you tl;dr what the current import failures are for?

I'm still tracking these down. Next couple PRs will probably be for that.

mongoimport doesn't tell you which documents failed to import or why, so it's going to be a lot of trial and error, unfortunately :(

data-serving/data-service/schemas/cases.schema.json
khmoran: right now it's not documented anywhere -- I'll do my best to document this in the next PR.

right now I'm including everything that's nonempty in importedCases, even if we converted it. may change this -- not sure yet

khmoran:

PR #74, opened 2020-05-04
Use material-table in list view

attwad:
- edit and delete do not work yet, just print to the console for now, will route those later.
- search is disabled until server can understand the ?q= params.
- loading/error/pagination is now handled by the table itself, yay.
- no more need for a custom row component, we could have special components to render some columns, right now the ones I'm displaying do not warrant such an effort.

Looks like that:
image


verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -1,105 +1,101 @@
...
 }
-interface TableState {
-    errorMessage: string,
-    isLoaded: boolean,
-    linelist: Case[],
-    page: number,
-    rowsPerPage: number
+
+interface RowProps {
+    background: string
```
axmb, line 13: Missed clean-up or used via material-table automagic?

data-serving/samples/line-list.json
```diff
@@ -1,6 +1,6 @@
 {
     "_id": {
-        "$oid": "5e99f21a1c9d440000ceb088"
+        "$oid": "5e99f21a1c9d440000ceb089"
```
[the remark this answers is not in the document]
attwad: ah sorry, reverted. I wanted more than one line so I imported that same record twice but had to change the id, mistakenly added it to the PR.

verification/curator-service/ui/src/components/LinelistTable.tsx
attwad: The state is managed by the table itself now so this is not needed anymore.
For the intermitent background color I haven't stylized it yet, I'll look into how to add it back.

PR #75, opened 2020-05-04
Convert notes and revisionMetadata fields

khmoran:
No lossiness since original fields are strings, so still 99.7% conversion success

100% successfully imported into mongodb (with removal of a couple of required fields in revisionMetadata in the schema)

Sample notes conversion:
```
{
   ...
   "notes": "https://new.qq.com/omn/20200215/20200215A0LD4100.html; Contact with other confirmed cases; quarantined",
    "importedCase": {
      "ID": "000-1-10046",
      ...
      "additional_information": "Contact with other confirmed cases; quarantined",
      "notes_for_discussion": "https://new.qq.com/omn/20200215/20200215A0LD4100.html",
      ...
    }
}
```

Sample revisionMetadata conversion:
```
{
    ...
    "revisionMetadata": {
      "id": 0,
      "moderator": "TR"
    },
    "importedCase": {
      "ID": "003-40161",
      ...
      "data_moderator_initials": "TR",
      ...
    }
}
```

### DIFF
--- a/data-serving/scripts/converters.py
+++ b/data-serving/scripts/converters.py
@@ -253,7 +253,7 @@ def convert_dictionary_field(
 def convert_revision_metadata_field(data_moderator_initials: str) -> Dict[
         str, str]:
     '''
-    Populates a revisionMetadata field with an initial revision version of 0 and
+    Populates a revisionMetadata field with an initial revision number of 0 and
     the data moderator's initials where available.
     '''
     revision_metadata = {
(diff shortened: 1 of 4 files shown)


data-serving/scripts/converters.py
```diff
@@ -250,7 +250,30 @@ def convert_dictionary_field(
     return {'provided': [value]}
 
 
-def convert_imported_case(id: str, values_to_archive: Series) -> Dict[str, Any]:
+def convert_revision_metadata_field(data_moderator_initials: str) -> Dict[
```
axmb, line 253: Adding date/notes later?

khmoran: Unfortunately, I don't think we can get these fields from the current data :\ It doesn't seem like the date that a row was added to the dataset was recorded. Tomorrow I'm going to see if we can get it added to the current tool for data going forward -- otherwise it will be hard to get many of Steven's metrics (e.x. last day data was added for a given country).

Notes is mostly to support revisions (per the PRD).

khmoran:

PR #78, opened 2020-05-05
support filtering of notes with regexp

attwad:


data-serving/data-service/src/controllers/case.ts
```diff
@@ -32,14 +32,28 @@ export const list = async (req: Request, res: Response): Promise<void> => {
         res.status(422).json('limit must be > 0');
         return;
     }
+    // Filter query param looks like &filter=notes:work,other_field:foo
+    const filterQuery = String(req.query.filter || '').trim();
+    // Query to mongo looks like { "notes": {$regex: /work/} }
+    // Note: We sould use $text instead of regexp but until we have indexes
```
allysonjp715, line 38: nit: sould -> should

data-serving/data-service/src/controllers/case.ts
```diff
@@ -32,14 +32,28 @@ export const list = async (req: Request, res: Response): Promise<void> => {
...
+    // that's not possible.
+    const query: { [k: string]: { [k: string]: RegExp } } = {};
+    if (filterQuery.length > 0) {
+        const filters = filterQuery.split(',');
+        for (const filter of filters || []) {
```
allysonjp715, line 43: When would the "|| []" come into effect? Shouldn't filters be at least one element long if filterQuery.length > 0?

data-serving/data-service/src/controllers/case.ts
attwad: thanks, fixed.

data-serving/data-service/src/controllers/case.ts
attwad: ah right sorry, leftover from before I added the length check above.

PR #82, opened 2020-05-05
Modify how the table actions are created and implement deleting a row

allysonjp715:
Screen Shot 2020-05-05 at 3 03 32 PM
Screen Shot 2020-05-05 at 3 03 41 PM
Screen Shot 2020-05-05 at 3 03 51 PM

### DIFF
--- a/verification/curator-service/ui/src/components/LinelistTable.tsx
+++ b/verification/curator-service/ui/src/components/LinelistTable.tsx
@@ -9,11 +9,6 @@ interface ListResponse {
     total: number,
 }
 
-interface RowProps {
-    background: string
-    case: Case
-}
-
 interface Event {
     name: string;
     dateRange: {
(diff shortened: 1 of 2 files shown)


verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -41,12 +36,46 @@ interface Case {
...
+            response.then(() => {
+                // Refresh the table data
+                this.state.tableRef.current.onQueryChange();
+            }).catch((e) => {
+                reject(e);
+            });
+        })
+     }
 
+     editCase(newRowData: Case, oldRowData: Case | undefined) {
```
attwad, line 66: In which occasion can we edit something undefined?

allysonjp715:
[Create a linelist data deletion mechanism in the curator UI.](https://trello.com/c/VgbursgA/96-create-a-linelist-data-deletion-mechanism-in-the-curator-ui)

verification/curator-service/ui/src/components/LinelistTable.tsx
allysonjp715: TBH no idea, but it wouldn't build without that type declaration...

allysonjp715:
Fixed test after merge

verification/curator-service/ui/src/components/LinelistTable.tsx
attwad: ok thanks... let's keep that in mind, we should fail gracefully when we understand when that happens.

PR #84, opened 2020-05-05
Implement adding a row directly from the table. I won't remove the ex…

allysonjp715:
…isting add page until the new add flow has parity with the old flow.

Note that the post request doesn't succeed yet. We'll need to add some more columns to the table to be able to successfully add a case. We will also need to add some input verifications such as Female/Male for sex, not inputting _id field, etc.

Screen Shot 2020-05-05 at 4 47 26 PM


allysonjp715:

attwad:
nice!

PR #86, opened 2020-05-06
Convert the travel history column

khmoran:
This is really just a first pass at handling this column -- it's very lossy. The travel_history_location field from the original data is especially unstructured, so we lose a lot of that. (The travel_history_date one is better.)

Stats:
- travelHistory.dateRange misses: 14 / 2134 (0.6%)
- travelHistory.location misses: 2202 / 5508 (40%)

I'm going for high precision, low recall in this PR but I know this can be improved. I have a PR coming that documents the status of everything and will call out how generally recall can be improved in the entire conversion.

Still TODO: Add in lat/longs for the travelHistory.location records that don't have them right now (possibly with geopy)

Example output:

```
    "travelHistory": [
      {
        "dateRange": {
          "start": {
            "$date": "2020-03-17T00:00:00Z"
          },
          "end": {
            "$date": "2020-03-17T00:00:00Z"
          }
        },
        "location": {
          "country": "France"
        }
      }
    ],
```

```
    "travelHistory": [
      {
        "location": {
          "administrativeAreaLevel1": "New York",
          "country": "United States"
        }
      }
    ],
    "importedCase": {
      "ID": "005-8201",
      ...
      "travel_history_location": "New York",
      ...
}
```


khmoran:

data-serving/scripts/constants.py
```diff
@@ -1,2 +1,22 @@
...
+        'administrativeAreaLevel1': 'Hubei',
+        'administrativeAreaLevel2': 'Wuhan City',
+        'locality': 'Wuhan City',
+        'geometry': {
+            'latitude': 30.62506,
+            'longitude': 114.3421
+        }
+    },
+    'iran': {
+        'country': 'Iran',
```
axmb, line 16: There's less resolution as compared to the `wuhan` entry. Is this just made to mirror what we see in the existing data?

data-serving/scripts/converters.py
```diff
@@ -343,14 +348,89 @@ def convert_outbreak_specifics(id: str, reported_market_exposure: str,
...
             lives_in_wuhan)
 
     return outbreak_specifics if outbreak_specifics else None
 
 
+def convert_location_string(location_str: str) -> Dict[str, Any]:
+    if pd.isna(location_str) or location_str.lower() == 'no' or location_str.lower() == 'none':
+        return None
+    if type(location_str) is not str:
+        raise ValueError('location is not a string')
```
axmb, line 363: Error handling is getting a little hard to track in the file. In top-level converters we're catching errors and logging. But, there isn't a totally clear scheme delineating top-level vs. private/sub-converters. Can you think of a clearer way to handle errors?

data-serving/scripts/converters.py
```diff
@@ -343,14 +348,89 @@ def convert_outbreak_specifics(id: str, reported_market_exposure: str,
...
+    except ValueError:
         logging.warning(
             '[%s] [outbreakSpecifics[livesInWuhan]] invalid value %s', id,
             lives_in_wuhan)
 
     return outbreak_specifics if outbreak_specifics else None
 
 
+def convert_location_string(location_str: str) -> Dict[str, Any]:
+    if pd.isna(location_str) or location_str.lower() == 'no' or location_str.lower() == 'none':
```
axmb, line 360: Are these two values the bulk of "str-as-no-value" that we're seeing? Mostly curious; I know this PR was advertised as low precision.

data-serving/scripts/converters.py
```diff
@@ -343,14 +348,89 @@ def convert_outbreak_specifics(id: str, reported_market_exposure: str,
...
+        return None
+    if type(location_str) is not str:
+        raise ValueError('location is not a string')
+    if ';' in location_str or ':' in location_str or location_str.count(',') > 1:
+        raise ValueError('location is a list')
+
+    # Remove common prefixes 'travelled to' and 'traveled to' and lowercase it.
+    normalized_location = re.sub(
```
axmb, line 368: Higher-level question: is this conversion approach (for location) the right strategy? This is a low precision PR, but it's already feeling a little belabored.

Are we ever going to do a good enough job coercing it to our validated schema? Should we just let old data lie?

data-serving/scripts/constants.py
khmoran: yep, it's because Wuhan is a locality, so we can infer locality, admin1, admin2, country, etc. 
whereas for entries where the full string is "Iran", we don't have any more detailed info than just the country (and the centroid lat/long of the country)

data-serving/scripts/converters.py
khmoran: agreed -- please get excited for a forthcoming refactor into converters, parsers, and utils, with clearly defined roles

data-serving/scripts/converters.py
khmoran: yeah, I have a colab where I run value_counts() on the columns to see what the most frequent values are and try to account for those

data-serving/scripts/converters.py
khmoran: TBH I don't know... hoping to get more clarity on this from UXR/product. It seems like the line-list data may be used for research, in which case I do think it would be worthwhile to try to get as much coverage as possible... but yeah, basically need to understand where on the ROC curve we want to be, and how much time we want to invest in getting there :)

khmoran: FYI added a change to the README with these open q's so they're documented somewhere for now

axmb: SG, let's revisit after we get more information from UXR/PM.

data-serving/scripts/converters.py
```diff
@@ -343,14 +348,89 @@ def convert_outbreak_specifics(id: str, reported_market_exposure: str,
...
+    subdivision = pycountry.subdivisions.lookup(normalized_subdivision)
+    return {
+        'administrativeAreaLevel1': subdivision.name,
+        'country': subdivision.country.name
+    }
+
+
+def convert_travel_history(id: str, date_str: str, location_str: str) -> Dict[str, Any]:
+    '''
+    Converts the travel history date and location fields to th enew travelHistory
```
axmb, line 404: sp

data-serving/scripts/converters.py
```diff
@@ -343,14 +348,89 @@ def convert_outbreak_specifics(id: str, reported_market_exposure: str,
...
+    if location is not None:
+        return location
+
+    # Next, attempt to look up the location in a countries database.
+    try:
+        country = pycountry.countries.lookup(normalized_location)
+        return {
+            'country': country.name
+        }
+    except:
```
axmb, line 384: Do we want to specify what we're except'ing, here?

khmoran: done (LookupError)

data-serving/scripts/converters.py
khmoran: done

PR #87, opened 2020-05-06
Update README for converter with details on status of conversion

khmoran:


khmoran:

data-serving/scripts/README.md
```diff
@@ -14,37 +14,92 @@ Logs will be written to `convert_data.log`.
...
-  + Of the 0.3% of rows with errors:
-    - 99.6% have age values outside the valid range
-    - 0.4% have invalid date formats
-- 100% of rows from `cases.json` import successfully into mongodb using `cases.schema.json` as a validator
+- 99.2% of rows from the CSV file convert entirely successfully to JSON. For those with errors, only the failed fields are ommitted.
+- 100% of rows from `cases.json` validate and import successfully into mongodb.
+
+### Lossy fields
+
+A few fields can't be converted losslessly because they need to be normalized upon insertion into the new data format. All of the original values are strings; some of the target representations are bools, ints, floats, or enums. Additionally, the new schema has validation rules and invalid values, e.g. an age over 300. When a value can't be successfully converted, the failure is logged in `convert_data.log` and the value for that field is dropped (though again, the original value is archived in the approrpiate `originalCase` field).
```
axmb, line 22: Mind linebreaking this file? 100 or 120 if you can deal with context switching from code

data-serving/scripts/README.md
```diff
@@ -14,37 +14,92 @@ Logs will be written to `convert_data.log`.
...
+
+The following fields are lossy:
+
+- `demographics.ageRange`: Some values are too large to be ages. Ex. row `002-23162` with age value `2073`.
+- `events[name='onsetSymptoms']`: Some values are in an invalid format, ex. row `000-1-20073` with value `08.03.20202`
```
axmb, line 27: Are we in the business of attempting to coerce data like this (e.g. 08.03.20202 -> 8th-March-2020)? For this field, or others.

khmoran: TBH not sure... we could either fix this in the source, or not at all. I think it's something we need to get a feel from UXR about, and it's not worth putting a ton of effort into it until we know which parts of this data are critical.

data-serving/scripts/README.md
khmoran: sure, did 120

PR #88, opened 2020-05-06
Add S3 integration to the retrieval Lambda

axmb:


axmb:
[Scope and create an AWS infrastructure PoC for automated ingestion.](https://trello.com/c/XmNRDHtQ/114-scope-and-create-an-aws-infrastructure-poc-for-automated-ingestion)

ingestion/functions/retrieval/retrieval.py
```diff
@@ -1,19 +1,66 @@
...
+    except requests.RequestException as e:
+        print(e)
+        raise e
+
+
+def upload_to_s3(file_name, s3_object_key):
+    try:
+        s3_client.upload_file(
+            file_name, OUTPUT_BUCKET, s3_object_key)
+        print("Uploaded source content to bucket {:s} with key {:s}".format(
```
attwad, line 44: in python 3 I'd recommend fstrings to format instead of .format()

f"Uploaded source content to bucket {OUTPUT_BUCKET} with key {s3_object_key}"

that's similar to typescript f strings

ingestion/functions/retrieval/retrieval.py
```diff
@@ -1,19 +1,66 @@
-import json
+import boto3
```
attwad, line 1: is that missing from the requirements.txt?

ingestion/functions/retrieval/retrieval.py
axmb: I'm living in the past! Updated, thanks.

ingestion/functions/retrieval/retrieval.py
axmb: requirements.txt is everything _besides_ the AWS SDK (which is included in the lambda execution environment by default). It's conceivable one might want to pin a version, but it doesn't seem common practice.

attwad: ah ok thanks for the clarification!

PR #89, opened 2020-05-06
Flatten Case data to display it in the table

attwad:
You have noticed I'm sure but we can't pass dotted.field.expression as material-table columns which means we should flatten all the received Case structure to display each field nicely and be able to update them as well.

I am not sure what to make of arrays yet, let's discuss tomorrow.


attwad:

PR #91, opened 2020-05-06
stop testing on node 10.x we're not using it

attwad:
```
timothe-macbookpro~/healthmap/healthmap-gdo-temp(nodes|…) % node --version
v12.16.2
```

No strong feeling about this but seems like we're just wasting free build time of github actions with that node version.


axmb:
Totally agreed. Had considered lumping this into my previous actions clean-up, but balked. Thanks for doing this.

PR #95, opened 2020-05-07
Add auth model and no-op passport integration

attwad:


attwad:

PR #97, opened 2020-05-07
Support authentication via google oauth 2.0

attwad:
Tested locally with a project of mine waiting for Gaurav to create a project for us.

Still a no-op because no /api routes are being forced to be authenticated yet.

### DIFF
--- a/verification/curator-service/api/env
+++ b/verification/curator-service/api/env
@@ -1,0 +1,5 @@
+DATASERVER_API_URL=http://localhost:3000/api
+DB_CONNECTION_STRING=mongodb://localhost:27017/covid19
+PORT=3001
+GOOGLE_OAUTH_CLIENT_ID=471626618881-gel6mfu4rg137qfr3slb1hgi4fsru74h.apps.googleusercontent.com
+GOOGLE_OAUTH_CLIENT_SECRET=-enmhcvQUBHDOoAR-D5do22b
--- a/verification/curator-service/api/src/controllers/auth.ts
+++ b/verification/curator-service/api/src/controllers/auth.ts
@@ -1,0 +1,88 @@
+import { Profile, Strategy, VerifyCallback } from 'passport-google-oauth20';
+import { Request, Response } from 'express';
+import { User, UserDocument } from '../model/user';
+
+import passport from 'passport';
+
+export const router = require('express').Router();
+
+router.get(
+    '/google/redirect',
+    passport.authenticate('google'),
+    (req: Request, res: Response): void => {
+        // After login, redirect to the home page.
+        res.redirect('/');
+    },
+);
+
+router.get('/logout', (req: Request, res: Response): void => {
+    req.logout();
+    res.redirect('/');
+});
+
+router.get(
+    '/google',
+    passport.authenticate('google', {
+        scope: ['email'],
+    }),
+);
+
+export const configurePassport = (
+    clientID: string,
+    clientSecret: string,
+): void => {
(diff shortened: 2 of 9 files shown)


attwad:

verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -0,0 +1,88 @@
...
+import { Request, Response } from 'express';
+import { User, UserDocument } from '../model/user';
+
+import passport from 'passport';
+
+export const router = require('express').Router();
+
+router.get(
+    '/google/redirect',
+    passport.authenticate('google'),
```
allysonjp715, line 11: Could you add a comment describing what this call does?

verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -0,0 +1,88 @@
...
+                    // Passport got our user profile from google.
+                    // Find or create a correpsonding user in our DB.
+                    let user = await User.findOne({ googleID: profile.id });
+                    if (!user) {
+                        user = await User.create({
+                            googleID: profile.id,
+                            name: profile.displayName,
+                            email: (profile.emails || []).map(
+                                (v) => v.value,
+                            )[0],
```
allysonjp715, line 78: Does [][0] not throw an error in typescript? 
Could this be something like:
profile.emails[0].value 
with null checking where appropriate?

verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -0,0 +1,88 @@
...
+                        });
+                    }
+                    cb(undefined, user);
+                } catch (e) {
+                    cb(e, null);
```
allysonjp715, line 83: Could you add a comment describing what value this has?

verification/curator-service/api/test/auth.test.ts
```diff
@@ -0,0 +1,17 @@
...
+import request from 'supertest';
+
+describe('auth', () => {
+    it('login with google', (done) => {
+        // Redirected to the oauth consent screen.
+        request(app).get('/auth/google').expect(302, done);
+    });
+    it('logout', (done) => {
+        // Redirects to /
+        request(app).get('/auth/logout').expect(302, done);
```
allysonjp715, line 11: Is there a way to check it redirected to the correct URL?

verification/curator-service/api/src/controllers/auth.ts
attwad: done.

verification/curator-service/api/src/controllers/auth.ts
attwad: No ends up being "undefined", which will then gets rejected by the storage schema so I'm not checking for it here to keep the code simple and not duplicate checks.

verification/curator-service/api/test/auth.test.ts
attwad: Yes I added check for the location where it was possible, I did not add it for the oauth consent screen as that's hosted by google and it could change and would break tests.

verification/curator-service/api/src/controllers/auth.ts
attwad: added a comment, cb() must be called from within that handler otherwise the page hangs. that's how the passport middleware works.

verification/curator-service/api/env
```diff
@@ -0,0 +1,5 @@
+DATASERVER_API_URL=http://localhost:3000/api
```
khmoran, line 1: Was this intentionally committed?

PR #100, opened 2020-05-07
Create a new script directory for data import, and move data conversion script to a sibling dir

khmoran:
This new node script will be used to set up a db -- primarily for setting up our local dbs prior to running servers (create db if it doesn't exist; collection if it doesn't exist; apply validator to collection; import sample data)

Choosing to use node because mongodb supports js for scripting for the mongo shell: https://docs.mongodb.com/manual/tutorial/write-scripts-for-the-mongo-shell/#scripting


khmoran:

data-serving/scripts/import-data/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "import-data",
+  "version": "1.0.0",
+  "description": "Creates a database and collection as necessary and imports data into it.",
+  "main": "dist/index.js",
+  "scripts": {
+    "build": "tsc",
+    "dev": "nodemon",
```
axmb, line 8: Don't think we want nodemon for this type of script. This should probably just be `"start": "node dist/index.js"`.

data-serving/scripts/import-data/tsconfig.json
```diff
@@ -0,0 +1,61 @@
+{
```
axmb, line 1: We should set up a base tsconfig and extend it for each ts app.

I'll do this in a separate PR -- just remarking here in case the thought occurred to anyone else.

data-serving/scripts/import-data/package.json
khmoran: yeah I think we don't even want to run it with node, we want to run it with mongodb -- going to omit this line from this PR and add it in the next when i have a valid script

data-serving/scripts/import-data/tsconfig.json
khmoran: sgtm

data-serving/scripts/import-data/index.ts
```diff
@@ -0,0 +1 @@
+console.log('Hello, wonderful friends')
```
attwad, line 1: Hi there

data-serving/scripts/import-data/tsconfig.json
attwad: thanks for doing that ab

khmoran:
> Maybe we'll want to move that folder one level higher (next to the dev/ folder for example) as I can see some DB setup coming for the curator service as well (setting up sources, acl roles, etc). I can do that when the time comes, no big deal.

good call!

PR #102, opened 2020-05-07
proxy requests from 3002 to 3001 so that the UI only talks to itself

attwad:
Yes I lied when I said I was out for today, couldn't let that one go sorry.

This is great, this means all /api calls are now made to the same origin. It will make auth much simpler.

I had to customize the proxy as described in https://create-react-app.dev/docs/proxying-api-requests-in-development/#configuring-the-proxy-manually because one cannot use "localhost" in docker composed networks (https://docs.docker.com/compose/networking/).

### DIFF
--- a/verification/curator-service/ui/Dockerfile
+++ b/verification/curator-service/ui/Dockerfile
@@ -20,4 +20,5 @@ EXPOSE 3002
 # Start service, will trigger a build.
 # Environment variables must be passed, notably:
 # - PORT
+# - REACT_APP_PROXY_URL
 CMD [ "npm", "run-script", "start-noenv" ]
--- a/verification/curator-service/ui/Dockerfile
+++ b/verification/curator-service/ui/Dockerfile
@@ -20,5 +20,4 @@ EXPOSE 3002
 # Start service, will trigger a build.
 # Environment variables must be passed, notably:
 # - PORT
-# - REACT_APP_API_ENDPOINT
 CMD [ "npm", "run-script", "start-noenv" ]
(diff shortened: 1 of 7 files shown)


verification/curator-service/ui/Dockerfile
```diff
@@ -20,5 +20,4 @@ EXPOSE 3002
 # Start service, will trigger a build.
 # Environment variables must be passed, notably:
 # - PORT
-# - REACT_APP_API_ENDPOINT
```
axmb, line 23: This is now REACT_APP_PROXY_URL, right?

verification/curator-service/ui/package.json
```diff
@@ -26,7 +27,7 @@
   },
   "scripts": {
     "start-noenv": "react-scripts start",
-    "start": "PORT=3002 REACT_APP_API_ENDPOINT=http://localhost:3001 react-scripts start",
+    "start": "PORT=3002 react-scripts start",
```
axmb, line 30: If you're in the business of `|| <default>` for the API endpoint, do you want to just do the same for the port?

verification/curator-service/ui/src/setupProxy.js
```diff
@@ -0,0 +1,11 @@
+const { createProxyMiddleware } = require('http-proxy-middleware');
```
axmb, line 1: Should probably `.ts` this, given that it's in src.

verification/curator-service/ui/Dockerfile
attwad: ah yeah sorry I did that before adding a default value that doesn't work with docker, added back.

verification/curator-service/ui/src/setupProxy.js
attwad: Actually no, the docs say that this should be named setupProxy.js.
react automatically calls that if it exists, I am not sure we can make it a typescript thing considering app is being passed as a param and comes from "somewhere" that I couldn't find out.

attwad:

PR #105, opened 2020-05-08
make ID column not editable

attwad:
Edit:
image

Add:
image


attwad:

PR #106, opened 2020-05-08
Proxy /api/cases calls manually in curator frontend

attwad:
The reason I'm going back to this is because once we add auth (which is the thing I'm doing now), we'll authenticate the user in the curator service and change the post request to contain its id (for storing contribution information), amongst other kind of data-mangling thing we're going to do in the curator service before sending the request to the data service.

I went all-in and removed the http proxy library I added yesterday to keep things consistent, I didn't want to proxy only certain methods and not others to keep things simple and in one place.


attwad:

verification/curator-service/api/src/controllers/cases.ts
```diff
@@ -0,0 +1,73 @@
+import { Request, Response } from 'express';
+
+import axios from 'axios';
+
+/**
+ * CasesController forwards requests to the data service.
+ * It handles CRUD operations from curators.
```
axmb, line 7: nit: default

PR #108, opened 2020-05-08
Add an /auth/profile API handler to prepare for showing login/logout buttons in the UI navbar

attwad:


attwad:

PR #110, opened 2020-05-08
Add a private global module for mongo-cli types

khmoran:
This allows the import-data script, developed in ts, to understand the types from the mongo CLI.

Used these references for building it:

- https://www.typescriptlang.org/docs/handbook/declaration-files/library-structures.html#global-libraries
- https://www.typescriptlang.org/docs/handbook/declaration-files/templates/global-d-ts.html

And these references for the types:

- https://docs.mongodb.com/manual/tutorial/write-scripts-for-the-mongo-shell/
- https://docs.mongodb.com/manual/reference/method/

### DIFF
--- a/data-serving/scripts/import-data/mongo-cli/index.d.ts
+++ b/data-serving/scripts/import-data/mongo-cli/index.d.ts
@@ -1,0 +1,74 @@
+// Type definitions for mongo-cli.
+
+/**
+ * MongoDB's main connection-type object, which can be coerced into the
+ * `Connection` interface in usage.
+ */
+declare let Mongo: any;
+
+/** Mongo CLI has its own print and read functions. OK, I guess! */
+declare function print(message: string): void;
+declare function printjson(message: number): void;
+declare function cat(path: string): string;
+
+/**
+ * The main connection object, representing the connection to the instance
+ * specified arguments to `mongo`. Provides access to databases.
+ */
+interface Connection {
+    host: string;
+    getDB: (name: string) => Promise<Database>;
+}
+
+/**
+ * An object representing a database in the instance. The database may have
+ * collections, and can have commands run on it.
+ */
+interface Database {
+    createCollection: (
+        name: string,
+        options?: CreateCollectionOptions,
+    ) => Promise<CommandResult>;
+
+    getCollectionNames: () => Promise<[string]>;
+
+    runCommand: (
+        command: InsertCommand | CollModCommand,
+    ) => Promise<CommandResult>;
+}
+
+/** A command to modify a collection, as an arg to `runCommand`. */
+interface CollModCommand {
+    collMod: string;
+    validator: JSON;
+}
+
+/**
+ * A command to insert documents into a collection, as an arg to `runCommand`.
+ */
+interface InsertCommand {
+    insert: string;
+    documents: JSON;
+}
+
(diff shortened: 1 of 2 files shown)


khmoran:

khmoran:
> Surprised that this doesn't exist in some popular capacity in the wild.
> 
> Perhaps most folks don't have the _courage_ to use TS for these mongo shell scripts.

I found one but it didn't support the parts of the API we need :'(

We must be the change we want to see in the world.

axmb:
Looking back at this, I forgot to ask -- where are you hosting the private module? Are you setting it up with GitHub, or are you creating an npm registry account?

khmoran:
I'm planning to reference it locally.

Made a draft PR so you can see it here: https://github.com/open-covid-data/healthmap-gdo-temp/pull/114/files#diff-f3ca199996293a7af9182c1d0e03223d

khmoran:
> Looking back at this, I forgot to ask -- where are you hosting the private module? Are you setting it up with GitHub, or are you creating an npm registry account?

I'm planning to reference it locally.

Made a draft PR so you can see it here: https://github.com/open-covid-data/healthmap-gdo-temp/pull/114/files#diff-f3ca199996293a7af9182c1d0e03223d

axmb:
> > Looking back at this, I forgot to ask -- where are you hosting the private module? Are you setting it up with GitHub, or are you creating an npm registry account?
> 
> I'm planning to reference it locally.
> 
> Made a draft PR so you can see it here: https://github.com/open-covid-data/healthmap-gdo-temp/pull/114/files#diff-f3ca199996293a7af9182c1d0e03223d

Not totally clear on this. If you're referencing it by file, and just using it in the import script, it'd be simpler just to put the dts file directly it in something like `src/types/`, then add that types dir to the `tsconfig` paths option. What you have in that pull request might work -- because I'm not sure if npm will require the package to "exist" in the registry if there's a local path provided. I would _think_ it won't work; what you're declaring is that you'd like to use the `mongo-cli` package from the `@types` [org/scope](https://github.com/DefinitelyTyped/DefinitelyTyped#readme), and you happen to have a copy of it locally that you'd like to use.

khmoran:
> > > Looking back at this, I forgot to ask -- where are you hosting the private module? Are you setting it up with GitHub, or are you creating an npm registry account?
> > 
> > 
> > I'm planning to reference it locally.
> > Made a draft PR so you can see it here: https://github.com/open-covid-data/healthmap-gdo-temp/pull/114/files#diff-f3ca199996293a7af9182c1d0e03223d
> 
> Not totally clear on this. If you're referencing it by file, and just using it in the import script, it'd be simpler just to put the dts file directly it in something like `src/types/`, then add that types dir to the `tsconfig` paths option. What you have in that pull request might work -- because I'm not sure if npm will require the package to "exist" in the registry if there's a local path provided. I would _think_ it won't work; what you're declaring is that you'd like to use the `mongo-cli` package from the `@types` [org/scope](https://github.com/DefinitelyTyped/DefinitelyTyped#readme), and you happen to have a copy of it locally that you'd like to use.
> 
> edit: Apologies that I didn't see the package file in the PR diff on review and raise this then.

that makes a lot of sense -- thanks for explaining. new to TS so i went in a weird direction but just sent you a PR to change it!

PR #109, opened 2020-05-08
Create/use a base tsconfig file.

axmb:

PR #111, opened 2020-05-08
Raise docker-compose build context dir to allow file sharing

axmb:
Specify services build context and dockerfile separately. Define the context as the epid root dir, allowing use to use shared files (e.g. #109).

There are two downsides here, which are:

- Forced compartmentalization can be a good thing at times (though Node TS provides this, to an extent).
- Slightly more cognitive overhead in our docker config.

Happy to revert if we don't want this, though I'm leaning "worth" right now.

### DIFF
--- a/data-serving/data-service/Dockerfile
+++ b/data-serving/data-service/Dockerfile
@@ -1,18 +1,18 @@
 FROM node:12
 
 # Create app directory
-WORKDIR /usr/src/app
+WORKDIR /usr/src/app/data-serving/data-service
 
 # Install app dependencies
 # A wildcard is used to ensure both package.json AND package-lock.json are copied
 # where available (npm@5+)
-COPY package*.json ./
+COPY data-serving/data-service/package*.json ./
 
 # RUN npm clean install
 RUN npm ci
 
 # Bundle app source
-COPY . .
+COPY data-serving/data-service/. .
 
 # Build the app
 RUN npm run-script build
--- a/dev/docker-compose.yml
+++ b/dev/docker-compose.yml
@@ -9,7 +9,9 @@ services:
         environment:
             MONGO_INITDB_DATABASE: "covid19"
     curator:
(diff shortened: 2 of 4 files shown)


data-serving/data-service/Dockerfile
```diff
@@ -1,18 +1,18 @@
...
 
 # Create app directory
-WORKDIR /usr/src/app
+WORKDIR /usr/src/app/data-serving/data-service
 
 # Install app dependencies
 # A wildcard is used to ensure both package.json AND package-lock.json are copied
 # where available (npm@5+)
-COPY package*.json ./
+COPY data-serving/data-service/package*.json ./
```
attwad, line 9: that's a bit weird no? now the Dockerfile is dependent on the docker compose context and can't be used as is.
If you go to this directory and do `docker build .` (which is how you "normally" would build stuff) it will fail no?
Is there a way for us to keep the dockerfiles independent of the docker compose setup? (I'm thinking this will become handy once we start deploying those to aws, we should reuse those docker files if we can)

axmb: I wouldn't say it's based on anything about the compose, per se -- just that the build context is now at the root of our repository. Definitely a downside that `docker build .` in the Node dir doesn't work; on the flip side, every service is now dockerized from the same dir (context). So we'd run:

```
$ cd healthmap-gdo-temp
$ docker build . -f data-serving/data-service/Dockerfile
$ docker build . -f [whatever else you wanted to build]
```

Basically the question is do we want to allow file sharing within our repo. If yes, then the "right" build context is the root of the repo. If no, and we want to keep services totally file-isolated, the right context is the project dirs as we have now (and I'll revert the file sharing PRs).

attwad: You're right we're having multiple services in the same repo so it might make sense to consider it a mini-mono-repo and have the build context consistent between all these services. I missed that aspect. LGTM!

PR #113, opened 2020-05-08
Add 99 more sample documents to the sample case data

khmoran:
Also renaming to cases.json for consistency

This gives us 100 sample files to work with, the first one still being fully-specified


khmoran:

PR #116, opened 2020-05-08
Initial script to setup db: create collection, apply validator, and drop data

khmoran:
Does NOT include inserting data; this can be used in conjunction with `mongoimport` to do that (and the npm script in data-serving/data-service). (I ran into some inconsistent behavior -- to put it lightly -- and think it's best to use mongoimport for that piece.)

Left a TODO for adding indexes once we have defined those.

I'm super aware that the npm script for running this is truly insane, so I'm open to suggestions.

TESTED:
```
npm run setup-cases

> import-data@1.0.0 setup-cases /usr/local/google/home/khmoran/Dev/healthmap-gdo-temp/data-serving/scripts/import-data
> tsc && mongo --eval "let args={databaseName: 'covid19', collectionName: 'cases', schemaPath: './../../data-service/schemas/cases.schema.json'}" dist/index.js

MongoDB shell version v4.2.6
connecting to: mongodb://127.0.0.1:27017/?compressors=disabled&gssapiServiceName=mongodb
Implicit session: session { "id" : UUID("95a3e73d-08c0-4d1c-a4fd-2d56a8528655") }
MongoDB server version: 4.2.6
Dropped existing data
Applied schema to collection "cases"
{ "collection" : "covid19.cases", "documents" : 0, "indexes" : 1 }
Thank you for joining us 💾
```

Then, from the data-serving/data-server:
```
npm run import-data

> data-service@1.0.0 import-data /usr/local/google/home/khmoran/Dev/healthmap-gdo-temp/data-serving/data-service
> mongoimport --db covid19 --collection cases --jsonArray --file ../samples/cases.json

2020-05-08T18:09:06.394-0400    connected to: mongodb://localhost/
2020-05-08T18:09:06.405-0400    100 document(s) imported successfully. 0 document(s) failed to import.
```

To run the setup-db script with arbitrary params:
`DB="somedb" COLL="somecollection" SCHEMA="../some/schema/path.json" npm run setup`

### DIFF
--- a/data-serving/scripts/import-data/src/index.ts
+++ b/data-serving/scripts/import-data/src/index.ts
@@ -23,35 +23,24 @@ const setupDatabase = async ({
         const database = await connection.getDB(databaseName);
         print(`Connected to database "${database}"`);
 
-        // Either update the collection with the schema if it already exists, or
-        // create the collection with the schema.
-        let collection;
+        // If the collection already exists, drop it so we can recreate it
+        // fresh.
         if (
             (await database.getCollectionNames()).some(
                 (c) => c == collectionName,
             )
         ) {
-            // Drop existing data, if any.
-            collection = await database.getCollection(collectionName);
-            collection.remove({});
-            print('Dropped existing data');
-
-            // Update the collection with the schema validator.
-            database.runCommand({
-                collMod: collectionName,
-                validator: schema,
-            });
-            print(`Applied schema to collection "${collectionName}"`);
-        } else {
-            await database.createCollection(collectionName, {
-                validator: schema,
-            });
-            print(`Created collection "${collectionName}" with schema`);
+            await (await database.getCollection(collectionName)).drop();
+            print('Dropped existing collection');
         }
 
+        await database.createCollection(collectionName, {
+            validator: schema,
+        });
+        print(`Created collection "${collectionName}" with schema`);
+
+        const collection = await database.getCollection(collectionName);
         // TODO(khmoran): Create indexes.
-        collection =
-            collection ?? (await database.getCollection(collectionName));
 
         // Print some stats -- for fun and confirmation!
         const stats = await collection.stats();
--- a/data-serving/scripts/import-data/src/index.ts
+++ b/data-serving/scripts/import-data/src/index.ts
@@ -1,1 +1,71 @@
-new Mongo()
+declare let args: SetupDatabaseParameters;
+
+interface SetupDatabaseParameters {
+    databaseName: string;
+    collectionName: string;
+    schemaPath: string;
+}
+
+const setupDatabase = async ({
+    databaseName,
+    collectionName,
+    schemaPath,
+}: SetupDatabaseParameters): Promise<void> => {
+    try {
+        const schema = JSON.parse(await cat(schemaPath));
+        print(`Read schema from ${schemaPath}`);
+
+        // Connect to the default MongoDb instance.
+        const connection: Connection = new Mongo();
+        print(`Connected to instance at ${connection['host']}`);
+
+        // Get or create the database. (If it doesn't exist, it will be created.)
+        const database = await connection.getDB(databaseName);
+        print(`Connected to database "${database}"`);
+
+        // Either update the collection with the schema if it already exists, or
+        // create the collection with the schema.
+        let collection;
+        if (
+            (await database.getCollectionNames()).some(
+                (c) => c == collectionName,
+            )
+        ) {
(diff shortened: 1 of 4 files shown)


khmoran:

data-serving/scripts/import-data/src/index.ts
```diff
@@ -1 +1,71 @@
-new Mongo()
\ No newline at end of file
+declare let args: SetupDatabaseParameters;
+
+interface SetupDatabaseParameters {
+    databaseName: string;
+    collectionName: string;
+    schemaPath: string;
```
attwad, line 6: nit, up to you but to simplify things a bit perhaps you can import the ./../../data-service/schemas/cases.schema.json directly?

https://mariusschulz.com/blog/importing-json-modules-in-typescript#importing-json-files-via-static-import-declarations

We can parameterize things later of if we need to, or not.

data-serving/scripts/import-data/src/index.ts
```diff
@@ -1 +1,71 @@
...
+        // create the collection with the schema.
+        let collection;
+        if (
+            (await database.getCollectionNames()).some(
+                (c) => c == collectionName,
+            )
+        ) {
+            // Drop existing data, if any.
+            collection = await database.getCollection(collectionName);
+            collection.remove({});
```
attwad, line 36: FYI I had a look at dropDatabase (https://docs.mongodb.com/manual/reference/method/db.dropDatabase/) but it seems that we need to restart a bunch of things if we create a new database after doing that, so doing it the way you did looks good.

attwad, line 36: remove is deprecated http://mongodb.github.io/node-mongodb-native/3.5/api/Collection.html#remove, welcome to goog^R^Rmongo.
The docs say to use deleteOne, deleteMany or bulkWrite.

attwad: Perhaps you can just use drop? 
http://mongodb.github.io/node-mongodb-native/3.5/api/Collection.html#drop

attwad, line 36: it's not written in the drop() docs but perhaps call dropAllIndexes just to be sure?

data-serving/scripts/import-data/src/index.ts
```diff
@@ -1 +1,71 @@
...
+        } else {
+            await database.createCollection(collectionName, {
+                validator: schema,
+            });
+            print(`Created collection "${collectionName}" with schema`);
+        }
+
+        // TODO(khmoran): Create indexes.
+        collection =
+            collection ?? (await database.getCollection(collectionName));
```
attwad, line 54: TIL about ??, thank you
Although in that case I don't understand why getCollection would return a boolean that would be mishandled by the || operator?

https://docs.mongodb.com/manual/reference/method/db.getCollection/#db.getCollection

data-serving/scripts/import-data/src/index.ts
khmoran: we def can -- was trying to leave it parameterized in case we wanted to move this up a dir and reuse for other collections (sources, users)

if we're not parameterizing that, is there any point to parameterizing the db/collection names, either? seems like those won't change for cases

attwad: okay then maybe let's see how much we can reuse once we run it for another kind of collection, for now let's keep it as is.

khmoran: btw for posterity, if we do keep the parameterization, this is the syntax for `npm run setup` (as opposed to `npm run setup-cases`

`DB="somedb" COLL="somecollection" SCHEMA="../some/schea/path.json" npm run setup`

data-serving/scripts/import-data/src/index.ts
khmoran: don't need this anymore but you're right that || would have worked! i am bad at polyglotting... was channeling dart

data-serving/scripts/import-data/src/index.ts
khmoran: this documentation (for the shell command specifically) confirms that drop() will remove indexes: https://docs.mongodb.com/manual/reference/method/db.collection.drop/

khmoran: will use drop, thanks!

khmoran: SG, thanks!

khmoran:
> Good use of unicode.

thank you 🙏

PR #117, opened 2020-05-11
add login/logout buttons to navbar

attwad:
image

After login:
image

Steps to try out locally:
```
# set ENV secret and client id from cloud project (I'm using a personal one for now until Gaurav gives us one)
./dev/run_stack.sh
Then go to http://localhost:3001/auth/google, accept the oauth prompt.
Then go to http://localhost:3002 to access the UI.
```

Because we run the server and the UI on different ports, clicking the login button doesn't work and you have to go to the localhost:3001/auth/google manually now , it will work in prod because they will be served from the same domain but until I find a better solution this manual step is needed.

Logout works as expected.

### DIFF
--- a/verification/curator-service/ui/src/components/EpidNavbar.tsx
+++ b/verification/curator-service/ui/src/components/EpidNavbar.tsx
@@ -1,52 +1,84 @@
-import React, { useEffect } from "react";
-import { Theme, createStyles, makeStyles } from '@material-ui/core/styles';
+import { Button, withStyles } from "@material-ui/core";
+import { Theme, createStyles } from '@material-ui/core/styles';
 
 import AppBar from '@material-ui/core/AppBar';
 import HomeIcon from '@material-ui/icons/Home';
 import IconButton from '@material-ui/core/IconButton';
 import { Link } from "react-router-dom";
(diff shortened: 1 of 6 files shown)


attwad:

verification/curator-service/ui/src/components/EpidNavbar.tsx
```diff
@@ -1,52 +1,84 @@
...
+                                data-testid="home-btn"
+                                edge="start"
+                                className={classes.menuButton}
+                                aria-label="menu">
+                                <HomeIcon />
+                            </IconButton>
+                        </Link>
+                        <Typography variant="h6" className={classes.title}>
+                            epid
                     </Typography>
```
allysonjp715, line 68: super-nit: increase indentation to match <Typography

attwad:
No login isn't required anywhere yet. I'll add instructions once it's required.

verification/curator-service/ui/src/components/EpidNavbar.tsx
attwad: thanks, weird I have autoformat on save but it didn't catch that. Done.

PR #118, opened 2020-05-11
add auth middleware that enforces user authentication

attwad:
This will simplify things once we start enforcing auth on API calls.


attwad:

PR #119, opened 2020-05-11
Store session info in Mongo

attwad:
Otherwise each time the server reloads when we change something the default "in-memory" store gets killed and we have to re-login that's annoying.

All tests importing the top-level app object now require a mongoose connection, so I've mocked them all up.


attwad:

PR #120, opened 2020-05-11
Add field validators when editting or adding rows

allysonjp715:
- Makes sex a dropdown between Male / Female.
Screen Shot 2020-05-11 at 2 07 30 PM

- Makes age be a number
Screen Shot 2020-05-11 at 2 07 38 PM

- Makes source URL required.
Screen Shot 2020-05-11 at 2 07 46 PM

### DIFF
--- a/verification/curator-service/ui/src/components/LinelistTable.tsx
+++ b/verification/curator-service/ui/src/components/LinelistTable.tsx
@@ -1,5 +1,6 @@
 import MaterialTable from 'material-table';
 import Paper from '@material-ui/core/Paper';
+import TextField from '@material-ui/core/TextField';
 import React from 'react';
 import axios from 'axios';
 import { isUndefined } from 'util';
@@ -93,6 +94,9 @@ export default class LinelistTable extends React.Component<{}, LinelistTableStat
 
     addCase(newRowData: TableRow) {
         return new Promise((resolve, reject) => {
+            if (!this.validateRequired(newRowData.source_url)) {
+                return reject();
+            }
             const newCase = this.createCaseFromRowData(newRowData);
             const response = axios.post(this.state.url, newCase);
             response.then(() => {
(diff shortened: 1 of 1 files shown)


allysonjp715:

allysonjp715:

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -124,6 +128,9 @@ export default class LinelistTable extends React.Component<{}, LinelistTableStat
             if (isUndefined(oldRowData)) {
                 return reject();
             }
+            if (!this.validateRequired(newRowData.source_url)) {
+                return reject();
```
attwad, line 132: Hmm a lot of existing data doesn't have a source, that will make us unable to add it if we wanted to.

@khmoran should source be required?

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -136,17 +143,33 @@ export default class LinelistTable extends React.Component<{}, LinelistTableStat
         });
     }
 
+    validateRequired(field: String) {
+        return field !== "";
```
attwad, line 147: trim before checking for length?

verification/curator-service/ui/src/components/LinelistTable.tsx
allysonjp715: Ah true. I added the validator because the API call failed without a source. I think it makes sense that all fields that are manually added from the UI should require a source.

verification/curator-service/ui/src/components/LinelistTable.tsx
allysonjp715: Done

attwad:
seems like tests are unhappy with trim, added Kelly as an explicit reviewer for the source_url requirement, code looks good otherwise.

verification/curator-service/ui/src/components/LinelistTable.tsx
khmoran: My thinking was that all new data should require a source -- but not existing data. So the mongodb validator schema doesn't require it, but Mongoose should on insert, IMO.

khmoran: But also if any of the current validation is impeding development, happy to ice it for now and bring it back when things are more stable.

PR #121, opened 2020-05-11
Setup login redirects to work in dev

attwad:
Those won't be required in production where we can just use /auth/google but in dev we have to fully qualify them to work.


attwad:

PR #122, opened 2020-05-11
Move tests for dataserver controllers into controllers subdir

khmoran:


khmoran:
[Separate JS code into classes with some object-oriented practices.](https://trello.com/c/rywyPfXt/92-separate-js-code-into-classes-with-some-object-oriented-practices)

khmoran:
adding @axmb since I know it's late for @attwad

khmoran:
> LGTM. I think this organization works -- having a distinct test folder makes excluding the files, e.g. from TS compilation, simple. The other option is to just place test files alongside source files; this avoids the need to sync dir structure between src/test, at the expense of more regex in tsconfig (and perhaps other places).

happy to do this for new but review the structure later if we're not feeling this

PR #125, opened 2020-05-11
Move the db setup from from scripts/import-data to scripts/setup-db

khmoran:
This script doesn't actually import data, so renaming it to reflect what it actually does!


khmoran:

PR #128, opened 2020-05-11
Add a dev script to set up the local cases db and populate it with sample data

khmoran:
(Also a small fix to tsconfig in the scripts dir that I noticed when testing the new script)

TESTED=Dropped my covid19 database, ran the script, and confirmed that:

- covid19 db exists
- it has a cases collection
- the collection has a validator
- the collection has 100 documents in it


khmoran:
@attwad LMK if you think we should call this script from `run_stack.sh` -- on the one hand it would ensure that everyone's db was up to date, but on the other hand it drops data (and it's easy enough to get into the habit of doing `setup-db.sh && run_stack.sh`)

khmoran:

PR #130, opened 2020-05-12
fix docker build with new context

attwad:
I noticed the build was *SLLLOOOOWWWW* this morning on my computer, it turns out that .dockerignore files are read from the build context, not from where the Dockerfile is. So we would send all node_modules/ directories to the docker builder daemon for each build which was gigs of data for nothing, now we only send a dozen mb ax and builds are fast again.

I also used a .env which is supported natively by docker-compose to store oauth secrets, I'll send you the secrets so that you can enable login once Gaurav send them to me.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,15 @@
+# verification/curation/(ui|api)
+*/*/*/.env
+*/*/*/coverage
+*/*/*/dist
+*/*/*/node_modules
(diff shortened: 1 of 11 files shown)


attwad:
FYI @axmb I'll submit that now because it makes it really tedious to work on other changes, let me know if you have any comments and I'll address them in a different PR.

PR #134, opened 2020-05-12
Fix null reference bug with source validator

khmoran:


data-serving/data-service/test/model/source.test.ts
```diff
@@ -27,4 +29,17 @@ describe('validate', () => {
...
+
+    it('validators work for embedded sources', async () => {
+        const FakeModel = mongoose.model(
+            'FakeDocument',
+            new mongoose.Schema({
+                source: sourceSchema,
+            }),
+        );
+        return new FakeModel({ source: {} }).validate((e) => {
+            console.log(e);
```
allysonjp715, line 41: Remove?

PR #137, opened 2020-05-12
Update schema to work with documents created by dataserver

khmoran:
Was debugging why the validator on mongoose was rejecting documents created from curator UI -- turns out our server is adding fun new fields I didn't expect, and the additionalProperties field was getting angry.

TL;DR: Sub-documents have their own _id fields, and the case doc gets a version key field: https://mongoosejs.com/docs/guide.html#versionKey

Now if I run:

`dev/run_stack.sh`
`dev/setup_db.sh`

and insert/update a row via the curator CRUD UI, it works!

### DIFF
--- a/data-serving/data-service/schemas/cases.schema.json
+++ b/data-serving/data-service/schemas/cases.schema.json
@@ -10,10 +10,16 @@
         "bsonType": "object",
         "additionalProperties": false,
         "properties": {
+          "_id": {
+            "bsonType": "objectId"
+          },
           "ageRange": {
             "bsonType": "object",
             "additionalProperties": false,
             "properties": {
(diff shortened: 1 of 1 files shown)


khmoran:
> This looks reasonable (thanks for sharing the background you found). Is there a test that feasibly catches this? Is it something we can catch by modifying our json `minimalCase`? Or is this issue sufficiently one-and-done?

it's tricky -- right now I guess i'm treating minimalCase/fullySpecifiedCase as what we get from the client, as opposed to what the server inserts into the db... but I will look into seeing if I can do some kind of integration test that involves validating a doc with the in-memory DB. Adding it to potential tasks for this sprint

PR #138, opened 2020-05-13
Split location schema & document into its own file

khmoran:


khmoran:
[Separate JS code into classes with some object-oriented practices.](https://trello.com/c/rywyPfXt/92-separate-js-code-into-classes-with-some-object-oriented-practices)

data-serving/data-service/src/model/location.ts
```diff
@@ -0,0 +1,82 @@
...
+        return (
+            this != null &&
+            this.country == null &&
+            this.administrativeAreaLevel1 == null &&
+            this.administrativeAreaLevel2 == null &&
+            this.locality == null
+        );
+    },
+    'One of location.country, location.administrativeRegion1, ' +
+        'location.administrativeRegion2, or locality is required',
```
axmb, line 14: Which schema do these prefixes refer to (i.e. why isn't locality prefixed with `location.`)?

khmoran:

data-serving/data-service/src/model/location.ts
khmoran: because i am bad at typinggg -- going to remove the prefixes, it's confusing

PR #139, opened 2020-05-13
add docker config to serve UI from curator express server

attwad:
I tried doing this using the existing docker files for both curator API and UI but didn't manage to compose them nicely.
Having another dockerfile at the `curator/` directory level makes that trivial.

I added a prod/ directory similar to the dev/ one to run the stack.
I didn't use another override like `docker-compose.prod.yml` as we can't "remove" services from the template (the curatorui service and mongo will be gone from prod compose setup).
Right now it still has the mongo image so I could test e2e but we should change that to talk to our atlas instance (I'm asking Kelly for creds).
I am hoping we cna use that /prod directory as a starter point to put stuff on amazon fargate using ecs-cli and compose files. (cf this [tuto](https://docs.aws.amazon.com/AmazonECS/latest/developerguide/ecs-cli-tutorial-fargate.html))

One last change is to only setup the react proxy if REACT_APP_PROXY_URL is specified, it's unused in prod anyway.


attwad:
[Tune docker images for production serving](https://trello.com/c/89a4rRPQ/88-tune-docker-images-for-production-serving)

attwad:
As per Slack discussion, this is ready for review.

.github/workflows/docker.yml
```diff
@@ -11,20 +11,37 @@ on:
...
-      - name: Test Data Service
+      - name: Test Data Service cases
         run: docker-compose -f dev/docker-compose.yml exec -T data curl --retry-connrefused --retry 20 --retry-delay 5 -f --silent --show-error http://localhost:3000/api/cases
+  test-prod:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build stack
+        run: docker-compose -f prod/docker-compose.yml -f prod/docker-compose.ci.yml up --build -d -V
+      # Containers are run detached above, which means we have to wait until they are up to produce meaningful logs.
```
axmb, line 37: Is running detached necessary? E.g., will the action block on the process if it isn't in the background.

Sleep solution feels like it's bound to flake. But if it's the simplest solution for now, we can come back to it later.

.github/workflows/docker.yml
```diff
@@ -11,20 +11,37 @@ on:
...
+      - name: Sleep for 30 seconds
+        uses: jakejarvis/wait-action@master
+        with:
+          time: "30s"
+      - name: Print logs
+        run: docker-compose -f prod/docker-compose.yml logs
+      - name: Fetch Curator Service index page
```
axmb, line 44: Is this the best "test" we can get in here? Seems like what we'd want to fail on from a docker action is a non-zero build code. I'm surprised this is our pass/fail condition.

verification/curator-service/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+# Build the UI
+WORKDIR /usr/src/app/verification/curator-service/ui
+# Install app dependencies
+COPY verification/curator-service/ui/package*.json ./
+# RUN npm clean install
+RUN npm ci
+# Bundle app source
+COPY verification/curator-service/ui/. .
+# Bundle common files
+COPY common/* ../../../common/
```
axmb, line 31: nit -- if both services are in the same Dockerfile, don't need to run this twice

prod/docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+services:
+  mongo:
+    image: mongo
+    restart: always
+    init: true
+    ports:
+      - "27017:27017"
+    environment:
+      MONGO_INITDB_DATABASE: "covid19"
+  curator:
```
axmb, line 11: Do we want to use this approach for the dev stack, too?

verification/curator-service/api/src/util/validate-env.ts
```diff
@@ -39,5 +40,9 @@ export default function validateEnv(): Readonly<{
             desc: 'URL to redirect to after the oauth consent screen',
             devDefault: 'http://localhost:3002/',
         }),
+        STATIC_DIR: str({
+            desc: 'directory to serve static files from',
+            default: '',
```
axmb, line 45: Is this necessary?

.github/workflows/docker.yml
```diff
@@ -11,20 +11,37 @@ on:
...
-      - name: Print Logs
+      - name: Print logs
         run: docker-compose -f dev/docker-compose.yml logs
-      - name: Test Curator Service
+      - name: Fetch Curator Service sources
         run: docker-compose -f dev/docker-compose.yml exec -T curator curl --retry-connrefused --retry 20 --retry-delay 5 -f --silent --show-error http://localhost:3001/api/sources
-      - name: Test Data Service
+      - name: Test Data Service cases
         run: docker-compose -f dev/docker-compose.yml exec -T data curl --retry-connrefused --retry 20 --retry-delay 5 -f --silent --show-error http://localhost:3000/api/cases
+  test-prod:
```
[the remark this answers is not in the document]
attwad: Nice! thanks for letting me know.

.github/workflows/docker.yml
attwad: Yes each step must return a non 0 return code otherwise the next ones do not run this is why it is detached, I haven't experienced a flake so far so I'm hopeful, the wait is really for the logs to print something meaningful (we could also remove the logs step tbh, it helped me debug stuff when I was first setting things up but now it's not that useful)

.github/workflows/docker.yml
attwad: A failed build step will trigger a whole workflow failure so we're testing that already actually, my plan is to remove those curl tests once we have a proper "integration test" that's hopefully coming this sprint :)

verification/curator-service/Dockerfile
attwad: good catch, removed.

prod/docker-compose.yml
attwad: sorry what approach? Using this image for the curator service you mean?
We could but it would be wasteful as we'd still need to build the UI separately (ending up building it twice) and start it with the react tool dev scripts.
Unless I misunderstood your question? sorry please clarify if so.

verification/curator-service/api/src/util/validate-env.ts
attwad: thanks that's actually a devDefault that I wanted here, in dev we use the reloadable react app and do not serve static files.

prod/docker-compose.yml
axmb: Yeah, using the single image. Plus side is that our dev testing would more closely reflect the prod configuration. Why do we need to still build the UI separately in the dev compose set up?

attwad: You certainly don't notice that on your überpowerful machine but for me it takes a few minutes to do an initial build (incremental builds are fine) ;-)
Building the images separately in dev allows us to not produce an (unused) optimized build of the UI.
BUT it also has the nice effect of allowing us to test the prod UI in with the dev curator service without having to run prod/run_stack.sh ...
So yeah, meh, I can deal with the longer initial build times I think.
Done.

PR #141, opened 2020-05-14
Barebones /sources curator UI page

axmb:
Breaking this story up to keep PR length sane. For now, just adding a minimal `/sources` page and `SourceTable` component; largely analogous to a simplified `LinelistTable`. The existing `Source` model isn't exactly what Tim and I landed on, but I'll address that after the bulk of the interface boilerplate is taken care of. Still adding the types to the component file to be "correct," though we could technically get away with just defining `_id`/`name`.

### DIFF
--- a/verification/curator-service/ui/src/components/SourceTable.tsx
+++ b/verification/curator-service/ui/src/components/SourceTable.tsx
@@ -40,7 +40,6 @@ interface Source {
 }
 
 interface SourceTableState {
-    tableRef: any,
     url: string,
 }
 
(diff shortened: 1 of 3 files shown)


verification/curator-service/ui/src/components/SourceTable.tsx
```diff
@@ -0,0 +1,96 @@
...
+interface SourceTableState {
+    tableRef: any,
+    url: string,
+}
+
+export default class SourceTable extends React.Component<{}, SourceTableState> {
+    constructor(props: any) {
+        super(props);
+        this.state = {
+            tableRef: React.createRef(),
```
allysonjp715, line 51: This (and the other tableRefs) can be removed. I'm removing it from the other table in an in-progress PR because it turned out to be unnecessary.

axmb: Excellent, thanks.

axmb:

PR #142, opened 2020-05-14
Factor demographics schema & document into own file

khmoran:


khmoran:

data-serving/data-service/src/model/demographics.ts
```diff
@@ -0,0 +1,41 @@
...
+
+export enum Species {
+    HomoSapien = 'Homo sapien',
+}
+
+export const demographicsSchema = new mongoose.Schema({
+    ageRange: {
+        start: {
+            type: Number,
+            min: -1,
```
axmb, line 18: Why would we include -1? And if we're constituting -1 as valid, shouldn't it be valid in `end`?

data-serving/data-service/test/model/demographics.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    demographicsSchema,
+);
+
+describe('validate', () => {
+    it('a start age under -1 is invalid', async () => {
+        return new Demographics({ ageRange: { start: -1.1 } }).validate((e) => {
+            expect(e.name).toBe(Error.ValidationError.name);
+        });
+    });
+    it('a start age over 300 is invalid', async () => {
```
axmb, line 20: The limit is 200, though I think it'd be reasonable just to pick some obviously-too-large age rather than focus on the exact cut-off.

khmoran: oops, the limit should be 300. was supposed to update this to be in sync with the schema definition (keeping these in sync is hard)

300 was Tim's suggestion to support other species

data-serving/data-service/src/model/demographics.ts
khmoran: -1 is for prenatal... I don't think there are any cases of this in the current line list but it was a suggestion from Tim to support it.

Fair that end should support it too, if there's a specific negative number in there.

PR #143, opened 2020-05-14
Display API errors in the UI

allysonjp715:
Screen Shot 2020-05-14 at 11 03 49 AM

### DIFF
--- a/verification/curator-service/ui/src/components/LinelistTable.test.tsx
+++ b/verification/curator-service/ui/src/components/LinelistTable.test.tsx
@@ -59,6 +59,57 @@ it('loads and displays cases', async () => {
   expect(items).toBeInTheDocument();
 })
 
+it('API errors are displayed', async () => {
+  const cases = [
+    {
+      _id: 'abc123',
+      importedCase: {
+        outcome: 'Recovered',
+      },
+      location: {
+        country: "France",
+      },
+      events: [
+        {
+          name: 'confirmed',
+          dateRange: {
+            start: new Date().toJSON(),
+          },
+        }
+      ],
+      notes: "some notes",
+      source: {
+        url: "http://foo.bar",
+      }
+    },
+  ];
+  const axiosResponse = {
+    data: {
+      cases: cases,
+      total: 15,
+    },
+    status: 200,
+    statusText: 'OK',
+    config: {},
+    headers: {},
+  };
+  mockedAxios.get.mockResolvedValueOnce(axiosResponse);
+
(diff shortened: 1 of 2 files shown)


allysonjp715:
[Display errors in Curator UI](https://trello.com/c/OW1PduoM/137-display-errors-in-curator-ui)

verification/curator-service/ui/src/components/LinelistTable.test.tsx
```diff
@@ -59,6 +59,57 @@ it('loads and displays cases', async () => {
...
+    statusText: 'OK',
+    config: {},
+    headers: {},
+  };
+  mockedAxios.get.mockResolvedValueOnce(axiosResponse);
+
+  const { getByText, findByText } = render(<LinelistTable />)
+
+  // Throw error on add request.
+  mockedAxios.post.mockRejectedValueOnce(new Error("Request failed"));
```
attwad, line 101: nit but wouldn't the test be simpler and test the same thing if the initial get response was returning the error?

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -146,83 +165,91 @@ export default class LinelistTable extends React.Component<{}, LinelistTableStat
...
+                        }}
+                        editable={{
+                            onRowAdd: (newRowData: TableRow) => this.addCase(newRowData),
+                            onRowUpdate: (newRowData: TableRow, oldRowData: TableRow | undefined) =>
+                                this.editCase(newRowData, oldRowData),
+                            onRowDelete: (rowData: TableRow) => this.deleteCase(rowData),
+                        }}
+                    />
+                </Paper>
+                <div className={classes.error}>{this.state.error}</div>
```
attwad, line 249: I suggest conditional rendering here to avoid that div if there is no error.

```
{this.state.error && <div className={classes.error}>{this.state.error}}
```

https://reactjs.org/docs/conditional-rendering.html#inline-if-with-logical--operator

allysonjp715: Cool! Done.

verification/curator-service/ui/src/components/LinelistTable.test.tsx
allysonjp715: Yes that was my initial thought, but the table rendering is not failing gracefully when the get errors out. We should fix this, but in a different PR IMO.

attwad: ah ok good to know (and to fix), lgtm

PR #144, opened 2020-05-14
basic setup for ecs fargate cluster running data service only

attwad:
This mostly reverts my previous PR as I work through things in small steps.

You can see the cluster with the dataservice task running in it here if you want
https://console.aws.amazon.com/ecs/home?region=us-east-1#/clusters/epid-dev/services


attwad:
[Run data service on Fargate connected to dev atlas mongo DB](https://trello.com/c/N5gybRoB/149-run-data-service-on-fargate-connected-to-dev-atlas-mongo-db)

aws/docker-compose.yml
```diff
@@ -0,0 +1,17 @@
+version: "3"
```
axmb, line 1: Did you intend to revert the prod/ dir in this PR?

aws/README.md
```diff
@@ -0,0 +1,55 @@
...
+aws configure
+```
+
+Configure ecs-cli to work on our cluster:
+
+```shell
+ecs-cli configure --cluster epid-dev --default-launch-type FARGATE --region us-east-1
+```
+
+Create the cluster:
```
axmb, line 21: Er, isn't this already done? Can we cut the stuff that's the completed one-time-setup?

aws/docker-compose.yml
attwad: yes, it had no benefit now that we're using the same immage for our dev stack.

attwad: ah sorry I just realized this wasn't done, somehow I cp instead of mv... my bad. Now done.

aws/README.md
attwad: No this is weird but if you look at the help for the configure command it tells you that resources aren't created until you issue the up command

```
...
   --cluster value, -c value            Specifies the ECS cluster name to use. If the cluster does not exist, it is created when you try to add resources to it with the ecs-cli up command. [$ECS_CLUSTER]
```

attwad:
> Shouldn't the AWS service/task name be data-service? The cluster name of epid-dev SGTM.

I tried changing that and managed to do so using --project-name data when upping the service but we'll have to see how to do this properly when multiple services are defined in the docker compose file...

PR #145, opened 2020-05-14
Split travel schema & document into own file

khmoran:
There are a few commits in this -- prob worth stepping through them to understand the large-ish file diff

Basically, I want to be able to have canonical full/minimal json examples for each schema so that other schemas that embed them can reuse them -- it'd be a pain to update a whole bunch of tests if one sub-document changes


khmoran:

data-serving/data-service/test/model/date-range.test.ts
```diff
@@ -0,0 +1,19 @@
...
+    'DateRange',
+    dateRangeSchema,
+);
+
+describe('validate', () => {
+    it('an empty date range document is valid', async () => {
+        return new DateRange({}).validate();
+    });
+
+    it('a fully specified date range document is valid', async () => {
```
axmb, line 16: Minor thought: do we want a minimal (half-open) range?

data-serving/data-service/src/model/travel.ts
```diff
@@ -0,0 +1,32 @@
...
+        location: locationSchema,
+        dateRange: dateRangeSchema,
+        purpose: {
+            type: String,
+            enum: Object.values(TravelPurpose),
+        },
+        additionalInformation: String,
+    },
+    {
+        strict: true,
```
axmb, line 23: Do we have this specified `false` or `throw` at some top-level point? `true` is the default value

data-serving/data-service/test/model/travel.test.ts
```diff
@@ -0,0 +1,31 @@
+import { TravelDocument, travelSchema } from '../../src/model/travel';
```
axmb, line 1: Planning to json this, like the other ones?

data-serving/data-service/test/model/date-range.test.ts
khmoran: done

data-serving/data-service/src/model/travel.ts
khmoran: nope -- removed (here & elsewhere)

data-serving/data-service/test/model/travel.test.ts
khmoran: sure -- I'll do it here

PR #146, opened 2020-05-14
Make sources table editable; add deletion.

axmb:
Planned to add full CRUD here, but will need to change the table row model (as in Linelist table) first. So just deletion for now, and full CRUD in a subsequent PR.

Lots of similar boilerplate here to the LinelistTable. I'm planning to continue on with two separate structures, but will see if we can share a generalized class between the two afterwards.


allysonjp715:
What do you mean by "need to change the table row model (as in Linelist table)"? 

These changes LGTM. I don't think it's worth trying to share anything with the Linelist table if they are going to stay as separate tables. 

BTW see the slack for the test-prod failure - syncing should get rid of it.

axmb:
> What do you mean by "need to change the table row model (as in Linelist table)"?

Sorry: creating a dedicated row type in order to flatten the data.

axmb:

PR #153, opened 2020-05-15
Add a cypress integration test for the linelist table

allysonjp715:
This definitely needs a lot of refinement, but achieves putting the cypress infrastructure in place.

Right now you must run the stack locally first, and then run the tests against the local server. In a future PR I will look to add this to our CI tests where it starts its own server before running. But for local development it's much easier to run your own local stack than to have the test start one up.

I left some example code + comments in while everyone is getting acquainted with Cypress.

### DIFF
--- a/verification/curator-service/ui/cypress/fixtures/example.json
+++ b/verification/curator-service/ui/cypress/fixtures/example.json
@@ -1,0 +1,5 @@
+{
+  "name": "Using fixtures to represent data",
+  "email": "hello@cypress.io",
+  "body": "Fixtures are a great way to mock data for responses to routes"
+}
--- a/verification/curator-service/ui/cypress/integration/components/LinelistTableTest.spec.js
+++ b/verification/curator-service/ui/cypress/integration/components/LinelistTableTest.spec.js
@@ -1,0 +1,9 @@
+describe('Linelist table', function () {
+    it('Can edit a case', function () {
+        cy.visit('/cases');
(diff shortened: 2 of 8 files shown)


allysonjp715:

verification/curator-service/ui/cypress/fixtures/example.json
```diff
@@ -0,0 +1,5 @@
+{
```
khmoran, line 1: do we need to commit this file?

FWM if this was generated by a command and we'll clean up example stuff later

allysonjp715: Yes, twas "generated by a command and we'll clean up example stuff later"

verification/curator-service/ui/cypress/integration/components/LinelistTableTest.spec.js
```diff
@@ -0,0 +1,9 @@
+describe('Linelist table', function () {
```
attwad, line 1: cypress apparently ships with typescript support, can we stick to typescript and avoid javascript please?

https://docs.cypress.io/guides/tooling/typescript-support.html#Install-TypeScript

allysonjp715: Ah, thanks for catching, just an oversight. Fixed in https://github.com/open-covid-data/healthmap-gdo-temp/pull/158

PR #157, opened 2020-05-15
Extract the origin sub-type from the top-level Source model

axmb:


axmb:
[Implement curator service CRUD API for automated source scraping runs](https://trello.com/c/LOOSItyi/80-implement-curator-service-crud-api-for-automated-source-scraping-runs)

PR #159, opened 2020-05-18
Run integration tests in CI

allysonjp715:
Requires changing the test to an add rather than edit test, because the stack that's brought up with docker-compose doesn't have any existing rows in the table. Will clean up this discrepancy between local / CI in a future PR.  

Will also remove the docker.yaml file once we add an integration test for /sources.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,6 +1,3 @@
-# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
-
 name: Integration Tests CI
 
 on:
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -18,19 +18,17 @@ on:
       - "data-serving/data-service/**"
 
 jobs:
-  build:
+  integration-tests:
     runs-on: ubuntu-latest
 
     steps:
       - uses: actions/checkout@v2
-      - name: Use Node.js
-        uses: actions/setup-node@v1
-        with:
-          node-version: "12.x"
+      - name: Build stack
+        run: docker-compose -f dev/docker-compose.yml -f dev/docker-compose.ci.yml up --build -d -V
       - name: Cypress integration tests
         run: |
           npm ci
-          npm run cypress-ci-test
+          npm run cypress-run
         working-directory: verification/curator-service/ui
         env:
           CI: true
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -1,0 +1,34 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: Integration Tests CI
+
+on:
+  push:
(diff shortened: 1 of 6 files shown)


verification/curator-service/api/src/util/validate-env.ts
```diff
@@ -26,10 +26,12 @@ export default function validateEnv(): Readonly<{
         PORT: port({ default: 3001 }),
         GOOGLE_OAUTH_CLIENT_ID: str({
             desc: 'OAuth client ID from the Google developer console',
+            default: 'replace to enable auth',
```
attwad, line 29: no please outside of dev we have to make sure we put in the right values here, same below

verification/curator-service/ui/package.json
```diff
@@ -26,11 +26,14 @@
...
     "start": "PORT=3002 react-scripts start",
+    "start-full-stack": "./../../../dev/run_stack.sh",
     "build": "react-scripts build",
     "test": "react-scripts test",
     "test-silent": "react-scripts test --watchAll=false",
     "eject": "react-scripts eject",
-    "cypress": "cypress open"
+    "cypress-open": "cypress open",
+    "cypress-run": "cypress run",
+    "cypress-ci-test": "WAIT_ON_TIMEOUT=600000 start-server-and-test start-full-stack http://localhost:3002 cypress-run"
```
attwad, line 36: building and starting the stack should be a github actions step, cf how the current docker workflow work:
https://github.com/open-covid-data/healthmap-gdo-temp/blob/master/.github/workflows/docker.yml#L18

once you've done that then you can npm run cypress-ci-test that points to localhost:3002, no need for start-server-and-test and start-full-stack scripts anymore

verification/curator-service/api/src/util/validate-env.ts
allysonjp715: Done

verification/curator-service/ui/package.json
allysonjp715: Ahh thanks! Done.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,34 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
```
attwad, line 1: remove stale comment or comment about cypress/usage of this workflow?

dev/test_all.sh
```diff
@@ -2,4 +2,5 @@
 set -e
 npm --prefix=`dirname "$0"`/../verification/curator-service/api/ run-script test-silent
 npm --prefix=`dirname "$0"`/../data-serving/data-service/ run-script test-silent
-npm --prefix=`dirname "$0"`/../verification/curator-service/ui/ run-script test-silent
\ No newline at end of file
+npm --prefix=`dirname "$0"`/../verification/curator-service/ui/ run-script test-silent
+npm --prefix=`dirname "$0"`/../verification/curator-service/ui/ run-script cypress-ci-test
```
attwad, line 6: s/cypress-ci-test/cypress-run ?

allysonjp715:

.github/workflows/integration-tests.yml
allysonjp715: Removed

dev/test_all.sh
allysonjp715: Good catch. Also added comment that local stack is necessary for that command.

PR #162, opened 2020-05-18
Configure our React code to use the ~same eslint/prettier config as other dirs

axmb:
I've separated this into three commits;:

1. The config.
2. Tons of (mostly auto) changes made by running `npm run lint`.
3. `.eslintignore`, which mostly serves to prevent errors from `ui/build/`.

I'd probably skip reviewing the second, which is mostly spacing/semicolon/quotation changes.

Note that the config isn't an exact copy of what exists in other directories, because a few changes are required to interoperate well with CRA/React. Per [this page](https://create-react-app.dev/docs/setting-up-your-editor/#experimental-extending-the-eslint-config), you'll need to create a `[...]/ui/.env` file with `EXTEND_ESLINT=true` if you want this to be used as part of the CRA scripts (i.e. if you want it to stop you from building and such). You don't have to do this to get editor integration.

We might wish to run this as part of commit (via husky/lint-staged), but not doing that here.

Note that I didn't resolve all warnings (I'll do that separately).


axmb:

PR #163, opened 2020-05-18
Clear DB at start of each i-test, and reseed DB at the end of all i-tests.

allysonjp715:


allysonjp715:

verification/curator-service/ui/cypress/plugins/index.ts
```diff
@@ -0,0 +1,19 @@
+// Need to use require in plugins file 
+// https://stackoverflow.com/questions/59743577/is-there-anyway-to-query-mongodb-in-cypress-test
+const MongoClient = require('mongodb').MongoClient;
+const url = "mongodb://localhost:27017/";
```
khmoran, line 4: is there any way to reuse the connection data defined in the .env file? this could get out of sync

verification/curator-service/ui/cypress/plugins/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+module.exports = (on: any, config: any) => {
```
khmoran, line 6: what's the advantage of doing this as opposed to calling the existing db setup script with `cy.exec('npm run...')`? the db-setup script will also make sure that the right schema is applied, and indexes when we have them (imaging a scenario where someone's testing changes to the validator and it gets out of sync)

(then you also wouldn't have to worry about which db to connect to above)

verification/curator-service/ui/cypress/plugins/index.ts
allysonjp715: It seems the port is hardcoded here: https://github.com/open-covid-data/healthmap-gdo-temp/blob/master/dev/docker-compose.yml

But good news is that if the port becomes out of sync the test fails with a clear error message (couldn't connect to local mongo). I will add that file to the integration tests yml change detection so if that file changes then the integration tests will run on CI.

verification/curator-service/ui/cypress/plugins/index.ts
allysonjp715: Worked with Kelly offline. The `npm run setup-db` command takes ~7 seconds to run. We don't want to add 7s to each test run, but that is okay for initial testing setup. So I changed the code to run setup-db at the start of the entire test suite, which will init the db with new schema and indexes. Then on each test run it will make this call which runs in <1s to remove the cases.

.github/workflows/integration-tests.yml
```diff
@@ -5,12 +5,14 @@ on:
     branches: [master]
     paths:
       - ".github/workflows/integration-tests.yml"
+      - "dev/docker-compose.yml"
       - "verification/curator-service/**"
       - "data-serving/data-service/**"
```
khmoran, line 10: kinda annoying how I set up this dir structure but we might also want to test on

data-serving/samples

in case someone (me??) updates the sample cases.json and it no longer imports successfully with the schema

allysonjp715: Good catch, done.

PR #165, opened 2020-05-18
Display API errors on the sources page

axmb:


verification/curator-service/ui/src/components/SourceTable.test.tsx
```diff
@@ -58,6 +60,53 @@ it('loads and displays sources', async () => {
     expect(items).toBeInTheDocument();
 });
 
+it('API errors are displayed', async () => {
+    const sources = [
```
khmoran, line 64: worth reusing a json file for this, since exact input doesn't matter?

axmb: Yeah -- this is a good idea. There's a ton of LoC tied up in these structures in this and the LinelistTable test. Even the cases that use data attributes could reference fields on loaded objects. Let's do this in a separate PR.

axmb:

PR #166, opened 2020-05-19
Don't include lossless fields in the importedCases field to save disk space

khmoran:
Removes some redundant data and saves some space.

Also to reduce bloat (would've done in a separate PR but too annoying):

- Removes the species field (I'm guessing we can assume human unless otherwise specified -- on the fence about whether to explicitly include)
- Only populates the pathogens field when there's sequence source data (i.e. not just the name of the covid pathogen)

We're still over quota by a bit after the latest cases are published; should figure out what to do there separately (start using prod?)

TESTED: Applied the new schema and data imported without validation errors.

```
2020-05-18T22:06:06.613-0400    Failed: (AtlasError) you are over your space quota, using 516 MB of 512 MB
2020-05-18T22:06:06.613-0400    777000 document(s) imported successfully. 0 document(s) failed to import.
```

### DIFF
--- a/data-serving/data-service/schemas/cases.schema.json
+++ b/data-serving/data-service/schemas/cases.schema.json
@@ -370,9 +370,6 @@
           "province": {
             "bsonType": "string"
           },
-          "country": {
-            "bsonType": "string"
-          },
           "geo_resolution": {
             "bsonType": "string"
           },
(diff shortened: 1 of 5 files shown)


khmoran:
[Line list import: Don't "archive" fields that are 100% losslessly imported](https://trello.com/c/yTDW6VcN/154-line-list-import-dont-archive-fields-that-are-100-losslessly-imported)

PR #170, opened 2020-05-19
Authorization

iamleeg:
Draft and definitely an RFC at the moment to see if I'm going in the right direction; I have a fully-tested middleware and haven't integrated it yet. Thoughts:

 - is the format of the RBAC configuration a good one? It uses regexes, which might be over-complex to use, but what alternative is easier _and_ sufficiently flexible?
 - where do we want to store the config object? In a kubernetes secret? In MongoDB?
 - when I integrate this middleware it'll break all of the existing `supertest` integration tests. I'm expecting that the fix is to configure things until those tests pass again (i.e. add access control configuration, and give the test user a role), rather than to make any test-only trapdoors in the RBAC. Is that right?
 - I've imported `ramda` for exactly one use, is that acceptable? What's a good alternative?

### DIFF
--- a/verification/curator-service/api/src/controllers/access.ts
+++ b/verification/curator-service/api/src/controllers/access.ts
@@ -1,12 +1,12 @@
 import { intersection } from 'ramda';
 import { NextFunction } from 'express';
 
-const mustHaveRoles = (roles: Array<string>, req: any, res: any, next: NextFunction) => {
+const mustHaveRoles = (roles: Array<string>) => ((req: any, res: any, next: NextFunction) => {
   if (intersection(roles, req.user.roles).length > 0) {
     next();
   } else {
     res.sendStatus(403);
   }
-}
+});
 
 export default mustHaveRoles;
--- a/verification/curator-service/api/src/controllers/access.ts
+++ b/verification/curator-service/api/src/controllers/access.ts
@@ -1,8 +1,16 @@
-import { intersection } from 'ramda';
 import { NextFunction } from 'express';
 
-const mustHaveRoles = (roles: Array<string>) => ((req: any, res: any, next: NextFunction) => {
-  if (req.user && req.user.roles && intersection(roles, req.user.roles).length > 0) {
+function setContainsAnyOf<T>(set: Set<T>, possibleElements: [T]): boolean {
+  for(let e of possibleElements) {
+    if (set.has(e)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+const mustHaveRoles = (roles: Set<string>) => ((req: any, res: any, next: NextFunction) => {
+  if (req.user && req.user.roles && setContainsAnyOf(roles, req.user.roles)) {
     next();
   } else {
     res.sendStatus(403);
--- a/verification/curator-service/api/src/controllers/access.ts
+++ b/verification/curator-service/api/src/controllers/access.ts
@@ -1,4 +1,4 @@
-import { curry } from 'ramda';
+import { curry, intersection } from 'ramda';
 
 const accessControl = (
   config: Object,
@@ -22,7 +22,7 @@ const accessControl = (
       }
     }
     const expectedRoles  = methodRoleMap[method];
-    if (expectedRoles.indexOf(req.user.role) !== -1) {
+    if (intersection(expectedRoles, req.user.roles).length > 0) {
       next();
     } else {
       res.sendStatus(403);
--- a/verification/curator-service/api/src/controllers/access.ts
+++ b/verification/curator-service/api/src/controllers/access.ts
@@ -1,33 +1,12 @@
-import { curry, intersection } from 'ramda';
+import { intersection } from 'ramda';
+import { NextFunction } from 'express';
 
-const accessControl = (
-  config: Object,
-) => (
-  curry((config, req, res, next) => {
-    const method = req.method === 'HEAD' ? 'GET' : req.method; // HEAD and GET have the same visibility
-    let methodRoleMap : { [index: string] : Array<string> };
-    if (config[req.url]) {
-      methodRoleMap = config[req.url];
-    } else {
-      const keys = Object.getOwnPropertyNames(config).filter((prop) => (
-        req.url.match(prop)
-      ));
-      if (keys.length === 1) {
-        // exactly one configuration found, use that
-        methodRoleMap = config[keys[0]];
-      } else {
-        // zero or 2+ configurations found, either way it's bad
-        res.sendStatus(500);
-        return;
-      }
-    }
-    const expectedRoles  = methodRoleMap[method];
-    if (intersection(expectedRoles, req.user.roles).length > 0) {
-      next();
-    } else {
-      res.sendStatus(403);
-    }
-  })(config)
-);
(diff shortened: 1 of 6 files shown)


verification/curator-service/api/src/model/user.ts
```diff
@@ -14,12 +14,17 @@ const userSchema = new mongoose.Schema({
         // supporting more identity platforms.
         required: 'User must be logged-in with Google',
     },
+    role: {
```
axmb, line 17: Do we have a defined set of roles for which we can add an `enum` validator?

https://mongoosejs.com/docs/schematypes.html#string-validators

verification/curator-service/api/src/controllers/access.ts
```diff
@@ -0,0 +1,33 @@
+import { curry } from 'ramda';
+
+const accessControl = (
+  config: Object,
+) => (
+  curry((config, req, res, next) => {
```
axmb, line 6: Neat. Is this just to simplify installation? I.e., load the config in our app declaration, then `app.use` the config-loaded-function?

verification/curator-service/api/src/model/user.ts
attwad: yes let's use an enum please, we can safely assume we'll have "reader", "curator" and "admin" for now.

verification/curator-service/api/src/model/user.ts
```diff
@@ -14,12 +14,17 @@ const userSchema = new mongoose.Schema({
...
+        type: String,
+        // Note: not required as assuming no defined role === no permission
+    }
 });
 
 export type UserDocument = mongoose.Document & {
     googleID: string;
     name: string;
     email: string;
+    role: string;
```
attwad, line 27: This should be an array of roles as one can have many roles.
We could use a hierarchy of roles and assume that admin implies curator implies reader but being explicit about each role is better imo.

verification/curator-service/api/src/controllers/access.ts
```diff
@@ -0,0 +1,33 @@
+import { curry } from 'ramda';
+
+const accessControl = (
```
attwad, line 3: Why do we need that RBAC config?
I was envisioning to just be explicit about roles in the express route config defined in index.ts.
For example if we want to restrict GET /api/cases to readers we should do:

```
apiRouter.get('/cases', mustBeAuthenticated(), mustHaveRoles(["reader"]), casesController.list);
```

attwad:
> Draft and definitely an RFC at the moment to see if I'm going in the right direction; I have a fully-tested middleware and haven't integrated it yet. Thoughts:
> 
> * is the format of the RBAC configuration a good one? It uses regexes, which might be over-complex to use, but what alternative is easier _and_ sufficiently flexible?
As I commented, I am not sure why we need such a separate config VS putting the RBAC in the express router directly.
> * where do we want to store the config object? In a kubernetes secret? In MongoDB?
If we end up needing that config, it definitely doesn't belong in a kubernetes secret as there's nothing "secret" in there, MongoDB could work but just bundling it statically as json works as well.
> * when I integrate this middleware it'll break all of the existing `supertest` integration tests. I'm expecting that the fix is to configure things until those tests pass again (i.e. add access control configuration, and give the test user a role), rather than to make any test-only trapdoors in the RBAC. Is that right?
Yeah I was wondering about that, authn and authz once enforced are going to break most of our tests. I haven't looked at how to create users manually but it doesn't seem like it's going to be easy with oauth, perhaps we should have an environment variable that installs the local passport strategy so that we can register a user with a simple username/password for tests and provide test set up utilities to log in users that way. I've also read a bit about superagent vs supertest but not sure if that's going to help much.
> * I've imported `ramda` for exactly one use, is that acceptable? What's a good alternative?
The good alternative is hopefully aggreeing that we don't need that config :) and as such ramda as well ;) happy to discuss.

verification/curator-service/api/src/controllers/access.ts
iamleeg: Yes, and to avoid having a constructor and a class just to close over one parameter.

verification/curator-service/api/src/controllers/access.ts
iamleeg: Ah, interesting, I hadn't thought of that. Then there's no question of where the RBAC lives, it's in the source. I'll take that approach, thanks.

verification/curator-service/api/src/controllers/access.ts
```diff
@@ -0,0 +1,12 @@
+import { intersection } from 'ramda';
+import { NextFunction } from 'express';
+
+const mustHaveRoles = (roles: Array<string>, req: any, res: any, next: NextFunction) => {
```
iamleeg, line 4: :clown_face: I'm not happy with `req` and `res` being `any` type, but couldn't use the interface types defined by Express. For example, `req` interface doesn't have a `user` property. How do I typescript?

verification/curator-service/api/src/controllers/access.ts
iamleeg: I've made that change, it's much simpler now!

verification/curator-service/api/src/model/user.ts
```diff
@@ -14,12 +14,21 @@ const userSchema = new mongoose.Schema({
...
+            'admin',
+        ],
+    },
 });
 
 export type UserDocument = mongoose.Document & {
     googleID: string;
     name: string;
     email: string;
+    role: string;
```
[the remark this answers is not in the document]
iamleeg: oh crap yes

verification/curator-service/api/src/controllers/access.ts
```diff
@@ -0,0 +1,20 @@
+import { NextFunction } from 'express';
+
+function setContainsAnyOf<T>(set: Set<T>, possibleElements: [T]): boolean {
```
attwad, line 3: Instead of this function, intersection of two sets (roles and wantedRoles) can be:

`set1.filter(x => set2.has(x)));`

verification/curator-service/api/test/model/user.test.ts
```diff
@@ -17,4 +17,22 @@ describe('User schema', () => {
             'User must be logged-in with Google',
         );
     });
+    it('should allow a user to have a role', () => {
```
attwad, line 20: maybe just add the role to the initial `should build a valid user` test above?

verification/curator-service/api/test/model/user.test.ts
```diff
@@ -17,4 +17,22 @@ describe('User schema', () => {
...
+        expect(errors).toBeUndefined();
+    });
+    it('should restrict roles to values in the enum', () => {
+        const errors = new User({
+            name: 'test source',
+            email: 'foo@bar.com',
+            googleID: 'baz',
+            roles: ['brigadier'],
+        }).validateSync();
+        expect(errors).toBeDefined();
```
attwad, line 36: expect brigadier to be in the error string?

attwad:
I added the middleware in #188 and Allyson added the users role in #185 , we can close this PR.

iamleeg:
Thanks @attwad and @allysonjp715 for taking this over!

PR #171, opened 2020-05-19
 Update line-list schema's source and pathogens.sequenceSource fields to be a list

khmoran:
Since we right now have to do the full stack at once, it's a large diff. I tried to separate the commits into meaningful chunks.

To make it a little easier:

@axmb: Can you review the first 4 commits (schema, sample data, python converter, dataserver)
@allysonjp715 Can you review the last commit (curator UI)


khmoran:
[Line-list schema: Change source field to sources plural; parse lists](https://trello.com/c/1ee1oKMG/155-line-list-schema-change-source-field-to-sources-plural-parse-lists)

data-serving/data-service/schemas/cases.schema.json
```diff
@@ -257,21 +257,25 @@
...
+        "bsonType": "array",
+        "uniqueItems": true,
+        "items": {
+          "bsonType": "object",
+          "additionalProperties": false,
+          "properties": {
+            "_id": {
+              "bsonType": "objectId"
+            },
+            "id": {
```
axmb, line 270: Can you remind me what the distinction between the two ID fields is?

data-serving/data-service/test/model/case.test.ts
```diff
@@ -1,25 +1,35 @@
...
+        return new Case({ ...minimalModel, sources: [] }).validate((e) => {
+            expect(e.name).toBe(Error.ValidationError.name);
+        });
     });
-    it('fully-specified model should be valid', async () => {
-        const c = new Case(fullCase);
 
-        return c.validate();
+    it('model without confirmed event is invalid', async () => {
+        return new Case({ ...minimalModel, events: [] }).validate((e) => {
```
axmb, line 14: nit: this test passes regardless of the `confirmed` name checking

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -270,7 +273,11 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
                                                     ? new Date(confirmedDate)
                                                     : null,
                                                 notes: c.notes,
-                                                sourceUrl: c.source?.url,
+                                                sourceUrl:
```
axmb, line 276: Do we want a TODO here, or is it captured elsewhere, to make the UI representation a list?

data-serving/data-service/schemas/cases.schema.json
khmoran: sure -- _id is auto-generated by mongodb on insert for sub-documents

id is a reference to a document in the `sources` collection. (if we don't want to use this, can remove later!)

verification/curator-service/ui/src/components/LinelistTable.tsx
khmoran: I was assuming this would generally be overhauled, since a few things in the UI don't reflect the model anymore (like location != country, age should be a range, etc)

data-serving/data-service/test/model/case.test.ts
khmoran: good point - separated into 2 tests (empty vs. nonempty, but missing name='sources')

PR #172, opened 2020-05-19
Add tests for editing and deleting cases

allysonjp715:
Adds a custom command for adding a case through a POST request. This is much quicker than having to go through the UI each time to add a case.


verification/curator-service/ui/cypress/integration/components/LinelistTableTest.spec.ts
```diff
@@ -15,4 +16,24 @@ describe('Linelist table', function () {
         cy.get('button[title="Save"]').click();
         cy.contains('test notes');
     })
+
+    it('Can edit a case', function () {
```
khmoran, line 20: nit: might be kind of helpful to arrange the tests in arrange/act/assert blocks, just for visual scanning purposes

verification/curator-service/ui/cypress/integration/components/LinelistTableTest.spec.ts
```diff
@@ -15,4 +16,24 @@ describe('Linelist table', function () {
...
+        cy.get('input[placeholder="Notes"]').clear().type('edited notes');
+        cy.get('button[title="Save"]').click();
+        cy.contains('some notes').should('not.exist');
+        cy.contains('edited notes');
+    })
+
+    it('Can delete a case', function () {
+        cy.addCase('France', 'some notes', 'www.example.com');
+        cy.visit('/cases');
+        cy.contains('some notes');
```
khmoran, line 34: how annoying would it be to have addCase return the case id from the RPC and use that to check existence instead of the notes field? i know we're clearing the db between test, but a uid seems more foolproof than comparing notes

fine with me if you think it's overkill and/or you want to do it later

allysonjp715:

verification/curator-service/ui/cypress/integration/components/LinelistTableTest.spec.ts
allysonjp715: Done.

verification/curator-service/ui/cypress/integration/components/LinelistTableTest.spec.ts
allysonjp715: Yea that's a good idea. I had a play around with this, but to get the result of cy.addCase, you had to wrap everything else in a .then() call. So it ended up looking like:

```typescript
it('Can delete a case', function () {
    cy.addCase('France', 'some notes', 'www.example.com').then((response) => {
        const newRowId = response;

        cy.visit('/cases');
        cy.contains(newRowId);

        cy.get('button[title="Delete"]').click();
        cy.get('button[title="Save"]').click();

       cy.contains(newRowId).should('not.exist');
    });
})
```

That's less readable IMO, and could get quite messy if we start adding more than one row and have nested .then()s, so I'll leave as is for now.

PR #173, opened 2020-05-20
Prep Source schema for AWS API integration

axmb:
Need to get a handful of validations and protections in place before we hit the AWS API from our Sources API. This is the bulk of doing so, less one last change for the inclusion of a schedule expression that I'll send subsequently.

Apologies for lumping these together; in this case seemed the lesser evil.

### DIFF
--- a/verification/curator-service/api/test/model/automation.test.ts
+++ b/verification/curator-service/api/test/model/automation.test.ts
@@ -1,0 +1,59 @@
+import {
+    AutomationDocument,
+    automationParsingValidator,
+    automationSchema,
+} from '../../src/model/automation';
+
+import { Error } from 'mongoose';
+import fullModel from './data/automation.full.json';
+import regexParsingModel from './data/regex-parsing.full.json';
+import mongoose from 'mongoose';
+
+const Automation = mongoose.model<AutomationDocument>(
+    'Automation',
+    automationSchema,
+);
+
+// Used to test automationParsingValidator.
+const wrapperSchema = new mongoose.Schema({
+    automation: {
+        type: automationSchema,
+        validate: automationParsingValidator,
+    },
+});
+type WrapperDocument = mongoose.Document & { automation: AutomationDocument };
+const Wrapper = mongoose.model<WrapperDocument>('Wrapper', wrapperSchema);
+
+describe('validate', () => {
+    it('an automation without a schedule is invalid', async () => {
+        const missingSchedule = { ...fullModel };
+        delete missingSchedule.schedule;
+
+        return new Automation(missingSchedule).validate((e) => {
+            expect(e.name).toBe(Error.ValidationError.name);
+        });
+    });
+
+    it('an automation without either a parser or regexParsing is invalid', async () => {
+        const noParsing = { ...fullModel };
+        delete noParsing.parser;
+        const wrapper = { automation: noParsing };
+
+        return new Wrapper(wrapper).validate((e) => {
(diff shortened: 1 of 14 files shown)


verification/curator-service/api/test/model/parser.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    it('a parser without an AWS lambda ARN is invalid', async () => {
+        const missingArn = { ...fullModel };
+        delete missingArn.awsLambdaArn;
+
+        return new Parser(missingArn).validate((e) => {
+            expect(e.name).toBe(Error.ValidationError.name);
+        });
+    });
+
+    it('a parser with a misformated AWS lambda ARN is invalid', async () => {
```
khmoran, line 19: super nit: s/misformated/misformatted

verification/curator-service/api/test/model/parser.test.ts
```diff
@@ -0,0 +1,31 @@
...
+        const badArn = { ...fullModel };
```
khmoran, line 20: up to you but can optionally inline as:

new Parser({ ...fullModel, awsLambdaArn: 'invalid...' });

not at all necessary if you think this is clearer, i just love to inline

verification/curator-service/api/test/model/automation.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    it('an automation without a schedule is invalid', async () => {
+        const missingSchedule = { ...fullModel };
+        delete missingSchedule.schedule;
+
+        return new Automation(missingSchedule).validate((e) => {
+            expect(e.name).toBe(Error.ValidationError.name);
+        });
+    });
+
+    it('an automation without either a parser or regexParsing is invalid', async () => {
```
khmoran, line 37: should there be a similar one for the inverse? (remove regexParser but leave parser)?

verification/curator-service/api/test/model/automation.test.ts
```diff
@@ -0,0 +1,59 @@
...
+        const noParsing = { ...fullModel };
+        delete noParsing.parser;
+        const wrapper = { automation: noParsing };
+
+        return new Wrapper(wrapper).validate((e) => {
+            expect(e.name).toBe(Error.ValidationError.name);
+        });
+    });
+
+    it('an automation with both a parser and regexParsing is invalid', async () => {
```
khmoran, line 47: maybe one for w/o either a parser or regexParsing?

verification/curator-service/api/test/model/parser.test.ts
axmb: Good catch, thanks.

verification/curator-service/api/test/model/parser.test.ts
axmb: Inline'd!

verification/curator-service/api/test/model/automation.test.ts
axmb: The loaded object contains only a parser -- so what we're testing is that neither exist. The full model contains a parser but no regexParsing. I can add a test for an object with just a regexParsing.

verification/curator-service/api/test/model/automation.test.ts
axmb: Sorry, that's the one above. This isn't completely clear, since the loaded model contains only a parser (which it must, because parser/regexParsing are mutually exclusive).

PR #177, opened 2020-05-20
include tests in tsconfig

attwad:
I also let autoformat do its json thing, not liking it but at least it's consistent.


attwad:
heh, those tests pass locally... no idea why they fail here, will look at that after a meeting.

PR #178, opened 2020-05-20
Add a simple profile page

allysonjp715:
I pulled the `axios.get<User>('/auth/profile')` out into the App file because the User information was now being used in EpidNavbar and Profile.

Logged in:
Screen Shot 2020-05-20 at 3 48 01 PM

logged out: 
Screen Shot 2020-05-20 at 3 47 41 PM

### DIFF
--- a/verification/curator-service/ui/src/components/App.tsx
+++ b/verification/curator-service/ui/src/components/App.tsx
@@ -4,8 +4,10 @@ import EpidNavbar from './EpidNavbar';
 import LinelistTable from './LinelistTable';
 import React from 'react';
 import SourceTable from './SourceTable';
+import Profile from './Profile';
 import { ThemeProvider } from '@material-ui/core/styles';
 import { createMuiTheme } from '@material-ui/core/styles';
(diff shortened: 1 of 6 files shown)


verification/curator-service/ui/src/components/App.tsx
```diff
@@ -54,6 +86,8 @@ function Home(): JSX.Element {
             <br />
             <Link to="/sources">Sources</Link>
             <br />
+            <Link to="/profile">Profile</Link>
```
attwad, line 89: only render this one conditionally if there's a user in the state?

allysonjp715: Good idea. I removed the Link in Home and also removed the /profile Route so it can't be navigated to manually.

allysonjp715:
[Display roles of users in UI (complete user profile page or simply in the navbar)](https://trello.com/c/4lk6NZpO/147-display-roles-of-users-in-ui-complete-user-profile-page-or-simply-in-the-navbar)

PR #179, opened 2020-05-20
Add a skeleton/partial open API definition for the dataserver

khmoran:
- This PR just adds a (partial) service definition, and adds a documentation endpoint -- it does NOT validate requests and responses or otherwise integrate with express
- GET is the only endpoint in there right now
- The GET response is incomplete because I didn't want to flesh it out until I was sure that this approach LG

Screenshot of /api-docs endpoint:

image

### DIFF
--- a/data-serving/data-service/api/dataserver.api.yaml
+++ b/data-serving/data-service/api/dataserver.api.yaml
@@ -1,0 +1,54 @@
+openapi: 3.0.3
+info:
+  title: EPID data service
+  description: Server for CRUD operations on EPID line-list data
+  termsOfService: 'https://www.healthmap.org/about/termsofuse/'
+  contact:
+    email: info@healthmap.org
+  license:
+    name: MIT
+  version: 1.0.0
+paths:
+  /api/cases/{id}:
+    summary: Line-list cases
+    get:
+      summary: Gets a specific case
+      parameters: 
+        - in: path
+          name: id
+          required: true
+          schema:
+            type: string
+          description: The case ID
+      responses:
+        '200':
+          description: Case found.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Case'
+        '400':
+          description: Bad request.
+        '401':
+          description: Unauthorized.
+        '404':
+          description: Case not found.
+      tags:
+        - reader
+        - curator
+        - admin
+tags:
+  - name: reader
+    description: An operation that can be executed by a reader.
+  - name: curator
+    description: An operation that can be executed by a curator.
+  - name: admin
+    description: An operation that can be executed by an admin.
+components:
+  schemas:
+    Case:
+      type: object
+      properties:
+        id:
+          type: string
+          description: The case ID.
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -2,36 +2,38 @@ openapi: 3.0.3
 info:
   title: EPID data service
   description: Server for CRUD operations on EPID line-list data
-  termsOfService: 'https://www.healthmap.org/about/termsofuse/'
+  termsOfService: "https://www.healthmap.org/about/termsofuse/"
   contact:
     email: info@healthmap.org
   license:
     name: MIT
   version: 1.0.0
 paths:
-  /api/cases/{id}:
+  "/cases/{id}":
     summary: Line-list cases
     get:
       summary: Gets a specific case
-      parameters: 
-        - in: path
-          name: id
+      parameters:
+        - name: id
+          in: path
           required: true
           schema:
             type: string
           description: The case ID
       responses:
-        '200':
+        "200":
           description: Case found.
           content:
             application/json:
               schema:
-                $ref: '#/components/schemas/Case'
-        '400':
-          description: Bad request.
-        '401':
+                $ref: "#/components/schemas/Case"
+        "400":
+          description: Malformed request.
+        "422":
+          description: Unprocessable entity.
+        "401":
           description: Unauthorized.
-        '404':
+        "404":
           description: Case not found.
       tags:
         - reader
@@ -52,3 +54,6 @@ components:
         id:
           type: string
           description: The case ID.
+servers:
+  - url: "http://localhost:3000"
+    description: Local server.
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -51,10 +51,6 @@ components:
   schemas:
     Case:
       type: object
-      properties:
-        id:
-          type: string
-          description: The case ID.
 servers:
   - url: "http://localhost:3000"
     description: Local server.
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -9,7 +9,7 @@ info:
     name: MIT
   version: 1.0.0
 paths:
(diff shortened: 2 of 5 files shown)


data-serving/data-service/api/dataserver.api.yaml
```diff
@@ -0,0 +1,54 @@
+openapi: 3.0.3
```
axmb, line 1: Since this is already in the data-service dir, we should use the recommended root spec file name of `openapi.yaml`. Or is your plan that other APIs might produce such a specification, and we'd organize a single epid root spec?

axmb, line 1: Glad you chose YAML over JSON. All of those defenseless curly braces...

axmb, line 1: There's this neat extension, in case you hadn't seen:

https://marketplace.visualstudio.com/items?itemName=42Crunch.vscode-openapi

data-serving/data-service/api/dataserver.api.yaml
```diff
@@ -0,0 +1,54 @@
...
+info:
+  title: EPID data service
+  description: Server for CRUD operations on EPID line-list data
+  termsOfService: 'https://www.healthmap.org/about/termsofuse/'
+  contact:
+    email: info@healthmap.org
+  license:
+    name: MIT
+  version: 1.0.0
```
axmb, line 10: Worth adding a `servers` object pointing to the port we've designated on localhost? Seems reasonably canonical to do so, and slap a `description: Local server` on it.

data-serving/data-service/api/dataserver.api.yaml
```diff
@@ -0,0 +1,54 @@
...
+paths:
+  /api/cases/{id}:
+    summary: Line-list cases
+    get:
+      summary: Gets a specific case
+      parameters: 
+        - in: path
+          name: id
```
axmb, line 18: hyper-nit: seems more fluent/conventional to put `name` before `in`

data-serving/data-service/api/dataserver.api.yaml
```diff
@@ -0,0 +1,54 @@
...
+          required: true
+          schema:
+            type: string
```
axmb, line 21: Might be good to add `format: uuid`, here.

data-serving/data-service/api/dataserver.api.yaml
```diff
@@ -0,0 +1,54 @@
...
+          description: The case ID
+      responses:
+        '200':
+          description: Case found.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Case'
+        '400':
```
axmb, line 30: This API (perhaps mistakenly) returns 422 for all bad requests.

data-serving/data-service/src/index.ts
```diff
@@ -27,6 +29,10 @@ apiRouter.put('/cases/:id([a-z0-9]{24})', caseController.update);
 apiRouter.delete('/cases/:id([a-z0-9]{24})', caseController.del);
 app.use('/api', apiRouter);
 
+// API documentation.
+const swaggerDocument = YAML.load('./api/dataserver.api.yaml');
+app.use('/api-docs', swaggerUi.serve, swaggerUi.setup(swaggerDocument));
```
axmb, line 34: So cool. Do you think documentation is the primary outcome here? Do we want to accomplish anything else based on the spec?

data-serving/data-service/api/dataserver.api.yaml
khmoran: nah probably just this one! renamed

khmoran: {
  {
    thanks!
  }
}

khmoran: nice! i was using their web editor (https://editor.swagger.io/) but in the IDE is even better!

data-serving/data-service/api/dataserver.api.yaml
khmoran: good call, thanks!

data-serving/data-service/api/dataserver.api.yaml
khmoran: done

data-serving/data-service/api/dataserver.api.yaml
khmoran: yeah i was waffling on that but what the hell, let's see what happens

data-serving/data-service/api/dataserver.api.yaml
khmoran: hmm... I'll add 422 (and keep 400 for malformed requests)

data-serving/data-service/src/index.ts
khmoran: I think documentation is the low-hanging fruit, but I think using the spec to enforce the API would be even nicer! I think we have a few options, would love feedback/thoughts:

- At a minimum I'd like to validate requests & responses, but still evaluating the different libraries for this (there are a couple and I want to play with an example to see which is better, like https://github.com/cdimascio/express-openapi-validator vs. https://github.com/urma/express-openapi vs https://github.com/isa-group/oas-tools).

- We could potentially use the tags in code as part of RBAC

- Trying to figure out if any of the validation libs generate nice TS bindings to help us construct requests or responses -- presumably yes, since they can validate but want to dig in and see

khmoran:
[Design & implement slice of dataserver API](https://trello.com/c/pZnsDRDg/135-design-implement-slice-of-dataserver-api)

khmoran:
> Really nice!

Thanks! LMK what you think about going with this. I sent another draft PR for validation -- happy to go about adding the spec for each endpoint, leaving the Cases obj undefined until we can (and want to) decide as a group what it should look like. WDYT?

axmb:
> > Really nice!
> 
> Thanks! LMK what you think about going with this. I sent another draft PR for validation -- happy to go about adding the spec for each endpoint, leaving the Case obj undefined until we can (and want to) decide as a group what it should look like. WDYT?

SGTM

data-serving/data-service/src/index.ts
axmb: > At a minimum I'd like to validate requests & responses

Validation SGTM. Will have to make sure the overall behavior is coherent with both request validation and data model validation in the mix.

> We could potentially use the tags in code as part of RBAC

Yeah, I'll uninformedly bet that with enough elbow grease we could manually operate on the tags in middleware. Will be interesting to see if something like this works well enough for us to use.

> Trying to figure out if any of the validation libs generate nice TS bindings to help us construct requests or responses

SG

data-serving/data-service/api/openapi.yaml
```diff
@@ -0,0 +1,56 @@
...
+      parameters:
+        - name: id
+          in: path
+          required: true
+          schema:
+            type: string
+            format: uuid
+          description: The case ID
+      responses:
+        "200":
```
axmb, line 25: Another nit, but we're also returning 201 for createCase, IIRC.

khmoran: I think this list of response codes is just for the GET request (slight drawback of YAML is hard to tell indentations) -- when i add the POST, i'll be sure to use 201

axmb: Right, sorry. Note to self: don't comment just based on PR code snippets. ;-)

PR #180, opened 2020-05-21
Use Express OpenAPI Validator to validate /api/cases/:id GET requests

khmoran:
Not sure if this is the lib we want to go with but a proof of concept -- LMK what you think!

Video of error when incorrect id format supplied: https://screencast.googleplex.com/cast/NTkyMzk5OTI0NDYxNTY4MHwyMzhhMzU2NS0xOQ

### DIFF
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -51,10 +51,6 @@ components:
   schemas:
     Case:
       type: object
-      properties:
-        id:
-          type: string
-          description: The case ID.
 servers:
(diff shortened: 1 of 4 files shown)


khmoran:
[Design & implement slice of dataserver API](https://trello.com/c/pZnsDRDg/135-design-implement-slice-of-dataserver-api)

khmoran:
> Ha, nice! Can we control the message/code of the response?

yes, in theory we can add custom error handling middleware to handle 500's (the validation error code) -- i'm actually having some trouble getting the error handler to be called so i'll do this in a separate PR

PR #182, opened 2020-05-21
Add a basic AWS CloudWatch Events client

axmb:
Just adding the client in this change. Initially exposing `PutRule` (which'll be used to create/update scheduled events based on source config).

Placing this in a client helps encapsulate the AWS logic, but mostly did this in order to be able to mock the AWS connection in tests. Can't do that if the AWS CW service object is instantiated directly in our app. I'll construct our client object and use it in our controller in a subsequent PR.

Note that I'll add credentials as environmental variables in a subsequent PR as well. For lack of these variables, the AWS SDK will try to snag your `~/.aws/credentials` for use locally ([and for lack of that, probably won't work](https://docs.aws.amazon.com/sdk-for-javascript/v2/developer-guide/setting-credentials-node.html)).


verification/curator-service/api/src/clients/aws-events-client.ts
```diff
@@ -0,0 +1,56 @@
...
+                name: ${ruleName}
+                schedule: ${scheduleExpression}`,
+            );
+            throw err;
+        }
+    };
+
+    private assertString(input: string | undefined): asserts input is string {
+        if (typeof input === 'string') return;
+        else throw new Error('AWS PutRule response missing RuleArn.');
```
attwad, line 54: can any further help be added? What are we supposed to do in that case? when does it happen?

axmb: It shouldn't happen ever for a successful response, but we need this assertion helper because the field is nullable (/shrug). I debated making this an AssertionError, because that'd be canonical in Java, but that didn't seem conventional in TS.

verification/curator-service/api/src/clients/aws-events-client.ts
```diff
@@ -0,0 +1,56 @@
+import AWS from 'aws-sdk';
```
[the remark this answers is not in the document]
axmb: There's already a "semantic"-driven directory structure in this app, IMO (e.g. controllers, model, util).

But, I don't really mind either way -- if putting this in `src/aws/` is clearer IYO I'm happy to change it. Shall I mv?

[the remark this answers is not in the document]
attwad: no big deal, you're right we already have a semantic driven directory architecture now, fine as is then.

verification/curator-service/api/src/clients/aws-events-client.ts
attwad: You mean adding an `assert()` call here isn't conventional in TS? how so?

attwad: ah nm https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions explains that well, lgtm

PR #183, opened 2020-05-25
Production tuning of docker images

attwad:
Prepare for push of prod images to docker hub.

- Follow security best practices, do not run as root
- Specify production node env
- Use multi staged builds to *drastically*  trim down the images size:
  - dev_curatorui: 1.72GB
  - curator: 322MB
  - dev_data: 1.28GB
  - data: 415MB

Went back to separate docker files for the curator and UI, the builds were really too slow on my machine when I wanted to run the dev stack, separate images make it so that we don't produce a production build of the FE when we don't need it, at the cost of a small dockerfile content duplication.

Also minor nit fixes along the way:
- no need to bind a volume mount on CI (-V)
- fix lint warnings in integration tests

Finaly I am not adding a github workflow to check that the prod images build similarly to how we were doing this before we had integration tests, this is because I am setting up automated builds on docker-hub for those images, I am not sure if we can do it at the PR stage though so I might do it once I could confirm.


attwad:
[Production docker images](https://trello.com/c/prFA7FHE/159-production-docker-images)

data-serving/data-service/Dockerfile
```diff
@@ -23,5 +26,25 @@ RUN npm run-script build
...
 EXPOSE 3000
 
-# Start service.
-CMD [ "npm", "start" ]
\ No newline at end of file
+# Specify a command for up to this step.
+CMD [ "npm", "start" ]
+
+# Multi-staged build, we don't need a full node image to run the app.
+FROM node:alpine as prod
```
axmb, line 33: What happens when we don't specify a version here? Does it just use the max version number?

https://hub.docker.com/_/node/

data-serving/data-service/Dockerfile
```diff
@@ -23,5 +26,25 @@ RUN npm run-script build
...
+# No need to run as root.
+USER node
+
+# Run with node env set to production.
+ENV NODE_ENV production
+
+WORKDIR /usr/src/app/
+
+# Copy compiled app from previous stage.
+COPY --from=dev /usr/src/app/data-serving/data-service/node_modules ./node_modules
```
axmb, line 44: What's up with copying the code to a different directory? Is it to avoid tmp file collisions?

verification/curator-service/Dockerfile
```diff
@@ -31,9 +32,27 @@ COPY verification/curator-service/ui/. .
 ENV REACT_APP_LOGIN_URL /auth/google
 RUN npm run-script build
```
axmb, line 33: Would be good to make these all `run` if we're in the business of changing them now.

verification/curator-service/Dockerfile
```diff
@@ -31,9 +32,27 @@ COPY verification/curator-service/ui/. .
...
 
-# Go back to curator service
-WORKDIR /usr/src/app/verification/curator-service/api
+
+# Multi-staged build, use a trimmed down version of node for prod.
+FROM node:alpine as prod
+
+# No need to run as root.
+USER node
```
axmb, line 40: Any prerequisites to do this? On my machine, `groups node` is empty, so this will run as ~root anyway. Is the `node` user correctly configured in our to-be prod VMs?

data-serving/data-service/Dockerfile
attwad: Ah I had forgotten the version indeed, done now, used 12 to match the build version.

data-serving/data-service/Dockerfile
attwad: there's no need to run the app from this long-named-folder-that-was-like-this-due-to-build-context, so in prod the app just runs from /usr/src/app/

verification/curator-service/Dockerfile
attwad: indeed, done

verification/curator-service/Dockerfile
attwad: Not sure why it's not present on your machine, it is when I run the docker image:

```
docker run -it --rm --init node:12-alpine sh
/ # whoami
root
/ # su node
/ $ whoami
node
/ $ id -u node
1000
/ $ id -u root
0
```

axmb: Yeah, this was my bad. WFM!

PR #185, opened 2020-05-26
Add roles to user model

allysonjp715:


allysonjp715:
[Add roles to user model for authorization](https://trello.com/c/2jhHGQy9/142-add-roles-to-user-model-for-authorization)

PR #188, opened 2020-05-26
add roles checking middleware

attwad:
Didn't plug it in any index.ts routing yet so it's a no-op for now.

### DIFF
--- a/verification/curator-service/api/src/controllers/auth.ts
+++ b/verification/curator-service/api/src/controllers/auth.ts
@@ -21,12 +21,12 @@ export const mustBeAuthenticated = (
 };
 
 // mustHaveAnyRole is an express middleware that checks that the currently logged-in user has any of the given roles.
-export const mustHaveAnyRole = (roles: Set<string>) => {
+export const mustHaveAnyRole = (requiredRoles: Set<string>) => {
     return (req: Request, res: Response, next: NextFunction): void => {
         if (
             req.isAuthenticated() &&
             (req.user as UserDocument).roles?.filter((r: string) =>
-                roles.has(r),
+                requiredRoles.has(r),
             ).length > 0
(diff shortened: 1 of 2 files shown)


attwad:
[Add authorization logic (i.e. require curator role to CRUD data)](https://trello.com/c/WhSgmMMw/150-add-authorization-logic-ie-require-curator-role-to-crud-data)

verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -20,6 +20,22 @@ export const mustBeAuthenticated = (
     res.sendStatus(403);
 };
 
+// mustHaveAnyRole is an express middleware that checks that the currently logged-in user has any of the given roles.
+export const mustHaveAnyRole = (roles: Set<string>) => {
+    return (req: Request, res: Response, next: NextFunction): void => {
+        if (
+            req.isAuthenticated() &&
+            (req.user as UserDocument).roles?.filter((r: string) =>
+                roles.has(r),
```
allysonjp715, line 29: I got confused reading this since roles here has the same name as the user's roles. Could you name it something like requiredRoles to differentiate?

verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -20,6 +20,22 @@ export const mustBeAuthenticated = (
...
+            ).length > 0
+        ) {
+            next();
+        } else {
+            res.sendStatus(403);
```
allysonjp715, line 34: Also send a message saying something about roles are not sufficient?

verification/curator-service/api/src/controllers/auth.ts
attwad: Sure, done

PR #190, opened 2020-05-26
Create AWS event rules from the Sources.create API

axmb:
Few notes:

- Small data model change needed here (it's the third commit; feel free to view in isolation).
- For the (large) diff in SourcesController -- this is largely an indentation change as a result of creating a class for the controller. The only logical change is to the `create` function.
- On testing: `jest.mock` with a factory function is the simplest way of successfully mocking our client object in the controller (because Typescript). If we want to verify on the result of that method (e.g. to make sure the created ARN is what we tell the mock to use), we'll need to create a [manual mock](https://jestjs.io/docs/en/manual-mocks#mocking-user-modules) for the module. I'll do this later if we think it's valuable.
- AWS credentials are bum values by default; you'll have to provide your credentials in a `[...]/api/.env` file to run this locally. In terms of our deployment -- we'll need to configure these as secrets with the mechanism provided by the system we choose. So let's add this after that's finalized.


verification/curator-service/api/test/sources.test.ts
```diff
@@ -3,6 +3,15 @@ import app from '../src/index';
 import mongoose from 'mongoose';
 import request from 'supertest';
 
+jest.mock('../src/clients/aws-events-client', () => {
+    const mockPutRule = jest
+        .fn()
+        .mockResolvedValue('arn:aws:events:fake:event:rule/name');
+    return jest.fn().mockImplementation(() => {
```
khmoran, line 10: you know me, always one-lineifying... does this work, or do the braces mess it up?

return jest.fn().mockImplementation(() => { putRule: mockPutRule});

axmb: Unfortunately we can't. This needs to be a HOF, if you believe [Jest docs](https://jestjs.io/docs/en/es6-class-mocks#calling-jestmockdocsenjest-objectjestmockmodulename-factory-options-with-the-module-factory-parameter).

PR #193, opened 2020-05-27
Cleanup of dataserver API spec in preparation for adding more endpoints

khmoran:
- Removes tags
- Makes error responses reusable (OK response content may differ between endpoints)
- Moves GET params to top-level since {id} is in the URL (and so it would apply to all requests of that path)
- Adds MIT license URL
- Makes /api part of the base path for all requests


khmoran:

PR #195, opened 2020-05-27
Add a validated schedule expression to the Source schedule model

axmb:
Small meta-note: we'd debated not including things like schedule in our source schema, given that it'd be potentially duplicative of (and ergo inconsistent with) the data in AWS. For now I think this is the right thing to do, as it simplifies the work that the curator service does. Later, we should harden this synchronization.


verification/curator-service/api/src/model/schedule.ts
```diff
@@ -6,8 +6,16 @@ export const scheduleSchema = new mongoose.Schema({
         required: 'Enter a CloudWatch schedule rule AWS Lambda ARN',
         match: [/^arn\:aws\:events\:.+\:.+\:rule\/.+/],
     },
+    awsScheduleExpression: {
+        type: String,
+        required: 'Enter a CloudWatch rule schedule expression',
+        match: [
+            /^((rate\(((1 (minute|hour|day))|([1-9]+ (minutes|hours|days)))\))|(cron\((.+\s){5}.+\)))/,
```
khmoran, line 17: what a delicious regex. the comment is v. helpful :D

the match property is cool!

verification/curator-service/api/test/model/data/automation.full.json
```diff
@@ -3,6 +3,7 @@
         "awsLambdaArn": "arn:aws:lambda:region:id:function:name"
     },
     "schedule": {
-        "awsRuleArn": "arn:aws:events:region:id:rule/name"
+        "awsRuleArn": "arn:aws:events:region:id:rule/name",
+        "awsScheduleExpression": "cron(0 10 * * ? *)"
```
khmoran, line 7: neither here nor there for this particular PR but wondering if we can actually use $ref so we don't have to re-specify sub-documents in these examples... not sure if they'll be dereferenced on import though... maybe something I'll try

i just wanna stay nice and DRY, you know?

verification/curator-service/api/test/model/schedule.test.ts
```diff
@@ -25,6 +25,24 @@ describe('validate', () => {
...
+    it('a schedule without an AWS schedule expression is invalid', async () => {
+        const missingExpression = { ...fullModel };
+        delete missingExpression.awsScheduleExpression;
+
+        return new Schedule(missingExpression).validate((e) => {
+            expect(e.name).toBe(Error.ValidationError.name);
+        });
+    });
+
+    it('a schedule with a misformated AWS schedule expression is invalid', async () => {
```
khmoran, line 37: super nit: s/misformated/misformatted

verification/curator-service/api/test/model/schedule.test.ts
```diff
@@ -25,6 +25,24 @@ describe('validate', () => {
...
+        const badSchedule = { ...fullModel };
```
khmoran, line 38: up to you, can inline as

`return new Schedule({ ...fullModel, awsScheduleExpression: 'rate(1 hour') }).validate ...`

but totally get it if you find that less readable

verification/curator-service/api/src/model/schedule.ts
axmb: Yeah -- I considered getting more precise with the `cron([...])` spec, but decided this was the right balance for now. Not that I don't love a good write-only regex.

verification/curator-service/api/test/model/data/automation.full.json
axmb: That'd be really neat -- let me know if you want a hand with that.

verification/curator-service/api/test/model/schedule.test.ts
axmb: I fear I'll never not misspell this.

verification/curator-service/api/test/model/schedule.test.ts
axmb: Ah, that's pretty nice! I'll shelf this for after the schema changes settle. This existing format is probably used in most tests.

axmb:

PR #196, opened 2020-05-27
Add endpoint definition for create case

khmoran:
(Note: first commit is minor cleanup)

Following the inheritance pattern for the entities, since I expect that the case that we POST will be a strict subset of what we GET/LIST, although both are just {} right now:

- Docs: https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/
- Example: https://github.com/OAI/OpenAPI-Specification/blob/master/examples/v3.0/petstore-expanded.yaml

Screencast:
FmkmFGjnrZJ

### DIFF
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -13,6 +13,7 @@ paths:
   /cases:
     get:
       summary: Lists cases
+      operationId: listCases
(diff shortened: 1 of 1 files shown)


khmoran:

khmoran:
> Well gif'd.

you honor me

PR #197, opened 2020-05-27
Add delete case endpoint definition to API

khmoran:
I had to remove the uuid format on the id field... according to OpenAPI, mongoDB's ids are not valid UUIDs: https://github.com/swagger-api/swagger-ui/issues/5281. I added a regex that matches MongoDB's ids instead

I also changed the server to return a 204, and not return the case object

Screencast:
delete


khmoran:

PR #198, opened 2020-05-27
Make source.automation.schedule.awsRuleArn optional

axmb:
See comment in the model file. Open to suggestions, if there's a better way to facilitate this pattern with Mongoose et al.


axmb:

PR #203, opened 2020-05-28
Add updateCase endpoint to API definition

khmoran:
Screencast:
update

### DIFF
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -88,11 +88,30 @@ paths:
       operationId: getCase
       responses:
         "200":
-          description: OK
-          content:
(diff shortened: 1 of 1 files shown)


khmoran:

data-serving/data-service/api/openapi.yaml
```diff
@@ -132,6 +151,12 @@ components:
       description: The subset of "#/components/schemas/Case" required to create the entity
       type: object
   responses:
+    "200Case":
```
axmb, line 154: Worth adding a 201Case, just so these are all in the same place?

khmoran: SG, made a 200CaseArray and moved it there as well

PR #204, opened 2020-05-28
Add automation details to the ingestion sources UI

axmb:
Couple notes:

- Considered doing something with "better" UX around schedule expression, but ended up reversing that. For lack of a bit more thought on the UX, we should probably punt on this for now.
- Not including AWS-esque validation on the schedule. There's a medium-sized regex in the Curator API data layer we can share here, but wasn't totally happy with just copy-pasta'ing that. Can revisit this later on.
- Excuse the inclusion of Docker changes, which I've added to locally E2E test.

### DIFF
--- a/verification/curator-service/ui/src/components/SourceTable.tsx
+++ b/verification/curator-service/ui/src/components/SourceTable.tsx
@@ -61,8 +61,7 @@ interface TableRow {
     name: string;
     // origin
     url: string;
-    // automation
-    // schedule
+    // automation.schedule
     awsRuleArn?: string;
     awsScheduleExpression?: string;
 }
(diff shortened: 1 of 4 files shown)


verification/curator-service/ui/src/components/SourceTable.tsx
```diff
@@ -59,6 +61,10 @@ interface TableRow {
     name: string;
     // origin
     url: string;
+    // automation
+    // schedule
```
allysonjp715, line 65: nit: should //schedule go below awsRuleArn?

axmb: Both of these fields are in automation.schedule. Sorry if this is a strange way to denote that -- I'll change the comment format.

axmb:

axmb:
[Integrate read/update/delete methods of the ingestion source API with AWS](https://trello.com/c/iKw7u0U1/183-integrate-read-update-delete-methods-of-the-ingestion-source-api-with-aws)

PR #205, opened 2020-05-29
Setup prod/dev instances on EKS and document release process

attwad:
- Renamed the cluster `healthmapidha` for consistency in what Gaurav created for us (docker hub for example) and that makes this cluster less epid-specific if Spencer wants to reuse it for other apps.

- current version is 0.0.3 because it took me 3 attemps to get the docker hub regexp right /sigh

### DIFF
--- a/aws/data.yaml
+++ b/aws/data.yaml
@@ -20,7 +20,7 @@ spec:
     spec:
       containers:
         - name: data
-          image: docker.io/healthmapidha/dataservice:latest
+          image: docker.io/healthmapidha/dataservice:0.0.1
           ports:
             - containerPort: 3000
           env:
@@ -58,7 +58,7 @@ spec:
     spec:
       containers:
         - name: data
-          image: docker.io/healthmapidha/dataservice:latest
+          image: docker.io/healthmapidha/dataservice:0.0.1
           ports:
             - containerPort: 3000
           env:
--- a/aws/data.yaml
+++ b/aws/data.yaml
@@ -20,7 +20,7 @@ spec:
     spec:
       containers:
         - name: data
-          image: docker.io/healthmapidha/dataservice:0.0.1
+          image: docker.io/healthmapidha/dataservice:0.0.3
           ports:
             - containerPort: 3000
           env:
@@ -58,7 +58,7 @@ spec:
     spec:
       containers:
         - name: data
-          image: docker.io/healthmapidha/dataservice:0.0.1
+          image: docker.io/healthmapidha/dataservice:0.0.3
           ports:
             - containerPort: 3000
           env:
--- a/aws/data.yaml
+++ b/aws/data.yaml
@@ -2,18 +2,20 @@
 apiVersion: apps/v1
 kind: Deployment
 metadata:
-  name: data
+  name: data-dev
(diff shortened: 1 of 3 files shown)


aws/data.yaml
```diff
@@ -2,30 +2,32 @@
...
   template:
     metadata:
       labels:
+        environment: dev
         app: data
     spec:
       containers:
         - name: data
-          image: docker.io/healthmapidha/dataservice:latest
+          image: docker.io/healthmapidha/dataservice:0.0.3
```
axmb, line 23: Shouldn't this be some sort of tag, rather than a concrete version, so that the image can be updated without committing changes to this config file? I haven't researched this -- just asking based on notions.

aws/data.yaml
```diff
@@ -2,30 +2,32 @@
...
           ports:
             - containerPort: 3000
           env:
             - name: DB_CONNECTION_STRING
               valueFrom:
                 secretKeyRef:
-                  name: data-878ckgc42b
+                  name: data-dev-ddhhd4mmhf
```
axmb, line 30: Silly question perhaps, but why is this binding name machine generated? Most examples I see have a human readable binding "category" for name, and the key is the specific secret being used.

aws/data.yaml
```diff
@@ -35,13 +37,65 @@ spec:
...
+                  name: data-prod-84t6gt4hhc
+                  key: db_connection_string
+          resources:
+            requests:
+              memory: "256Mi"
+              cpu: "100m"
+            limits:
+              memory: "1Gi"
+              cpu: "500m"
+---
```
axmb, line 77: Any way to share common config between dev/prod? Or is it more trouble than it's worth?

aws/data.yaml
attwad: constant tags are not considered best practices because new images will be pulled only when a container is restarted, not when a new tag is applied to the image (kubernetes master does not watch docker hub). I sent you a chat with a link to why this is also considered bad practice internally.

aws/data.yaml
attwad: I'm using kubernetes kustomize to generate the secrets, you can check out the official docs here
https://kubernetes.io/docs/concepts/configuration/secret/
The machine generated suffix is here for that reason:
_Note: When a Secret is generated, the Secret name is created by hashing the Secret data and appending this value to the name. This ensures that a new Secret is generated each time the data is modified._

aws/data.yaml
attwad: Hmm I think one can do inheritance in yaml I think but honestly I don't think we want to go this way unless we feel we get huge benefits from it...

aws/data.yaml
axmb: Makes sense for prod, but I'm asking specifically re: the change from latest to 0.0.3 for dev. Do we want to update/push this file prior to performing a restart every time we want to update dev? Perhaps there's a way to automate this?

attwad: For dev we can keep using `:latest` but I am not sure how to trigger an application of that deployment upon the build of a new image, I'll look into docker hub hooks.

PR #208, opened 2020-05-29
Propagate user role changes to other pages without refresh

allysonjp715:


verification/curator-service/ui/src/components/Users.test.tsx
```diff
@@ -110,3 +112,117 @@ test('updates roles on selection', async () => {
...
+        status: 200,
+        statusText: 'OK',
+        config: {},
+        headers: {},
+    };
+    mockedAxios.put.mockResolvedValueOnce(axiosPutResponse);
+    expect(functionCalledCounter).toBe(0);
+
+    // Select new role
+    fireEvent.mouseDown(getByRole('button'));
```
attwad, line 162: I'm thinking of this now but will all these tests break once we add another button to this page? Can we be a bit more explicit that this is a button to change the roles?

allysonjp715: Took some ugliness to get a test id in the right place but finally achieved! Unfortunately I couldn't find a way of chaining lookups e.g. execute getByText() on only children of a different getByTestId() call :/

PR #210, opened 2020-06-01
Converter: Handle lists of locations in the travelHistory field

khmoran:


khmoran:

data-serving/scripts/convert-data/converters.py
```diff
@@ -129,7 +129,7 @@ def convert_event(id: str, name: str, dates: Any) -> Dict[str, Any]:
 
     Returns:
       None: When the input value is empty.
-      Dict[str, str]: When the values are present and succesfully parsed. The
+      Dict[str, str]: When the values are present and successfully parsed. The
```
axmb, line 132: Could this have been the plug-in?

data-serving/scripts/convert-data/converters.py
```diff
@@ -462,28 +462,26 @@ def convert_travel_history(id: str, dates: str, location: str) -> Dict[
...
+
+    if not location_list and not date_range:
+        return None
+    if not location_list:
+        return [{'dateRange': date_range}]
+    if not date_range:
+        return [{'location': l} for l in location_list if l]
 
-    return [travel_history] if travel_history else None
+    return [{'dateRange': date_range, 'location': l} for l in location_list if l]
```
axmb, line 484: Is the idea behind including date range at this level that it'd be useful to have for each location, but for the existing data, travel history is guaranteed to have either 0 or 1 date range? I agree with that reasoning -- but making sure I grok'd correctly.

data-serving/scripts/convert-data/parsers.py
```diff
@@ -239,6 +236,35 @@ def parse_location(value: Any) -> Dict[str, Any]:
...
+    separator = None
+    if pd.isna(value):
+        return None
+    if ';' in value:
+        separator = ';'
+    if ':' in value:
+        separator = ':'
+    if ' and ' in value:
+        separator = ' and '
+    if value.count(',') > 1:
```
axmb, line 261: Making sure I understand this one: there are singular locations of the form "Miami, FL" -- but we're positive we've never inputted multiple comma-containing locations as a comma-delimited list?

data-serving/scripts/convert-data/parsers.py
```diff
@@ -239,6 +236,35 @@ def parse_location(value: Any) -> Dict[str, Any]:
...
+        A list of dictionaries containing the location data.
+    '''
+    separator = None
+    if pd.isna(value):
+        return None
+    if ';' in value:
+        separator = ';'
+    if ':' in value:
+        separator = ':'
+    if ' and ' in value:
```
axmb, line 259: I'm not sure if there are many instances that'd trip this up, but there's [at least one](https://en.wikipedia.org/wiki/Rough_and_Ready,_California). :-P

I take it you've briefly sanity checked this separator list?

khmoran: eek... OK, let me remove this separator.

what i'm working on right now will throw an error if we can't match the location string, so if we end up having something like "NY and FL", this will get logged and we should fix it in the source to be "NY; FL"

similarly, I'm actually going to remove the check above for commas -- I'd really like us to standardize on a non-comma delimiter so commas can be used e.g. for "NY, New York, USA"

data-serving/scripts/convert-data/parsers.py
khmoran: see below -- TL;DR: no, definitely not sure, and following PR will help track these down and log them

data-serving/scripts/convert-data/converters.py
khmoran: exactly! i'll leave a comment

data-serving/scripts/convert-data/converters.py
khmoran: you bet it was!!

PR #211, opened 2020-06-01
Add deletion to the AWS CWE client API; integrate this with the Sources API

axmb:
Doesn't include delete-as-update -- that'll be subsequent.

delete-source

### DIFF
--- a/verification/curator-service/api/test/clients/aws-events-clients.test.ts
+++ b/verification/curator-service/api/test/clients/aws-events-clients.test.ts
@@ -57,3 +57,30 @@ describe('putRule', () => {
             .catch((e) => expect(e.message).toContain('missing RuleArn'));
     });
 });
+
+describe('deleteRule', () => {
+    it('deletes the AWS CloudWatch Event rule via the SDK', async () => {
+        const deleteRuleSpy = jest.fn().mockResolvedValueOnce({});
+        AWSMock.mock('CloudWatchEvents', 'deleteRule', deleteRuleSpy);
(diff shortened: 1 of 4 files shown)


verification/curator-service/api/test/clients/aws-events-clients.test.ts
```diff
@@ -57,3 +57,30 @@ describe('putRule', () => {
...
+            'CloudWatchEvents',
+            'deleteRule',
+            (params: any, callback: Function) => {
+                callback(expectedError, null);
+            },
+        );
+        const client = new AwsEventsClient('us-east-1');
+
+        expect.assertions(1);
+        return client
```
khmoran, line 82: guessing you investigated this -- but we can't do an expect.toThrow here instead of expecting an assertion from the catch statement?

axmb: We can -- either's perfectly valid javascript(tm) according to Jest (see [Promise](https://jestjs.io/docs/en/asynchronous#promises) vs. [`rejects.toThrow()`](https://jestjs.io/docs/en/asynchronous#asyncawait)). I like that the Promise approach is more declarative/flexible, but we can give toThrow a swing for the sake of slaying `catch` statements.

axmb:

axmb:
[Integrate read/update/delete methods of the ingestion source API with AWS](https://trello.com/c/iKw7u0U1/183-integrate-read-update-delete-methods-of-the-ingestion-source-api-with-aws)

PR #212, opened 2020-06-01
Converter: Geocode the travelHistory with the same system as the location field

khmoran:
Sorry for the big diff :\ This PR is doing a few things:

- Reusing the same geocoding data for the travelHistory.location as is used in the original CSV for geocoding the location field. This gets us (1) better coverage than pycountries (countries & counties only), (2) more consistency, and (3) full geospatial data (lat/longs).

- Changing the script to take in a path to the location of a locally-checked out nCoV19 repo, for (1) access to the CSV, and (2) for access to the geocoding script & data.

- Adding some very common abbreviations used in the travelHistory.location source, since fixing every one of these would be a lot of work.

- Remove hacks and workarounds, like removing 'Travelled to' prefixes. We're logging these as errors so we can fix them at the source.

Note that the existing geocoding script from the other doesn't do exactly what we need here, hence `geocode_util.py`.

This reduces errors quite a bit, down to 2,441 total errors, the vast majority of which seems like problems we need to solve in the CSV, rather than fixable geocoding misses.

Some example outputs:
```
    "travelHistory": [
      {
        "dateRange": {
          "start": {
            "$date": "2020-03-09T00:00:00Z"
          },
          "end": {
            "$date": "2020-03-09T00:00:00Z"
          }
        },
        "location": {
          "country": "Spain",
          "geometry": {
            "latitude": 40.3396,
            "longitude": -3.6226599999999998
          }
        }
      }
    ],
```

```
    "travelHistory": [
      {
        "location": {
          "country": "United Arab Emirates",
          "geometry": {
            "latitude": 25.204849,
            "longitude": 55.270782
          },
          "administrativeAreaLevel1": "Dubai",
          "administrativeAreaLevel2": "Dubai"
        }
      }
    ],
```

### DIFF
--- a/data-serving/scripts/convert-data/convert_data.py
+++ b/data-serving/scripts/convert-data/convert_data.py
@@ -40,24 +40,7 @@ def main():
 
     args = parser.parse_args()
 
-    # Load the geocoder.
-    geocoder_path = os.path.join(args.ncov2019_path, GEOCODER_REPO_PATH)
-    geocodes_path = os.path.join(geocoder_path, GEOCODER_DB_FILENAME)
-    f, path, desc = imp.find_module(GEOCODER_MODULE, [geocoder_path])
-    geocoder_module = imp.load_module(GEOCODER_MODULE, f, path, desc)
-    geocoder = geocoder_module.CSVGeocoder(geocodes_path, lambda x: None)
-
-    # Load the case data.
-    gzip_path = os.path.join(
-        args.ncov2019_path, DATA_REPO_PATH,
-        DATA_GZIP_FILENAME)
-    csv_path = os.path.join(
-        args.ncov2019_path, DATA_REPO_PATH, DATA_CSV_FILENAME)
-
-    print('Unzipping', gzip_path)
-    latest_data_gzip = tarfile.open(gzip_path)
-    latest_data_gzip.extract(DATA_CSV_FILENAME)
-    latest_data_gzip.close()
+    csv_path = extract_csv(args.ncov2019_path)
 
     print('Reading data from', csv_path)
     original_cases = read_csv(csv_path)
@@ -70,12 +53,34 @@ def main():
             f'{original_rows} to {original_cases.shape[0]} rows')
 
     print('Converting data to new schema')
+    geocoder = load_geocoder(args.ncov2019_path)
     converted_cases = convert(original_cases, geocoder)
 
     print('Writing results to', args.outfile.name)
     write_json(converted_cases, args.outfile)
 
     print('Great success! 🎉')
+
+
+def load_geocoder(repo_path: str) -> Any:
+    geocoder_path = os.path.join(repo_path, GEOCODER_REPO_PATH)
+    geocodes_path = os.path.join(geocoder_path, GEOCODER_DB_FILENAME)
+    f, path, desc = imp.find_module(GEOCODER_MODULE, [geocoder_path])
+    geocoder_module = imp.load_module(GEOCODER_MODULE, f, path, desc)
+    return geocoder_module.CSVGeocoder(geocodes_path, lambda x: None)
+
+
+def extract_csv(repo_path: str) -> str:
+    gzip_path = os.path.join(
+        repo_path, DATA_REPO_PATH,
+        DATA_GZIP_FILENAME)
+    print('Unzipping', gzip_path)
+    latest_data_gzip = tarfile.open(gzip_path)
+    latest_data_gzip.extract(DATA_CSV_FILENAME)
+    latest_data_gzip.close()
+
+    return os.path.join(
+        repo_path, DATA_REPO_PATH, DATA_CSV_FILENAME)
 
 
 def read_csv(infile: str) -> DataFrame:
--- a/data-serving/scripts/convert-data/convert_data.py
+++ b/data-serving/scripts/convert-data/convert_data.py
@@ -79,8 +79,7 @@ def extract_csv(repo_path: str) -> str:
     latest_data_gzip.extract(DATA_CSV_FILENAME)
     latest_data_gzip.close()
 
-    return os.path.join(
-        repo_path, DATA_REPO_PATH, DATA_CSV_FILENAME)
+    return DATA_CSV_FILENAME
 
 
 def read_csv(infile: str) -> DataFrame:
--- a/data-serving/scripts/convert-data/convert_data.py
+++ b/data-serving/scripts/convert-data/convert_data.py
@@ -6,6 +6,7 @@ schema.
 import argparse
 import csv
 import logging
+import imp
 import json
 import pandas as pd
 import sys
@@ -16,7 +17,11 @@ from converters import (
     convert_outbreak_specifics, convert_travel_history)
 from pandas import DataFrame
 from typing import Any
-from constants import LOSSLESS_FIELDS
+from constants import (
+    DATA_CSV_FILENAME, DATA_GZIP_FILENAME, DATA_REPO_PATH, GEOCODER_DB_FILENAME,
+    GEOCODER_MODULE, GEOCODER_REPO_PATH, LOSSLESS_FIELDS)
+import tarfile
+import os
 
 
 def main():
@@ -26,7 +31,7 @@ def main():
     parser = argparse.ArgumentParser(
         description='Convert CSV line-list data into json compliant with the '
         'MongoDB schema.')
-    parser.add_argument('--infile', type=argparse.FileType('r'))
+    parser.add_argument('--ncov2019_path', type=str, required=True)
     parser.add_argument('--outfile',
                         nargs='?',
                         type=argparse.FileType('w', encoding='UTF-8'),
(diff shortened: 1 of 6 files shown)


data-serving/scripts/convert-data/geocode_util.py
```diff
@@ -0,0 +1,200 @@
...
+
+    raise ValueError('Too many tokens in location')
+
+
+def lookup_two_part_location(
+        geocoder: Any, higher_res_location_token: str,
+        lower_res_location_token: str) -> Any:
+    '''
+    Attempts to match the two location tokens against known locations,
+    trying two-token pairs from broadest to most specific resolution. First we
```
axmb, line 67: Why coarsest to finest? To avoid over-specifying in the event of a collision (on the basis that we'd rather be generally wrong than specifically wrong)? Seems weird to not error here in the event of multiple hits, TBH.

data-serving/scripts/convert-data/geocode_util.py
```diff
@@ -0,0 +1,200 @@
...
+
+    if city_and_province:
+        return city_and_province
+
+    return None
+
+
+def lookup_single_part_location(geocoder: Any, value: str) -> Any:
+    '''
+    Attempts to match the location token against known locations, beginning from
```
axmb, line 121: Would the "Paris" confusion be a result of granularity direction? IOW, aren't both Paris cities? Are we representing likelihood some other way?

data-serving/scripts/convert-data/geocode_util.py
```diff
@@ -0,0 +1,200 @@
...
+        }
+    '''
+    matches = [
+        geocode for id, geocode in geocoder.geocodes.items()
+        if predicate(geocode)]
+
+    if len(matches) > 1:
+        raise ValueError(f'Too many possible geocode matches: {matches}')
+
+    return matches[0] if len(matches) == 1 else None
```
axmb, line 200: Another feel-free-to-ignore [Pythonism](https://docs.python.org/3/library/functions.html#next): `next(iter(matches), None)`

data-serving/scripts/convert-data/geocode_util.py
```diff
@@ -0,0 +1,200 @@
...
+    if normalized_tokens == []:
+        return None
+    if len(normalized_tokens) == 1:
+        # A location of a single token, which may be a country, city, county,
+        # etc. Ex. "Paris", "China", "FL"
+        return lookup_single_part_location(geocoder, normalized_tokens[0])
+    if len(normalized_tokens) == 2:
+        # A two-part location, which may be (county, country), (city, county),
+        # or (city, country) pairs. Ex. "Paris, France", "Hubei, China",
+        # "Boston, MA"
```
iamleeg, line 47: documentation nit: MA isn't currently a country :)

data-serving/scripts/convert-data/geocode_util.py
```diff
@@ -0,0 +1,200 @@
...
+            normalized_tokens[1])
+    if len(normalized_tokens) == 3:
+        # A three-part location, presumed to including city, county, and
+        # country.
+        return geocoder.geocode(
+            normalized_tokens[0],  # city
+            normalized_tokens[1],  # county
+            normalized_tokens[2])  # country
+
+    raise ValueError('Too many tokens in location')
```
iamleeg, line 59: Another nit: this has the feeling of being a precondition (i.e. caller has to get this right or they won't go into space today), so I'd put it at the top of the function before any other statements like normalizing the tokens. Partly because we don't need to do that work if the input is wrong, partly so that the whole method body (as it stands now, and as it may change later) can work with the assumption that the input is valid.

data-serving/scripts/convert-data/geocode_util.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+def lookup_two_part_location(
```
iamleeg, line 62: The `Any`s here can be `Union`s, the return type of the function is

```python
from typing import Optional, Union

Optional[Dict[str, Union[str,float]]]
```

`Optional[T]` being sugar for `Union[T, None]`.

data-serving/scripts/convert-data/geocode_util.py
iamleeg: Related: https://daringfireball.net/2020/05/what_time_is_it_in_london. Paris, TX and Paris, France are indeed both cities, and I don't know that it's a good idea to assume any particular Paris if it's not clear from context. There are 23 Parises in the US and a couple in Canada.

My reading of this code is that it does a right thing (I'm not feeling qualified to say _the_ right thing, but definitely _a_ right thing): if we get here with the value `Paris`, then we don't find a country or province called Paris, then we find a bajillion towns called Paris and raise `ValueError`.

However, if we get here because the single location token is `Georgia`, we'll choose the country as a coarser grain than the state, and I don't know if that's correct.

data-serving/scripts/convert-data/parsers.py
```diff
@@ -191,52 +191,42 @@ def parse_location(value: Any) -> Dict[str, Any]:
...
+    # try to find a match with the geocoder.
+    geocode_result = lookup_location(geocoder, parse_list(value.lower(), ','))
 
-    # First try to look up the location in a hard-coded map that accounts for
-    # most popular locations that the lookup (below) misses.
-    location = LOCATIONS.get(normalized_location)
-    if location is not None:
-        return location
+    if geocode_result is None:
+        raise ValueError('no geocode found for location')
```
iamleeg, line 209: The geocode functions raise `ValueError` if more than one result is found, and return `None` if less than one result is found, which then gets converted here into `ValueError`. Can these paths be simplified to "raise `ValueError` if there is not exactly one geocode match"?

khmoran:
> Just curious -- is the precision of the geocoder pretty good? On one hand we're already using it in the sheet; on the other hand, we're trying to achieve a higher level of data validity/fidelity here.

yeah it seems pretty decent to me -- the main limitation is just the breadth (i.e. it doesn't have paris, tx 😝), since it will generally encompass data that's been seen in the location field of the dataset so far. but i don't think that's going to be a major issue here, and if a row is missing, this script will log an error and we can add it.

what's nice about the dataset is that it does have thinks broken down by geo precision, so it's easy to get the correct data (like centroid) for *just* the country

data-serving/scripts/convert-data/parsers.py
```diff
@@ -191,52 +191,42 @@ def parse_location(value: Any) -> Dict[str, Any]:
...
-          'latitude': float,
-          'longitude': float
+            'latitude': float,
+            'longitude': float
           }
         }
     '''
     if pd.isna(value) or value.lower() in ['no', 'none', 'unknown']:
-        return None
+        raise ValueError('boolean value is not a valid location')
```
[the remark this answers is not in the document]
khmoran: i'll let this error through the rest of the function instead of special-casing it

data-serving/scripts/convert-data/geocode_util.py
```diff
@@ -0,0 +1,200 @@
...
+          country_new: str
+          admin_id: int
+          location: str
+          admin3: str
+          admin2: str
+          admin1: str
+        }
+    '''
+    normalized_tokens = [COMMON_LOCATION_ABBREVIATIONS[l.lower()]
+                         if l.lower() in COMMON_LOCATION_ABBREVIATIONS else l
```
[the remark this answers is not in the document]
khmoran: nice, thanks!

data-serving/scripts/convert-data/geocode_util.py
```diff
@@ -0,0 +1,200 @@
...
+                         for l in location_tokens]
+
+    if normalized_tokens == []:
```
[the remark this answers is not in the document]
khmoran: done

data-serving/scripts/convert-data/geocode_util.py
```diff
@@ -0,0 +1,200 @@
...
+        return lookup_single_part_location(geocoder, normalized_tokens[0])
+    if len(normalized_tokens) == 2:
+        # A two-part location, which may be (county, country), (city, county),
+        # or (city, country) pairs. Ex. "Paris, France", "Hubei, China",
+        # "Boston, MA"
+        return lookup_two_part_location(
+            geocoder, normalized_tokens[0],
+            normalized_tokens[1])
+    if len(normalized_tokens) == 3:
+        # A three-part location, presumed to including city, county, and
```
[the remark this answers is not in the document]
khmoran: spellchecker needs grammar feature

data-serving/scripts/convert-data/geocode_util.py
khmoran: fair enough -- done (see more details below)

data-serving/scripts/convert-data/geocode_util.py
khmoran: OK I did pick a terrible, terrible example -- in reality, Paris, TX doesn't actually raise an issue here because it's not in this location db -- I'll clean up the comment

*However*: the bigger question you raised is good, and I've been back and forth on it. @iamleeg your example about Georgia is spot on. I was trying to save cycles by greedily matching but it's probably not worth the loss in precision, so I'll attempt all matches for the # of tokens, and if there's more than one result, I'll throw an error about disambiguation.

The problem with this approach is that there isn't really a way to "clarify" Georgia the country -- when I send the logs to the curators to fix ambiguous locations, what would they put in there to indicate that it's NOT the state? We'll have to figure out how to fix this and I'm open to suggestions.

data-serving/scripts/convert-data/geocode_util.py
khmoran: thanks!

data-serving/scripts/convert-data/geocode_util.py
khmoran: hah true -- but this is no paris, TX error on my part. I think these two-token locations *can* be (city, county) -- but country looks a lot like county, so I'll s/county/province in the comments here

data-serving/scripts/convert-data/geocode_util.py
khmoran: good point, done!

data-serving/scripts/convert-data/geocode_util.py
khmoran: done! also made it a quasi-typedef since I use it so many places

data-serving/scripts/convert-data/parsers.py
khmoran: yeah fair -- i tried to simplify quite a bit by note using None at all -- just ValueErrors (and tried to centralize that logic). LMK what you think

data-serving/scripts/convert-data/convert_data.py
```diff
@@ -35,8 +40,27 @@ def main():
 
     args = parser.parse_args()
 
-    print('Reading data from', args.infile.name)
-    original_cases = read_csv(args.infile)
+    # Load the geocoder.
```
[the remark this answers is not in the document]
khmoran: done

khmoran:

data-serving/scripts/convert-data/geocode_util.py
iamleeg: I saw some "convenient" abbreviations in `constants.py`, can the ISO 3166-2 country code for Georgia (`ge`) be added? Then `ge` matches Georgia the country, `ga` matches Georgia, US, and hopefully that's the only example where we need to make people do that.

Sadly `ga` is the ISO code for Gabon, but we can ask people to write "Gabon" when they mean that. Gah!

khmoran: I think this will take a bit more work because of the disambiguation errors now -- 'ge' will return 'georgia', but then we will raise an error because we don't know if 'georgia' is the country or the state. So I think I may need to rework the abbreviations, perhaps to map to the correct result ID instead of just an abbreviation (brittle, maybe, but more precise). I'll send a separate PR for this today since this changes the flow in `geocode_util`.

PR #214, opened 2020-06-02
Associate changes to cases with logged in user

allysonjp715:


allysonjp715:
[Associate logged in user with edited/created cases](https://trello.com/c/MYSGS8a4/173-associate-logged-in-user-with-edited-created-cases)

PR #216, opened 2020-06-02
Update the setup-db script to take in an optional connection string

khmoran:
This is the first step in letting us trigger updates to the prod db; this will be one of the scripts called by the workflow when the CSV data is updated.

Specifically, after this change, I can update prod MongoDB with the latest CSV by:

1. Git pull on the ` nCoV2019` repo
2. Running `healthmap-gdo-temp/data-serving/scripts/convert-data.py`
3. Running this script (in `healthmap-gdo-temp/data-serving/scripts/setup-db`)
4. Running `mongoimport` on the results of (2)

I'll chain these things together in a master script that will get triggered by the webhook.

And you get a bonus small schema change! Issue discovered after setting up the prod db and importing the newest data. You're welcome!

TESTED=Successfully updated the prod db using this script

### DIFF
--- a/data-serving/data-service/schemas/cases.schema.json
+++ b/data-serving/data-service/schemas/cases.schema.json
@@ -179,9 +179,6 @@
         "items": {
           "bsonType": "object",
           "additionalProperties": false,
-          "required": [
-            "location"
-          ],
           "properties": {
             "_id": {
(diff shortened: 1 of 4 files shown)


khmoran:

PR #217, opened 2020-06-03
Populate rule description in our AWS CW events client library

axmb:
Small QoL change as I'm getting down to brass tacks and configuring rules.

ds1


axmb:

verification/curator-service/api/src/model/source.ts
```diff
@@ -23,11 +23,17 @@ const sourceSchema = new mongoose.Schema({
     },
 });
 
+sourceSchema.methods.toAwsRuleDescription = function (): string {
```
khmoran, line 26: nice!

PR #219, opened 2020-06-03
build and push images to github packages

attwad:


attwad:
[Move to github packages instead of docker hub.](https://trello.com/c/SfZI7WkB/187-move-to-github-packages-instead-of-docker-hub)

PR #220, opened 2020-06-03
Add a page for cumulative charts

allysonjp715:
I will add the Dataset freshness metrics and Dataset completeness metrics in the next PR.

Screen Shot 2020-06-03 at 11 12 27 AM

### DIFF
--- a/verification/curator-service/ui/src/components/App.tsx
+++ b/verification/curator-service/ui/src/components/App.tsx
@@ -3,6 +3,7 @@ import { Link, Route, Switch } from 'react-router-dom';
 import EpidNavbar from './EpidNavbar';
 import LinelistTable from './LinelistTable';
 import React from 'react';
+import CumulativeCharts from './CumulativeCharts';
 import SourceTable from './SourceTable';
 import Profile from './Profile';
 import Users from './Users';
@@ -79,6 +80,11 @@ class App extends React.Component<{}, User> {
                                 <SourceTable />
                             </Route>
                         )}
+                        {this.hasAnyRole(['reader']) && (
+                            <Route path="/charts/cumulative">
+                                <CumulativeCharts />
+                            </Route>
+                        )}
                         {this.state.email && (
                             <Route path="/profile">
                                 <Profile user={this.state} />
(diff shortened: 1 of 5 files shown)


allysonjp715:
[Embed dataset charts in the curator UI](https://trello.com/c/4MtmQZ6l/141-embed-dataset-charts-in-the-curator-ui)

verification/curator-service/ui/src/components/CumulativeCharts.tsx
```diff
@@ -0,0 +1,119 @@
...
+        const countBySexChart = sdk.createChart({
+            chartId: 'ef5b9e34-e6f4-4f44-bf5f-5fea92971a0a',
+            height: '200px',
+        });
+        const geospatialTravelHistoryChart = sdk.createChart({
+            chartId: 'bc8fa65e-ff70-46ff-b607-8781de4c0b2d',
+            height: '400px',
+        });
+
+        await totalNumberCasesChart.render(
```
attwad, line 60: how fast are these render calls? if it's slow (as in if charts showing up one by one on the screen and it's noticeable) then you can use promise.all to run the in parallel.

verification/curator-service/ui/src/components/App.tsx
```diff
@@ -79,6 +80,11 @@ class App extends React.Component<{}, User> {
                                 <SourceTable />
                             </Route>
                         )}
+                        {this.hasAnyRole(['reader']) && (
```
attwad, line 83: client side role checks won't give us much as there's no "reader" role check when trying to access the charts on the mongo side. If somebody has the URL they can view it, so I'm not sure we need this check here it just seems to add toil as we'll have to ask admins to set the reader role on people.
Consider removing?

allysonjp715: Sounds good, done.

verification/curator-service/ui/src/components/CumulativeCharts.tsx
allysonjp715: Good call, they are pretty slow. The awaits aren't necessary and removing those allows them to load in parallel, so that's what I've done.

PR #222, opened 2020-06-03
Improve reporting of conversion errors

khmoran:
This lets us easily import the results into a spreadsheet to share with the folks who will be fixing them

- Format as TSV for easy import into sheet
- Filter out logs from geocoding script by setting log level to error
- Include original field names so that folks correcting original data know where to look


khmoran:

PR #225, opened 2020-06-04
Conversion script: Improve the specificity of the location abbreviation lookups

khmoran:
Map abbreviated locations to full location tuples, which can then be used as precise inputs to the geocoder.

This allows us to (1) provide multiple-token abbreviations, and to (2) allow us to fully disambiguate a given abbreviation; previously, "GA" was mapped to "Georgia", but then "Georgia" itself is ambiguous; however, the tuple (None,'Georgia','United States') is unambiguous.

I also added a test for the geocoding util because WHY NOT

This reduces 'Ambiguous location' errors from 88 instances to 62.


khmoran:

data-serving/scripts/convert-data/constants.py
```diff
@@ -24,69 +24,74 @@
...
-    'tx': 'texas',
-    'ut': 'utah',
-    'vt': 'vermont',
-    'vi': 'virgin islands',
-    'va': 'virginia',
-    'wa': 'washington',
-    'wv': 'west virginia',
-    'wi': 'wisconsin',
-    'wy': 'wyoming',
+    ('wuhan',): ('Wuhan City', 'Hubei', 'China'),
```
iamleeg, line 32: this is a great format for expanding the abbreviations!

PR #226, opened 2020-06-04
Add OpenAPI for curator service api

allysonjp715:
I wasn't sure whether or not to add /cases to the yaml, since it is already documented/verified in the data-service OpenAPI. I refrained for now, but let me know if you think it would be useful to add here as well.

Screen Shot 2020-06-04 at 11 27 57 AM

### DIFF
--- a/verification/curator-service/api/openapi/openapi.yaml
+++ b/verification/curator-service/api/openapi/openapi.yaml
@@ -183,12 +183,12 @@ paths:
       summary: Updates a specific user's roles
       operationId: updateUser
       requestBody:
-        description: User to update
+        description: New roles for the user
         required: true
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/User"
+              $ref: "#/components/schemas/Roles"
       responses:
         "200":
           $ref: "#/components/responses/200User"
@@ -223,6 +223,10 @@ components:
       type: array
       items:
         $ref: "#/components/schemas/User"
+    Roles:
+      type: array
+      items:
+        type: string
   responses:
     "200Source":
       description: OK
--- a/verification/curator-service/api/openapi/openapi.yaml
+++ b/verification/curator-service/api/openapi/openapi.yaml
@@ -12,6 +12,7 @@ info:
 paths:
   /sources:
     get:
+      tags: [Source]
       summary: Lists sources
       operationId: listSources
       parameters:
@@ -46,6 +47,7 @@ paths:
         "500":
           $ref: "#/components/responses/500"
     post:
+      tags: [Source]
       summary: Creates a new source
       operationId: createSource
       requestBody:
@@ -76,6 +78,7 @@ paths:
           type: string
           pattern: '^[a-f\d]{24}$'
     get:
+      tags: [Source]
       summary: Gets a specific source
       operationId: getSource
       responses:
@@ -92,6 +95,7 @@ paths:
         "500":
           $ref: "#/components/responses/500"
     put:
+      tags: [Source]
       summary: Updates a specific source
       operationId: updateSource
       requestBody:
@@ -115,6 +119,7 @@ paths:
         "500":
           $ref: "#/components/responses/500"
     delete:
+      tags: [Source]
       summary: Deletes a specific source
       operationId: deleteSource
       responses:
@@ -130,6 +135,7 @@ paths:
           $ref: "#/components/responses/500"
   /users:
     get:
+      tags: [User]
       summary: Lists users
       operationId: listUsers
       parameters:
@@ -173,6 +179,7 @@ paths:
           type: string
           pattern: '^[a-f\d]{24}$'
     put:
+      tags: [User]
       summary: Updates a specific user's roles
       operationId: updateUser
       requestBody:
--- a/verification/curator-service/api/openapi/openapi.yaml
+++ b/verification/curator-service/api/openapi/openapi.yaml
@@ -1,0 +1,262 @@
+openapi: 3.0.3
+info:
+  title: EPID curator service API
+  description: Server for CRUD operations on EPID data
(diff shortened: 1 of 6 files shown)


allysonjp715:
[Setup OpenAPI as done for Data server](https://trello.com/c/HyEH2GoM/172-setup-openapi-as-done-for-data-server)

verification/curator-service/api/openapi/openapi.yaml
```diff
@@ -0,0 +1,262 @@
...
+  description: Server for CRUD operations on EPID data
+  termsOfService: "https://www.healthmap.org/about/termsofuse/"
+  contact:
+    email: info@healthmap.org
+  license:
+    name: MIT
+    url: "https://opensource.org/licenses/MIT"
+  version: 1.0.0
+paths:
+  /sources:
```
axmb, line 13: Since we have multiple resources, it'd be nice to add tags for the resource (`Source`; `User`). Then, in the UI, we'd have grouped method definitions by resource. WDYT?

khmoran: +1

khmoran:
> I wasn't sure whether or not to add /cases to the yaml, since it is already documented/verified in the data-service OpenAPI. I refrained for now, but let me know if you think it would be useful to add here as well.

I guess we should? I'm guessing they'll have the flexibility to diverge at a certain point, as each service becomes more tailored to its individual purpose. But we don't have to do it now. @axmb WDYT

verification/curator-service/api/package.json
```diff
@@ -51,16 +51,22 @@
     "typescript": "^3.9.3"
   },
   "dependencies": {
+    "@types/multer": "^1.4.3",
```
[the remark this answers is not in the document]
allysonjp715: I'm not sure why multer is needed... But the stack doesn't build without it. For the passport line, that just had a comma added to the end of it.

verification/curator-service/api/openapi/openapi.yaml
```diff
@@ -0,0 +1,262 @@
...
+          $ref: "#/components/responses/404"
+        "422":
+          $ref: "#/components/responses/422"
+        "500":
+          $ref: "#/components/responses/500"
+    delete:
+      summary: Deletes a specific source
+      operationId: deleteSource
+      responses:
+        "204":
```
khmoran, line 121: I found this with the data service too -- currently it looks like this is returning a 200 OK for success instead of 204, and it's returning the Source object with the 200:

https://github.com/open-covid-data/healthmap-gdo-temp/blob/master/verification/curator-service/api/src/controllers/sources.ts#L158

you can see my change for this in the data service here: https://github.com/open-covid-data/healthmap-gdo-temp/pull/197/files

verification/curator-service/api/openapi/openapi.yaml
```diff
@@ -0,0 +1,262 @@
...
+        in: path
+        description: The user ID
+        required: true
+        schema:
+          type: string
+          pattern: '^[a-f\d]{24}$'
+    put:
+      summary: Updates a specific user's roles
+      operationId: updateUser
+      requestBody:
```
khmoran, line 178: it looks like the body of this request is currently a list of roles, not the whole user obj: https://github.com/open-covid-data/healthmap-gdo-temp/blob/master/verification/curator-service/api/src/controllers/users.ts#L58

axmb:
> > I wasn't sure whether or not to add /cases to the yaml, since it is already documented/verified in the data-service OpenAPI. I refrained for now, but let me know if you think it would be useful to add here as well.
> 
> I guess we should? I'm guessing they'll have the flexibility to diverge at a certain point, as each service becomes more tailored to its individual purpose. But we don't have to do it now. @axmb WDYT

It should probably be documented, so that the OpenAPI page is a complete list of endpoints provided by the service. I don't think it has to be done now, either. Perhaps there's a simple way to create a section for this, and say "see <:3000/api/docs>" until they actually diverge.

verification/curator-service/api/openapi/openapi.yaml
khmoran: beautiful!

verification/curator-service/api/openapi/openapi.yaml
allysonjp715: Ah, thanks. Done.

verification/curator-service/api/openapi/openapi.yaml
allysonjp715: Ah yea I was thinking it's possible to pass the whole user object, and we'll only access the roles field. But it's worth making it explicit that roles is the only field we're using. Done.

allysonjp715:
> > > I wasn't sure whether or not to add /cases to the yaml, since it is already documented/verified in the data-service OpenAPI. I refrained for now, but let me know if you think it would be useful to add here as well.
> > 
> > 
> > I guess we should? I'm guessing they'll have the flexibility to diverge at a certain point, as each service becomes more tailored to its individual purpose. But we don't have to do it now. @axmb WDYT
> 
> It should probably be documented, so that the OpenAPI page is a complete list of endpoints provided by the service. I don't think it has to be done now, either. Perhaps there's a simple way to create a section for this, and say "see <:3000/api/docs>" until they actually diverge.

The swagger docs say you can use a $ref for a path:
Screen Shot 2020-06-04 at 6 01 43 PM

And the $ref docs say you can use a relative path to another yaml file:
Screen Shot 2020-06-04 at 6 05 10 PM

But there is a bug in openapi v3 that relative paths aren't working: https://github.com/swagger-api/swagger-editor/issues/1409

That bug was assigned to someone 12 days ago, so maybe they'll make some progress there. I'll wait to see if there are any updates there and add the ref for cases here if it gets fixed.

PR #228, opened 2020-06-04
Wire up the source API to set rule targets to the global retrieval Lambda ARN

axmb:
Includes a couple changes to the AWS CWE client itself. Mostly to keep logic as AWS-oriented as possible, and leaving the intepretation from source data model-to-AWS params in the source controller (and source data model custom instance methods).


axmb:

PR #229, opened 2020-06-05
Add a script that runs the full data conversion & import pipeline

khmoran:
Hoping we can trigger this script from a github action -- LMK if you foresee any issues with that.


khmoran:

data-serving/scripts/data-pipeline/run_pipeline.sh
```diff
@@ -0,0 +1,102 @@
...
+    else
+        print "Cloning nCoV2019 repo to $NCOV2019_REPO_PATH"
+        git clone https://github.com/beoutbreakprepared/nCoV2019.git \
+            -- $NCOV2019_REPO_PATH
+    fi
+}
+
+function convert_data() {
+    print 'Converting the latest data archive to the MongoDB schema'
+    python3 $SCRIPT_PATH/../convert-data/convert_data.py \
```
axmb, line 50: If you want to get fancy, you can create an appropriate venv and `source /path/to/activate` prior to this call. Probably not worth, but just in case you haven't had enough fun with this bash script yet...

data-serving/scripts/data-pipeline/run_pipeline.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
```
axmb, line 1: This is a very well written bash script.

data-serving/scripts/data-pipeline/README.md
```diff
@@ -0,0 +1,18 @@
...
+
+Have you ever wanted to import the freshest possible data directly into a
+database?
+
+This script is for you.
+
+## How to run it
+
+```shell
+./run_pipeline.sh [-m <mongodb_connection_string>] [-d <database>] [-c <collection>] [-r <sample_rate>]
```
attwad, line 13: can we name that script with what it does? "running a pipeline" is open to interpretation


image

I suggest
fetch_latest_data_convert_data_setup_db_import_data_cleanup.sh

data-serving/scripts/data-pipeline/run_pipeline.sh
attwad: meh venv, get fancier and just use docker instead :) you get a clean env for sure this way and we all have it installed ;)

data-serving/scripts/data-pipeline/run_pipeline.sh
attwad: This goes above my understanding of bash but I approve

data-serving/scripts/data-pipeline/run_pipeline.sh
khmoran: I thought about it but I couldn't tell if this was a good use case for docker, since I'm not running any services. wdyt? (not saying I'm going to change it now but if it's worth it, down the road!)

data-serving/scripts/data-pipeline/run_pipeline.sh
khmoran: you honor and somehow simultaneously shame me

data-serving/scripts/data-pipeline/README.md
khmoran: hah so you DO see why I named it that!

how about... convert_and_import_latest_data.sh?

attwad: I like that proposal

data-serving/scripts/data-pipeline/run_pipeline.sh
attwad: We all have python3 installed on our machines so it's not a big deal but being able to do `docker run kellyrepo/datapipeline` without having to install anything else would be nice for bigger projects.

PR #231, opened 2020-06-05
s/aws-alb/kubernetes-nginx and s/fargate/managedNodeGroups

attwad:
Demos:
- https://curator.ghdsi.org/
- https://dev-curator.ghdsi.org/

Why nginx:
- much simpler than ALB setup
- well known/maintained
- kubernetes-maintained nginx controller is better documented than the AWS ALB ingress controller
- we don't need an L7 balancer with WAF support, request inspection and deep integration with AWS services, a Network Load Balancer such as nginx is more performant, more common and doesn't lock us down in AWS-land (can basically be re-used as is)

Why no more fargate:
- Network Load Balancers are not supported by fargate (https://github.com/aws/containers-roadmap/issues/617), this makes it impossible to use cert-manager to get certs from let's encrypt because cert-manager spawns an nginx ingress to respond to the let's encrypt challenges. There _is_ support for TLS termination with AWS ALB controller if you deep integrate with route 53 and have your DNS managed by amazon, honestly it's not worth the added complexity, you can't beat let's encrypt on simplicity.
- We were actually getting much lower utilization on fargate and it was spawning a crazy bunch of nodes for no good reason. With server metrics installed this became obvious, on fargate:

```
kubectl top nodes
NAME                                      CPU(cores)   CPU%   MEMORY(bytes)   MEMORY%
fargate-ip-192-168-107-98.ec2.internal    13m          0%     315Mi           8%
fargate-ip-192-168-109-194.ec2.internal   17m          0%     171Mi           4%
fargate-ip-192-168-110-83.ec2.internal    14m          0%     179Mi           4%
fargate-ip-192-168-117-236.ec2.internal   13m          0%     179Mi           4%
fargate-ip-192-168-124-31.ec2.internal    15m          0%     159Mi           4%
fargate-ip-192-168-64-207.ec2.internal    13m          0%     345Mi           9%
fargate-ip-192-168-81-109.ec2.internal    15m          0%     312Mi           8%
fargate-ip-192-168-85-164.ec2.internal    16m          0%     317Mi           8%
fargate-ip-192-168-87-55.ec2.internal     12m          0%     166Mi           4%
fargate-ip-192-168-93-139.ec2.internal    14m          0%     164Mi           4%
```

On managed cluster nodes:

```
kubectl top nodes
NAME                             CPU(cores)   CPU%   MEMORY(bytes)   MEMORY%
ip-192-168-10-235.ec2.internal   79m          4%     547Mi           37%
ip-192-168-54-58.ec2.internal    91m          4%     662Mi           44%
```

- managed cluster nodes apparently give you more or less what fargate gives you, that is managed nodes that you don't have to care about... I picked the `t3.small` instance type that has 2vCPUs and 2GiB of RAM. It costs $0.0209 an hour which is 15$ per month, I'm running two of those so the cost of the cluster is now 2*15+74(master at 0.1$/h) ~= 100$/month.
I let it scale up to 5 nodes max for now I don't expect that we'll need much more really, even then that'll cost 150$/month.
- cluster is now defined in a cluster config file to be consistent with how other kubernetes resources are defined.

I'll tear-down the `healthmapidha` cluster in favor of this new `ghdsi` cluster after this PR is merged in.


attwad:
[expose https endpoints on aws](https://trello.com/c/bXHPGNBc/193-expose-https-endpoints-on-aws)

aws/README.md
```diff
@@ -8,12 +8,14 @@ Install `eksctl` and `kubectl` to interact with Amazon EKS and the Kubernetes co
 
 You can learn more about `eksctl` [here](https://eksctl.io/).
 
-To configure kubectl to talk to the healthmapidha cluster, do:
+To configure kubectl to talk to the ghdsi cluster, do:
 
 ```shell
-aws eks --region us-east-1 update-kubeconfig --name healthmapidha
+aws eks --region us-east-1 update-kubeconfig --name ghdsi
```
axmb, line 14: I can't connect via kubectl. Did you update the `aws-auth` ConfigMap? If so, could be a problem with my local authenticator, but checking to see if this was done first.

aws/curator-ingress.yaml
```diff
@@ -1,40 +1,33 @@
 # Ingress defines how load balancers should forward traffic
 # from the internet traffic to the cluster.
 # https://kubernetes.io/docs/concepts/services-networking/ingress/
-apiVersion: extensions/v1beta1
+apiVersion: networking.k8s.io/v1beta1
 kind: Ingress
 metadata:
-  name: curator-dev
+  name: curator
```
axmb, line 7: nit: curator-ingress?

aws/letsencrypt.yaml
```diff
@@ -0,0 +1,39 @@
+# Cf. https://cert-manager.io/docs/configuration/acme/
+apiVersion: cert-manager.io/v1alpha2
+kind: Issuer
```
axmb, line 3: Seems like `Issuer`s should be namespaced (else you'd use a `ClusterIssuer`)?

aws/curator-ingress.yaml
attwad: I removed it on purpose because it felt redundant, I'm not adding -service or -deployment to the other objects so this is consistent.

aws/letsencrypt.yaml
attwad: The ingress for curator service is defined in the default namespace which is also where the deployments live, the requirement is that Issuer must be in the same namespace as the ingress one so this is fine as I understand it.

attwad:
Yes I suspect the ALB is also costlier because we had to run two balancers whereas here we only have one.
Good call on adding Spencer, done.

aws/README.md
attwad: Solved, and added documentation on how to get access in the readme.

PR #232, opened 2020-06-05
Creates github workflow to execute the data update pipeline against prod mongodb

khmoran:
Note that is only run on updates to the data conversion script & on a once-a-day schedule -- the next step is to figure out the webhook piece

TESTED=Using https://github.com/nektos/act, ran:
`act -j update-case-data-prod -P ubuntu-latest=nektos/act-environments-ubuntu:18.04-full -s DB_CONNECTION_URL=<db_conn_url>`

### DIFF
--- a/.github/workflows/case-data-update.yml
+++ b/.github/workflows/case-data-update.yml
@@ -1,0 +1,37 @@
+name: Update case data
+
+on:
+  push:
+    branches: [master]
+    paths: ["/data-serving/scripts/convert-data/**"]
+
+jobs:
+  update-case-data:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install MongoDB CLIs
+        run: |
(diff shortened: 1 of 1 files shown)


khmoran:

.github/workflows/case-data-update.yml
```diff
@@ -0,0 +1,41 @@
...
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.x"
+
+      - name: Install Python dependencies
+        run: python -m pip install --upgrade pandas
+
+      - name: Run update script
+        run: ./data-serving/scripts/data-pipeline/convert_and_import_latest_data.sh -m mongodb+srv://${{ secrets.MONGODB_USER }}:${{ secrets.MONGODB_PASS }}@covid19-map-cluster01-sc7u9.mongodb.net
```
attwad, line 41: Hmmm I'm not a fan of exposing the db url like that, why not having one secret db_connection_url instead that contains everything as we do in the other services?

khmoran: good call, thanks for pointing this out!

PR #233, opened 2020-06-05
Improve performance of convert_data.py

khmoran:
Now that the CSV has over 2M rows, the script is taking a long time to run. Especially since it will be run from a GitHub action, I wanted to optimize it a bit. I figured that reducing the amount of json in memory would help, so I'm now reading in 1 row at a time and writing out 1 case at a time. It unfortunately requires a little JSON hackery but it's not too bad.

Since we're doing manipulations only on a single row, and not across columns, I think pandas isn't really speeding things up, so I switched to a pure CSV reader (it's a bit simpler).

The current dataset (2M+ rows) converts on my machine in:
- 9min16s with this version
- 39m44s with the prior version is still running 

Note that the sampling is no longer random; it's systematic. (Which maybe is preferable for diffing changes across runs tbh.)

LMK what you think!

### DIFF
--- a/.github/workflows/case-data-update.yml
+++ b/.github/workflows/case-data-update.yml
@@ -35,7 +35,7 @@ jobs:
           python-version: "3.x"
 
       - name: Install Python dependencies
-        run: python -m pip install --upgrade pandas
+        run: python -m pip install --upgrade progressbar
 
       - name: Run update script
         run: ./data-serving/scripts/data-pipeline/convert_and_import_latest_data.sh -m ${{ secrets.DB_CONNECTION_URL }}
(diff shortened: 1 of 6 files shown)


khmoran:

khmoran:
> The diff is pretty telling -- we're mostly just writing Python business logic that's applied at a very high level by Pandas. Pandas _can_ be fast if the work has a high ratio of vectorized operations to Python business logic, but we've got plenty of business logic.
> 
> I bet we could performance hack this by a few X (better typing, cython, etc.), but if we're under 10 minutes by just using plain ol' Python, that sounds like it's good enough. Are you pretty happy with ~9 minute runs?
> 
> One way to somewhat confirm your suspicion re: memory pressure would be to operate on the data [in chunks](https://towardsdatascience.com/why-and-how-to-use-pandas-with-large-data-9594dda2ea4c). What's the file size of the csv?

I think I'm OK with 9 min runs TBH. I was mostly concerned because the slowdown seemed nonlinear (again assuming because of memory pressure), and now it *ought* to be linear-ish, so there's room for growth without it being ridic.

And this "should be temporary" :D

Def agreed that pandas would make a lot more sense if we were doing vectorized operations. I think it just ended up being all row-wise. The current csv is 442M so not sure this is a pandas problem per se -- I think it probably just got rough to be doing so many row lookups. WDYT?

PR #234, opened 2020-06-08
Add an HTTP bearer token auth strategy for the curator server

axmb:
Since I'm adding a programmatic API consumer elsewhere (ingestion), I've found that the passport Google auth strategy is largely built around the browser flow. While it's conceivable we could _somehow_ manually spoof that flow, the canonical solution for API support is typically to accept bearer token auth.

I'm adding it as a fallback for RBAC'd requests in which there isn't an active user session. Passport's bearer strategy inspects the request for a token (in various RFC 6750 formats), and if present, will execute the strategy I defined.

While I was adding tests, I noticed that a handful of the existing auth tests have a classic async Jest bug -- not returning or awaiting the promise. I can clean this up in a separate PR; some of the cases didn't pass.

### DIFF
--- a/verification/curator-service/api/src/controllers/auth.ts
+++ b/verification/curator-service/api/src/controllers/auth.ts
@@ -25,6 +25,13 @@ export const mustBeAuthenticated = (
     res.sendStatus(403);
 };
 
+const userHasRequiredRole = (
+    user: UserDocument,
+    requiredRoles: Set<string>,
+): boolean => {
+    return user.roles?.filter((r: string) => requiredRoles.has(r)).length > 0;
+};
+
 /**
  * Express middleware that checks to see if there's an authenticated principal
  * that has any of the specified roles.
@@ -38,9 +45,7 @@ export const mustHaveAnyRole = (requiredRoles: string[]) => {
         const requiredSet = new Set(requiredRoles);
         if (
             req.isAuthenticated() &&
-            (req.user as UserDocument).roles?.filter((r: string) =>
-                requiredSet.has(r),
-            ).length > 0
+            userHasRequiredRole(req.user as UserDocument, requiredSet)
         ) {
             return next();
         } else {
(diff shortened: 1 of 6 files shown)


verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -14,27 +20,49 @@ export const mustBeAuthenticated = (
     next: NextFunction,
 ): void => {
     if (req.isAuthenticated()) {
-        next();
-        return;
+        return next();
```
attwad, line 23: interesting, I didn't know you could return a void function

verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -14,27 +20,49 @@ export const mustBeAuthenticated = (
...
     }
     res.sendStatus(403);
 };
 
-// mustHaveAnyRole is an express middleware that checks that the currently logged-in user has any of the given roles.
+const userHasRequiredRole = (
+    user: UserDocument,
+    requiredRoles: Set<string>,
+): boolean => {
+    return user.roles?.filter((r: string) => requiredRoles.has(r)).length > 0;
```
attwad, line 32: I wrote that but now I look again at it and I think we can use some() instead of building the whole filtered array and checking the length

verification/curator-service/api/src/model/user.ts
```diff
@@ -1,19 +1,12 @@
...
         type: String,
         required: 'User must have an email',
     },
-    googleID: {
-        type: String,
-        // Note: we can relax that requirement if we start
-        // supporting more identity platforms.
-        required: 'User must be logged-in with Google',
-    },
+    googleID: String,
```
attwad, line 9: don't you need a "token" here in this schema?

verification/curator-service/api/src/controllers/auth.ts
axmb: SG, done.

verification/curator-service/api/src/model/user.ts
axmb: Ah, thanks for the catch. I had initially persisted the token, but went back and thought better of it. We exchange it for an identity which we persist (the email addr), which lets us avoid persisting the token itself. I'll remove the dangling line in the auth controller.

axmb:
[Set up auth for programmatic access in the curator API service](https://trello.com/c/5NReBKCM/194-set-up-auth-for-programmatic-access-in-the-curator-api-service)

axmb:

PR #235, opened 2020-06-08
Add a /api/users/roles endpoint and use it in the UI

allysonjp715:


verification/curator-service/ui/src/components/Users.test.tsx
```diff
@@ -37,7 +37,14 @@ test('lists users', async () => {
         config: {},
         headers: {},
     };
-    mockedAxios.get.mockResolvedValueOnce(axiosResponse);
+    mockedAxios.get.mockImplementation((url) => {
+        if (url === '/api/users/roles') {
```
attwad, line 41: can you use a switch on the url please and default to reject the promise? this way we're sure we're mocking the right calls.
switch (url) {
  case roles: resolve roles
  case users: resolve users
  default: reject promise
}

verification/curator-service/ui/src/components/Users.test.tsx
```diff
@@ -37,7 +37,14 @@ test('lists users', async () => {
...
+            return Promise.resolve({
+                data: { roles: ['admin', 'curator', 'reader'] },
```
attwad, line 43: this promise data is similar in all mocked calls, consider refactoring it

verification/curator-service/ui/src/components/Users.test.tsx
allysonjp715: Done

verification/curator-service/ui/src/components/Users.test.tsx
allysonjp715: Done

PR #237, opened 2020-06-08
Update the case data update workflow to fix mongodb cli install

khmoran:
- The workflow failed to run over the weekend because the machine it's running on has newer versions of the mongodb packages installed. I think we want to downgrade to our versions because to ensure a consistent env, so we're now passing the `--allow-downgrades` arg to `apt-get install` (see one of the failed runs here: https://github.com/open-covid-data/healthmap-gdo-temp/runs/748668598)

- I got warnings locally about not being able to install 2 of the packages because their public keys weren't available, so I added the `apt-key adv` line to retrieve those keys from a key server

### DIFF
--- a/.github/workflows/case-data-update.yml
+++ b/.github/workflows/case-data-update.yml
@@ -21,8 +21,9 @@ jobs:
         run: |
           wget -qO - https://www.mongodb.org/static/pgp/server-4.2.asc | sudo apt-key add -
           echo "deb [ arch=amd64,arm64 ] https://repo.mongodb.org/apt/ubuntu xenial/mongodb-org/4.2 multiverse" | sudo tee /etc/apt/sources.list.d/mongodb-org-4.2.list
+          sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 5DC22404A6F9F1CA 656408E390CFB1F5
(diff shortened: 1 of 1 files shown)


khmoran:
> `--allow-downgrades` SG, but it'd be nice if there were a less fragile way to solve the second problem. Which packages, OOC?

This was the error:
```
| W: An error occurred during the signature verification. The repository is not updated and the previous index files will be used. GPG error: https://cli-assets.heroku.com/apt ./ InRelease: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 5DC22404A6F9F1CA
| W: GPG error: https://repo.mongodb.org/apt/ubuntu xenial/mongodb-org/4.2 Release: The following signatures couldn't be verified because the public key is not available: NO_PUBKEY 656408E390CFB1F5
| E: The repository 'https://repo.mongodb.org/apt/ubuntu xenial/mongodb-org/4.2 Release' is not signed.
| N: Updating from such a repository can't be done securely, and is therefore disabled by default.
| N: See apt-secure(8) manpage for repository creation and user configuration details.
```

Hopefully we won't be installing more packages than these but if we end up needing to do that, happy to look for a more robust solution.

khmoran:

PR #238, opened 2020-06-08
Add a repository_trigger event to the case-data-update workflow

khmoran:
This prepares our workflow action to run when we receive a repository_dispatch event from the nCoV2019 repo (which isn't yet sending them)


khmoran:

PR #241, opened 2020-06-09
use same site cookie strict for session cookie

attwad:
I was originally going for a full blown csrf token based approach (https://github.com/open-covid-data/healthmap-gdo-temp/compare/csrf?expand=1) but this was waaaaay too involved imo and was breaking the ingestion pipeline calls that Alex wrote yesterday.

You can see here how same site cookie helps defend against CSRF: https://portswigger.net/web-security/csrf/samesite-cookies

We are not using a lax framework in terms of POST becoming GET (express doesn't do that kind of change itself) so I am fine with just relying on samesite cookies as a starting point or even staying like this long-term.

We can also think about moving to "lax" mode if navigation from other sites becomes impractical, for now there should be no issues like that.

image

### DIFF
--- a/verification/curator-service/api/src/index.ts
+++ b/verification/curator-service/api/src/index.ts
@@ -1,12 +1,13 @@
 import * as usersController from './controllers/users';
 
 import { AuthController, mustHaveAnyRole } from './controllers/auth';
-import { OpenApiValidator } from 'express-openapi-validator';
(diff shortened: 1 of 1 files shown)


attwad:
[Add XSRF protection to API methods](https://trello.com/c/S9MNRYx2/160-add-xsrf-protection-to-api-methods)

PR #247, opened 2020-06-10
Upload coverage to codecov

attwad:
That's free for public repo and provides nice viz on PRs

### DIFF
--- a/.github/workflows/curator-api-node.yml
+++ b/.github/workflows/curator-api-node.yml
@@ -33,3 +33,8 @@ jobs:
         working-directory: verification/curator-service/api
         env:
           CI: true
+      - name: Upload coverage
+        uses: codecov/codecov-action@v1
+        with:
+          file: verification/curator-service/api/coverage-final.json
+          fail_ci_if_error: true
--- a/.github/workflows/curator-api-node.yml
+++ b/.github/workflows/curator-api-node.yml
@@ -33,7 +33,7 @@ jobs:
         working-directory: verification/curator-service/api
         env:
           CI: true
-      - name: Upload coverage
+      - name: Upload test coverage
         uses: codecov/codecov-action@v1
         with:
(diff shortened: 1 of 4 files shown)


.github/workflows/curator-api-node.yml
```diff
@@ -33,3 +33,8 @@ jobs:
         working-directory: verification/curator-service/api
         env:
           CI: true
+      - name: Upload coverage
```
allysonjp715, line 36: Is this about test coverage? If so, could we name it as so?

.github/workflows/curator-api-node.yml
```diff
@@ -33,3 +33,8 @@ jobs:
...
+        uses: codecov/codecov-action@v1
+        with:
+          file: verification/curator-service/api/coverage/coverage-final.json
+          fail_ci_if_error: true
```
allysonjp715, line 40: Is there an override option to be able to merge something without coverage if necessary?

attwad: github doesn't enforce CI to pass to be able to merge, you can --force anytime.
This will make the CI fail if the upload to codecov fails which I think is reasonable
https://github.com/marketplace/actions/codecov

.github/workflows/curator-api-node.yml
attwad: yup, done

.github/workflows/curator-api-node.yml
allysonjp715: Got it, thanks.

.github/workflows/curator-api-node.yml
allysonjp715: I find it confusing that the Codecov Report doesn't mention that it's test coverage anywhere. But I suppose there's nothing we can do about that.

PR #252, opened 2020-06-11
Add India parser function SAM definition; trigger source parser lambdas in retrieval

axmb:
Couple small notes:

- Small additional change to the retrieval function, so that I could generalize the SourceApiUrl SAM template param to work for both retrieval and parsing.
- Testing isn't super thorough for the AWS Lambda client. Gave it a fair shot, but it's a PITA to test -- even with the stub library I'm using to test s3 (`moto`). This code will probably change a few times in the next couple weeks, and be relatively stable thereafter, so I'm OK with saving the testing gore.


axmb:
[S#4: Write/deploy a parser Lambda to operate on the example source data](https://trello.com/c/sbulkaqd/185-s4-write-deploy-a-parser-lambda-to-operate-on-the-example-source-data)

axmb:
[S#4: Write/deploy a retrieval Lambda with support for the example source content type](https://trello.com/c/PuBedipv/184-s4-write-deploy-a-retrieval-lambda-with-support-for-the-example-source-content-type)

PR #254, opened 2020-06-11
Make navbar fixed and add a fixed table of contents (not yet interact…

allysonjp715:
…able) to the new case form.

I can't add a video to this comment, but here is the initial state:
Screen Shot 2020-06-11 at 3 21 29 PM

and the scrolled state:
Screen Shot 2020-06-11 at 3 21 38 PM

### DIFF
--- a/verification/curator-service/ui/src/components/NewCaseForm.tsx
+++ b/verification/curator-service/ui/src/components/NewCaseForm.tsx
@@ -2,6 +2,9 @@ import { Button, LinearProgress } from '@material-ui/core';
 import { Field, Form, Formik } from 'formik';
 import React from 'react';
 import { Select, TextField } from 'formik-material-ui';
+import { withStyles } from '@material-ui/core';
+import { Theme, createStyles } from '@material-ui/core/styles';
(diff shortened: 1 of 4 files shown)


verification/curator-service/ui/src/components/NewCaseForm.tsx
```diff
@@ -82,87 +96,112 @@ export default class NewCaseForm extends React.Component<
...
-                        >
-                            Submit case
-                        </Button>
-                        {this.state.errorMessage && (
-                            <h3> {this.state.errorMessage as string} </h3>
+                                    Submit case
+                                </Button>
+                                {this.state.errorMessage && (
+                                    <h3>
+                                        {' '}
```
attwad, line 194: what is this? some kind of emoji?
\\{'o'}/

just kidding, but maybe just use &nbsp; here instead or some left margin style if you want the error message indented?

allysonjp715: Oh weird, good catch. The formatter added those in because in the old version there were spaces between the h3 tags and the {this....}. Removed them.

PR #255, opened 2020-06-11
Set trust proxy to allow the use of secure session cookies behind nginx.

axmb:
Re: https://www.npmjs.com/package/express-session#cookiesecure


axmb:

PR #256, opened 2020-06-11
Bring demographics schema in line with data model standard spec

khmoran:
- Species has been nixed
- Profession, nationality, and ethnicity have been added: All optional string fields. In the future should be validated against a data dictionary.

In the future, profession and nationality will be validated against data dictionaries (no free text entry); ethnicity will have a data dictionary but the option to write in free text

### DIFF
--- a/data-serving/data-service/src/model/demographics.ts
+++ b/data-serving/data-service/src/model/demographics.ts
@@ -7,10 +7,6 @@ export enum Sex {
     Other = 'Other',
 }
 
-export enum Species {
-    HomoSapien = 'Homo sapien',
-}
-
 export const demographicsSchema = new mongoose.Schema({
     ageRange: {
         start: {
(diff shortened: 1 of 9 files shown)


khmoran:

data-serving/data-service/src/model/demographics.ts
```diff
@@ -24,18 +20,20 @@ export const demographicsSchema = new mongoose.Schema({
...
     },
-    species: {
-        type: String,
-        enum: Object.values(Species),
-    },
     sex: {
         type: String,
         enum: Object.values(Sex),
     },
+    // TODO: The below 3 fields should be data dictionaries.
```
axmb, line 27: Even profession? Seems tough.

data-serving/data-service/test/model/data/case.full.json
```diff
@@ -5,7 +5,9 @@
             "end": 59.0
         },
         "sex": "Female",
-        "species": "Homo sapien"
+        "profession": "Horse trainer",
```
axmb, line 8: My point exactly (re: dictionary). :-P

data-serving/data-service/src/model/demographics.ts
khmoran: Can ask Stephen but currently there's a whole list here: https://docs.google.com/spreadsheets/d/1qOnGJQg1gtbmNxAW8c9hC1TvFpIQmSnORD-x-iXrpJ0/edit?ts=5ee23b03#gid=1189434977

data-serving/data-service/test/model/data/case.full.json
khmoran: Could've gone with "Jockey – steeple", just saying.

PR #257, opened 2020-06-12
Remove the outbreakSpecifics field

khmoran:
Per the standard, the subfields of outbreakSpecifics will just live on in `importedCase`


khmoran:

PR #258, opened 2020-06-12
Bring revisionHistory schema in line with spec

khmoran:
Would like your feedback on this one. This is my interpretation of the flat list of fields: 

`
RevisionData | Date of creation | Date | ISO 8601
-- | -- | -- | --
RevisionData | Created by | String | Curator username
RevisionData | Date of edit | Date | ISO 8601
RevisionData | Edit made by | String | Curator username
RevisionData | Edit details | Text | None
`

It seems like we want the current version of the case to show original creation metadata (date/curator/notes) -- but also show the latest edit, if any. I think those two are sub-documents that share a schema.

I also took this opportunity to s/id/revisionNumber to be clearer, and s/moderator/curator, since I pretty much only hear the latter term used.

### DIFF
--- a/data-serving/data-service/src/model/revision-metadata.ts
+++ b/data-serving/data-service/src/model/revision-metadata.ts
@@ -1,17 +1,8 @@
 import { dateFieldInfo } from './date';
 import mongoose from 'mongoose';
 
-export const revisionMetadataSchema = new mongoose.Schema({
-    id: {
-        type: Number,
-        min: 0,
-        validate: {
-            validator: Number.isInteger,
-            message: '{VALUE} is not an integer value',
-        },
-        required: 'Enter a revision id',
-    },
-    moderator: {
+const editMetadataSchema = new mongoose.Schema({
+    curator: {
         type: String,
         required: 'Enter a revision moderator id',
     },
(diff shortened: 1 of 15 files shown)


khmoran:

ingestion/functions/parsing/india/india.py
```diff
@@ -72,9 +72,11 @@ def parse_cases(raw_data_file):
         return [
             {
                 "revisionMetadata": {
```
axmb, line 74: Thanks!

data-serving/data-service/src/model/revision-metadata.ts
```diff
@@ -1,8 +1,20 @@
 import { dateFieldInfo } from './date';
 import mongoose from 'mongoose';
 
+const editMetadataSchema = new mongoose.Schema({
+    curator: {
+        type: String,
+        required: 'Enter a revision moderator id',
```
axmb, line 7: Is this still the right phrasing?

data-serving/data-service/src/model/revision-metadata.ts
```diff
@@ -11,20 +23,29 @@ export const revisionMetadataSchema = new mongoose.Schema({
...
+        required: 'Enter creation metadata',
     },
-    date: {
-        ...dateFieldInfo,
-        required: 'Enter a revision date',
+    updateMetadata: {
+        type: editMetadataSchema,
+        required: [
+            function (this: RevisionMetadataDocument): boolean {
+                return this.revisionNumber > 0 && !this.updateMetadata;
```
axmb, line 34: Is the second part of this conditional necessary? Isn't this function the conditions under which the data is required?

data-serving/data-service/src/model/revision-metadata.ts
```diff
@@ -11,20 +23,29 @@ export const revisionMetadataSchema = new mongoose.Schema({
...
+            },
+            'Enter update metadata',
```
axmb, line 36: I've wondered -- is it required to supply text to `required`? If not, does this type of comment provide value?

data-serving/data-service/src/model/revision-metadata.ts
khmoran: nope -- will remove with the others

data-serving/data-service/src/model/revision-metadata.ts
khmoran: ugh no I ran into a weird bug and was just throwing stuff against the wall and forgot to remove. thanks!

data-serving/data-service/src/model/revision-metadata.ts
khmoran: good call, prob not, I'm actually going to remove it for all and just replace with required: true.

PR #261, opened 2020-06-12
handle empty names and display email

attwad:
Some users (ingestion lambda API users for example) have no public name, show emails in the users list to help identify all users.

image

### DIFF
--- a/verification/curator-service/ui/src/components/Users.test.tsx
+++ b/verification/curator-service/ui/src/components/Users.test.tsx
@@ -1,6 +1,7 @@
-import Users from './Users';
+import { fireEvent, render, within } from '@testing-library/react';
+
 import React from 'react';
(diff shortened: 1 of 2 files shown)


verification/curator-service/ui/src/components/Users.test.tsx
```diff
@@ -1,6 +1,7 @@
-import Users from './Users';
+import { fireEvent, render, within } from '@testing-library/react';
```
allysonjp715, line 1: Out of curiosity, do you manually reorder these imports or does it happen on save? I don't have any automatic reordering of imports configured.

attwad: I do have the sort-imports vs code extension installed yes and recommend it.
https://marketplace.visualstudio.com/items?itemName=amatiasq.sort-imports

PR #262, opened 2020-06-12
Add geo admin areas to linelist table

attwad:
Also remove column sorting ability as it is not supported by the backend yet.

### DIFF
--- a/verification/curator-service/ui/src/components/LinelistTable.tsx
+++ b/verification/curator-service/ui/src/components/LinelistTable.tsx
@@ -297,12 +297,10 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
                                                 country: c.location.country,
                                                 adminArea1:
                                                     c.location
-                                                        ?.administrativeAreaLevel1 ||
-                                                    '',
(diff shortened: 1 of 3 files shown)


attwad:
[Implement geocoding of new cases based on aggreed-upon geocoder backend](https://trello.com/c/K6SeR1oF/218-implement-geocoding-of-new-cases-based-on-aggreed-upon-geocoder-backend)

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -279,6 +295,14 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
                                                 sex: c.demographics?.sex,
                                                 age: c.demographics?.age,
                                                 country: c.location.country,
+                                                adminArea1:
+                                                    c.location
+                                                        ?.administrativeAreaLevel1 ||
```
allysonjp715, line 300: Does it need the `|| ''` ? The other fields don't do that.

attwad: on you're right we don't need that

PR #264, opened 2020-06-12
Scroll field into view when clicking on table of contents row

allysonjp715:
Screen-Recording-2020-06-15-at-1

### DIFF
--- a/verification/curator-service/ui/src/components/NewCaseForm.tsx
+++ b/verification/curator-service/ui/src/components/NewCaseForm.tsx
@@ -12,6 +12,7 @@ import FormControl from '@material-ui/core/FormControl';
 import InputLabel from '@material-ui/core/InputLabel';
 import MenuItem from '@material-ui/core/MenuItem';
 import React from 'react';
+import Scroll from 'react-scroll';
 import axios from 'axios';
 
 interface User {
@@ -89,7 +90,7 @@ class NewCaseForm extends React.Component<Props, NewCaseFormState> {
                     creationMetadata: {
                         curator: this.props.user.email,
                         date: new Date().toISOString(),
-                    }
+                    },
                 },
             });
             this.setState({ errorMessage: '' });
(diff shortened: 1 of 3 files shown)


allysonjp715:

verification/curator-service/ui/src/components/NewCaseForm.tsx
```diff
@@ -123,6 +124,14 @@ class NewCaseForm extends React.Component<Props, NewCaseFormState> {
         );
     }
 
+    scrollTo(name: string): void {
+        Scroll.scroller.scrollTo(name, {
+            duration: 100,
+            smooth: true,
+            offset: -64, // Account for header height
```
axmb, line 131: Is this available as a reference (without too much sweat)? So that we don't goof/miss this if the header's changed. Nw if not.

verification/curator-service/ui/src/components/NewCaseForm.tsx
```diff
@@ -139,31 +148,48 @@ class NewCaseForm extends React.Component<Props, NewCaseFormState> {
                 {({ submitForm, isSubmitting, values }): JSX.Element => (
                     <div className={classes.container}>
                         <nav className={classes.tableOfContents}>
```
axmb, line 150: Changes here LG. With the anticipated size of this form, would it be cleaner/easier to develop moving forward if we broke out both the ToC and a generalized form fieldset (/`Scroll.Element`) into separate components?

allysonjp715:
> Nice! gif would be awesome, but I know I ask for too much.

Added

verification/curator-service/ui/src/components/NewCaseForm.tsx
allysonjp715: I'm quite sure the entire header + create new form page will be reconstructed to fit the mocks. So I'll leave for now but keep in mind for the final product.

verification/curator-service/ui/src/components/NewCaseForm.tsx
allysonjp715: Yes, probably. I believe the scrollTo(name) command works across files too. I'll submit this PR as is, and work on a refactoring one separately. I also want to double check I'm understanding the refactor the same way you are so will chat when you're online :)

PR #266, opened 2020-06-12
Create resource-specific Lambda permission in the Source API

axmb:


verification/curator-service/api/src/controllers/sources.ts
```diff
@@ -159,6 +156,35 @@ export default class SourcesController {
...
+    /**
+     * Performs creation of AWS assets corresponding to the provided source,
+     * based on the content of the create operation.
+     *
+     * If an automation schedule is present, a CloudWatch scheduled rule will
+     * be created with a target of the global retrieval function. A resource-
+     * based permission will be added to the global retrieval function such
+     * that it can be invoked by the rule.
+     */
+    private async createSourceAwsProperties(
```
khmoran, line 168: is there ultimately going to be more in this fn? right now it seems specific to the schedule, so the comment/name don't seem to jive

axmb: Good point -- I'll rename. Named the analogous function before the source schema/logic really solidified, and so made it relatively general.

axmb:

PR #269, opened 2020-06-15
Add fields for age in new case form

allysonjp715:
- Adds Yup for form validation.
- Can enter age range or age, but not both
- Changes schema to accept either double or int

Screen Shot 2020-06-15 at 6 56 09 PM


allysonjp715:

PR #270, opened 2020-06-15
Configure resource-based policy for Lambdas in update operations

axmb:
Moved the `AddPermission` invocation from the sources controller to the CloudWatch Events client, which covers us for update operations now, in addition to creations. Needing the permission is a result of creating the target -- so it's cleaner to make this call here. Keeping the `AwsLambdaClient` separate, because it's still a helpful abstraction and I'm anticipating calling it directly in a future PR.


axmb:

PR #271, opened 2020-06-15
Remove resourced-based policy permissions in source API delete operations

axmb:
Based on #270 -- diff here is from the 3rd commit onwards.

### DIFF
--- a/verification/curator-service/api/src/clients/aws-events-client.ts
+++ b/verification/curator-service/api/src/clients/aws-events-client.ts
@@ -99,16 +99,14 @@ export default class AwsEventsClient {
         sourceId: string,
     ): Promise<void> => {
         try {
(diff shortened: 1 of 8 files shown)


khmoran:
> Based on #270 -- diff here is from the 3rd commit onwards.

I discovered that you can do diffbases in github by changing which branch it would merge into; once that parent PR is merged into master and the branch is deleted, it will switch to master here too

axmb:
> > Based on #270 -- diff here is from the 3rd commit onwards.
> 
> I discovered that you can do diffbases in github by changing which branch it would merge into; once that parent PR is merged into master and the branch is deleted, it will switch to master here too

Awesome, thanks.

axmb:

PR #272, opened 2020-06-15
Bring location schema in line with spec

khmoran:
Sorry as usual for the large diff :\ womp womp -- some of it is autoformatting though

Major changes:

- Remove location.id field; @attwad thinks we won't need it (and I agree)
- Add enum field `location.geoResolution`, and modify the converter to populate it (for location and travelHistory.location)
- Replace `location.locality` with `location.administrativeArea3` -- we do need this to be able to fully specify locations, and it is present in current data (it's not in the spec, though I'll see if Stephen wants to add it; even so, would rather not lose this precision even if we're not going to export this field)
- Add `location.place`: A POI, like a hospital
- Add `location.name`: A human-readable, short name for the location
- In the converter, use all the fields that are populated by the geocoder, instead of the curator input (more normalized)

`geoResolution` will be required for new data, since it can always be imputed from the other location fields, and location is required.

TESTED=Tested the full pipeline of convert data --> apply new schema --> import data, no validation errors

### DIFF
--- a/data-serving/data-service/schemas/cases.schema.json
+++ b/data-serving/data-service/schemas/cases.schema.json
@@ -82,7 +82,7 @@
               "Admin3",
               "Admin2",
               "Admin1",
-              "Admin0"
+              "Country"
             ]
           },
           "geometry": {
@@ -225,7 +225,7 @@
                     "Admin3",
                     "Admin2",
                     "Admin1",
-                    "Admin0"
+                    "Country"
                   ]
                 },
                 "geometry": {
--- a/data-serving/data-service/schemas/cases.schema.json
+++ b/data-serving/data-service/schemas/cases.schema.json
@@ -76,6 +76,9 @@
           "place": {
             "bsonType": "string"
           },
+          "name": {
+            "bsonType": "string"
+          },
           "geoResolution": {
             "enum": [
               "Point",
@@ -217,6 +220,9 @@
                   "bsonType": "string"
                 },
                 "place": {
+                  "bsonType": "string"
+                },
+                "name": {
                   "bsonType": "string"
                 },
                 "geoResolution": {
--- a/data-serving/data-service/schemas/cases.schema.json
+++ b/data-serving/data-service/schemas/cases.schema.json
@@ -61,9 +61,6 @@
           "_id": {
             "bsonType": "objectId"
           },
-          "id": {
-            "bsonType": "string"
-          },
           "country": {
             "bsonType": "string"
           },
@@ -73,8 +70,20 @@
           "administrativeAreaLevel2": {
             "bsonType": "string"
           },
-          "locality": {
+          "administrativeAreaLevel3": {
             "bsonType": "string"
+          },
+          "place": {
+            "bsonType": "string"
+          },
+          "geoResolution": {
+            "enum": [
+              "Point",
+              "Admin3",
+              "Admin2",
+              "Admin1",
+              "Admin0"
+            ]
           },
           "geometry": {
             "bsonType": "object",
@@ -195,9 +204,6 @@
                 "_id": {
                   "bsonType": "objectId"
                 },
-                "id": {
-                  "bsonType": "string"
-                },
                 "country": {
                   "bsonType": "string"
                 },
@@ -207,8 +213,20 @@
                 "administrativeAreaLevel2": {
                   "bsonType": "string"
                 },
-                "locality": {
+                "administrativeAreaLevel3": {
                   "bsonType": "string"
+                },
+                "place": {
+                  "bsonType": "string"
+                },
+                "geoResolution": {
+                  "enum": [
+                    "Point",
+                    "Admin3",
+                    "Admin2",
+                    "Admin1",
+                    "Admin0"
+                  ]
                 },
                 "geometry": {
                   "bsonType": "object",
@@ -389,7 +407,13 @@
           "ID": {
             "bsonType": "string"
           },
+          "city": {
+            "bsonType": "string"
+          },
           "province": {
+            "bsonType": "string"
+          },
+          "country": {
             "bsonType": "string"
           },
           "geo_resolution": {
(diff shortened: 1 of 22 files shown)


khmoran:

data-serving/data-service/schemas/cases.schema.json
```diff
@@ -73,9 +70,21 @@
           "administrativeAreaLevel2": {
             "bsonType": "string"
           },
-          "locality": {
+          "administrativeAreaLevel3": {
+            "bsonType": "string"
+          },
+          "place": {
```
axmb, line 76: Making sure I grok -- e.g., "Boston Children's Hospital"?

data-serving/data-service/src/model/location.ts
```diff
@@ -1,21 +1,29 @@
 import mongoose from 'mongoose';
 
+export enum GeoResolution {
+    Point = 'Point', // Lat/long coordinates
+    Admin3 = 'Admin3', // ~City
```
axmb, line 5: nit: these comments feel a little antithetical to the purpose of the administrative area designation ("~" totally granted). Might be more useful to drop a link to [this article](https://en.wikipedia.org/wiki/List_of_administrative_divisions_by_country).

data-serving/data-service/src/model/location.ts
```diff
@@ -1,21 +1,29 @@
...
             this.administrativeAreaLevel1 == null &&
             this.administrativeAreaLevel2 == null &&
-            this.locality == null
+            this.administrativeAreaLevel3 == null &&
+            this.place == null
         );
     },
     'One of country, administrativeAreaLevel1, administrativeAreaLevel2, ' +
-        'or locality is required',
+        'administrativeAreaLevel3, or place is required',
```
axmb, line 23: Related to my prior question on the content of `place` -- is this "safe"? If `place` is a string, is it guaranteed to stand in for an administrative area or country?

data-serving/data-service/src/model/location.ts
```diff
@@ -1,21 +1,29 @@
 import mongoose from 'mongoose';
 
+export enum GeoResolution {
+    Point = 'Point', // Lat/long coordinates
+    Admin3 = 'Admin3', // ~City
+    Admin2 = 'Admin2', // ~County
+    Admin1 = 'Admin1', // ~Province
+    Admin0 = 'Admin0', // Country
```
axmb, line 8: Country/Admin0 isn't really an administrative area concept, is it? Might be clearer to call it country, so that it aligns with the schema field name.

data-serving/scripts/convert-data/constants.py
```diff
@@ -115,3 +114,8 @@
 
 ''' Valid values for sex field. '''
 VALID_SEXES = ['female', 'male', 'other']
+
+''' Value values for geographical resolution. '''
```
axmb, line 118: sp

data-serving/scripts/data-pipeline/convert_and_import_latest_data.sh
```diff
@@ -94,7 +94,7 @@ main() {
     convert_data
     setup_db
     import_data
-    cleanup
+    #cleanup
```
axmb, line 97: Intentional?

data-serving/data-service/schemas/cases.schema.json
```diff
@@ -73,9 +70,21 @@
...
             "bsonType": "string"
           },
+          "geoResolution": {
+            "enum": [
+              "Point",
+              "Admin3",
+              "Admin2",
+              "Admin1",
+              "Admin0"
```
attwad, line 85: admin0 is a country right?

data-serving/data-service/schemas/cases.schema.json
attwad: I think this is supposed to be a human readable name of the most granular feature?

data-serving/data-service/src/model/location.ts
attwad: wow, on this article we can see that France with its 6 admin areas (!) isn't hierarchical in ways that affect me negatively. /sigh

data-serving/data-service/src/model/location.ts
attwad: +1 for country, I made another comment about this

data-serving/data-service/src/model/location.ts
attwad: Maybe we should rename this to "name" and have it stand for a "human readable name of the finest grain location that this represent"?
(that would be the "text" in the mapbox api responses for example)

attwad:

data-serving/data-service/schemas/cases.schema.json
khmoran: Yes, per the spec, "Specific place location" -- so I think this would be a hospital, etc.

khmoran: Per our convo offline:

"place": specific location, e.g. hospital; cannot currently get from mapbox API, but it would map to POI; curators will be able to input it in a text field, and imported and/or parsed cases may have it
"name": human-readable name of the location, ironically to be populated by mapbox's place field

data-serving/data-service/src/model/location.ts
khmoran: removed and added link, thanks!

data-serving/data-service/src/model/location.ts
khmoran: replaced admin0 with country

data-serving/scripts/data-pipeline/convert_and_import_latest_data.sh
khmoran: nope, reverted

data-serving/data-service/schemas/cases.schema.json
khmoran: yep, changed

PR #273, opened 2020-06-16
add openapi def for suggestions of locations

attwad:


attwad:

PR #275, opened 2020-06-16
add component to display geolocation information

attwad:
Not plugging it in yet in the new case form until we can query geolocation info from the curator service.

With content from mocks:
image

Without location:
image


attwad:

PR #276, opened 2020-06-16
Add ethnicity to NewFormCase and Linelist table and cleanup sex and age

allysonjp715:


allysonjp715:

PR #278, opened 2020-06-16
add files that will be used for async fetch of suggestions in new case form

attwad:
This is based on the sheets in "Line list data model: Standard specification" spreadsheet from Stephen.

I didn't add the professions because they are way too noisy and need cleaning before we check it in.

They will be fetched using https://material-ui.com/components/autocomplete/#asynchronous-requests

### DIFF
--- a/suggest/README.md
+++ b/suggest/README.md
@@ -1,0 +1,3 @@
+# Autocomplete suggestions
+
+This directory contains files that serve as autocomplete suggestions for various input forms.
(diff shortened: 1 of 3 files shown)


attwad:

attwad:
Yes the top-level approach was to share it but I am not sure yet if data service will have a use for it... we'll see.

khmoran:
> Yes the top-level approach was to share it but I am not sure yet if data service will have a use for it... we'll see.

I think ideally we'd validate against these on the data server as well in case it has other clients

PR #279, opened 2020-06-16
Add all event dates to UI

allysonjp715:
Some of these will need to be hidden behind another field, but adding them all for now. I also removed a test line in NewCaseForm.test.tsx since I don't think it adds much value. Let me know if you disagree.

### DIFF
--- a/verification/curator-service/ui/src/components/NewCaseForm.tsx
+++ b/verification/curator-service/ui/src/components/NewCaseForm.tsx
@@ -128,6 +128,8 @@ class NewCaseForm extends React.Component<Props, NewCaseFormState> {
                 },
                 location: {
                     country: values.country,
+                    // TODO: Infer the geo resolution from the location.
+                    geoResolution: 'Country',
                 },
                 events: [
                     {
--- a/verification/curator-service/ui/src/components/NewCaseForm.tsx
+++ b/verification/curator-service/ui/src/components/NewCaseForm.tsx
@@ -60,6 +60,12 @@ interface FormValues {
     ethnicity?: string;
     country: string;
     confirmedDate: string | null;
+    onsetSymptomsDate: string | null;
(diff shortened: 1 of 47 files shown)


verification/curator-service/ui/src/components/new-case-form-fields/Events.tsx
```diff
@@ -4,17 +4,99 @@ import { KeyboardDatePicker } from 'formik-material-ui-pickers';
...
+                                className={classes.field}
+                                name="confirmedDate"
+                                label="Confirmed case date"
+                                format="yyyy/MM/dd"
+                                maxDate={new Date()}
+                                minDate={new Date('2019/12/01')}
+                                component={KeyboardDatePicker}
+                            />
+                        </div>
+                        <div className={classes.fieldRow}>
```
khmoran, line 41: worth having some a util to generate these fields, since they're very similar? i'm OK with keeping as is too

verification/curator-service/ui/src/components/NewCaseForm.tsx
```diff
@@ -123,12 +129,57 @@ class NewCaseForm extends React.Component<Props, NewCaseFormState> {
...
+                events: [
+                    {
+                        name: 'confirmed',
+                        dateRange: {
+                            start: values.confirmedDate,
+                            end: values.confirmedDate,
+                        },
                     },
-                },
+                    {
```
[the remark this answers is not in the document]
allysonjp715: sg thanks! Saw your PR - will take a look there.

verification/curator-service/ui/src/components/new-case-form-fields/Events.tsx
allysonjp715: Yea good call, done.

PR #281, opened 2020-06-17
Bring events schema up to spec

khmoran:
Adding a "value" field to the event schema to capture the content of the event, as well as the date; 

The values we expect include:

- Method of confirmation | String | Multiple choice | - PCR test- Serological test- Clinical diagnosis- 
- Hospital admission | String | Multiple choice | - Yes- No- Unknown
- Admitted to isolation unit in hospital | String | Multiple choice | - Yes- No- Unknown
- Outcome | String | Multiple choice | - Death- Recovered- Unknown

Not going to specify these values in the schema to keep it more open; I think it's probably OK for the UI to be the arbiter here.

### DIFF
--- a/data-serving/data-service/schemas/cases.schema.json
+++ b/data-serving/data-service/schemas/cases.schema.json
@@ -126,6 +126,9 @@
             "name": {
               "bsonType": "string"
             },
+            "value": {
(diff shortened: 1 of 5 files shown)


khmoran:

PR #283, opened 2020-06-17
Update enum values for travelHistory.purpose per spec

khmoran:
Still TBD whether we're putting enums in the schema so will put it for now but happy to pare down the schema later

### DIFF
--- a/data-serving/data-service/schemas/cases.schema.json
+++ b/data-serving/data-service/schemas/cases.schema.json
@@ -275,7 +275,12 @@
               }
             },
             "purpose": {
-              "bsonType": "string"
+              "enum": [
+                "Business",
+                "Leisure",
+                "Family",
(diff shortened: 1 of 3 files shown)


khmoran:

data-serving/data-service/schemas/cases.schema.json
```diff
@@ -275,7 +275,12 @@
               }
             },
             "purpose": {
-              "bsonType": "string"
+              "enum": [
+                "Business",
+                "Leisure",
+                "Family",
```
axmb, line 281: What's the distinction between family/leisure?

khmoran: I believe leisure = vacation, and family is visiting family. I believe some people do not consider the latter to be leisure :D

PR #284, opened 2020-06-17
Rename chronicDisease field to preexistingConditions per spec

khmoran:


khmoran:

PR #287, opened 2020-06-17
Geocode new cases

attwad:
If a new case doesn't have lat/lng it will trigger geocoding based on the request's `location.query` string.

I replaced `country` in the new form with this `query` field. It will be an autocomplete field once we have it working (ETA 4-5 years based on early assessment).

In order to be able to run the dev stack connected to map box and the integration tests as well I made the cases controller accept an array of geocoders that are called in sequence. This way we can fake locations in integration tests and play with real geocodes on our local stack.

We also should probably retire the edit/new functionality of the linelist table soon, fixing the tests is not super fun now that the new form and the table have drifted a bit.

### DIFF
--- a/verification/curator-service/api/src/controllers/cases.ts
+++ b/verification/curator-service/api/src/controllers/cases.ts
@@ -10,7 +10,7 @@ import axios from 'axios';
 export default class CasesController {
     constructor(
         private readonly dataServerURL: string,
-        private readonly geocoder: Geocoder,
+        private readonly geocoders: Geocoder[],
     ) {}
 
     list = async (req: Request, res: Response): Promise<void> => {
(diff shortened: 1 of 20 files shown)


attwad:

verification/curator-service/api/src/controllers/cases.ts
```diff
@@ -63,7 +63,27 @@ export default class CasesController {
     };
 
     create = async (req: Request, res: Response): Promise<void> => {
-        // TODO: Geocode if no lat/lng provided.
+        // Geocode query if no lat lng were provided.
```
allysonjp715, line 66: What if there is a lat/lng but no country, admin1, etc? Would we want to geocode to get those values from the lat/lng?

attwad: There's no reverse geocoding yet, I could add that but the current geocoder only does forward geocoding.

PR #290, opened 2020-06-17
Couple small adjustments based on clarifications to spec

khmoran:
Left a few comments for Stephen overnight and adjusting the spec based on responses>

- Pathogen.id should be an int
- Travel purpose enum should have 'Unknown' option

Also fixed a bug that would be caught by a test if I had one <shame bell>


khmoran:

PR #292, opened 2020-06-17
Add nationalities autocomplete input

allysonjp715:
Requires making nationalities plural in the schema


verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -75,6 +76,8 @@ interface TableRow {
     sex: string;
     age: number;
     ethnicity: string;
+    // Represents a list in a string e.g. Afghan, Albanian
```
attwad, line 79: comment that this is comma+space separated please

verification/curator-service/ui/src/components/NewCaseForm.tsx
```diff
@@ -377,7 +383,9 @@ class NewCaseForm extends React.Component<Props, NewCaseFormState> {
                         </nav>
                         <div className={classes.form}>
                             <Form>
-                                <Demographics></Demographics>
+                                <Demographics
+                                    setFieldValue={setFieldValue}
```
attwad, line 387: so you didn't need `setTouched` as well to do: `onBlur={ () => setTouched({ [name]: true }) }` ?

verification/curator-service/ui/src/components/new-case-form-fields/Demographics.tsx
```diff
@@ -30,7 +32,13 @@ const styles = () =>
         },
     });
 
-type Props = WithStyles<typeof styles>;
+interface DemographicsProps extends WithStyles<typeof styles> {
+    setFieldValue: (
```
attwad, line 36: just to check but formik doens't export a type for that?

verification/curator-service/ui/src/components/NewCaseForm.tsx
allysonjp715: I suppose we could, but we don't use the value of touched anywhere so I'm not sure if its worth adding.

attwad: ok fine, the docs seem to imply this will help validating the field (otherwise perhaps only the initial value is checked), and this is done for all other fields automatically because we're using <Field> which sets onBlur and onChange automatically.
https://jaredpalmer.com/formik/docs/api/formik#settouched-fields--field-string-boolean--shouldvalidate-boolean--void

But if things work as is let's keep it simple then :)

verification/curator-service/ui/src/components/LinelistTable.tsx

verification/curator-service/ui/src/components/NewCaseForm.tsx
allysonjp715: Sorry I deleted my comment after I realized it was wrong but before I saw your second comment :P

I added it in for just the reason you said - if there's an error on the field it shouldn't be shown until touched is set to true. Added it in.

verification/curator-service/ui/src/components/new-case-form-fields/Demographics.tsx
allysonjp715: Not specifically, but I found a much cleaner way to get access to it through using useFormikContext(). Thanks for checking!

allysonjp715:

PR #294, opened 2020-06-18
proper display of lat lng in the linelist table and remove add functionality

attwad:
Also adds admin3 and proper geo resolution fields.


attwad:

PR #296, opened 2020-06-18
Move Location form to its own component

attwad:
In preparation to adding autosuggest there.


attwad:

PR #297, opened 2020-06-18
Add profession to NewCaseForm and LinelistTable

allysonjp715:
Pulled the autocomplete component into it's own file so it can be used in multiple places.

### DIFF
--- a/verification/curator-service/ui/src/components/NewCaseForm.tsx
+++ b/verification/curator-service/ui/src/components/NewCaseForm.tsx
@@ -59,6 +59,7 @@ interface FormValues {
     age?: number;
     ethnicity?: string;
     nationalities: string[];
+    profession: string | null;
     locationQuery: string;
     confirmedDate: string | null;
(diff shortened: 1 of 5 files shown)


allysonjp715:

verification/curator-service/ui/src/components/NewCaseForm.tsx
```diff
@@ -59,6 +59,7 @@ interface FormValues {
     age?: number;
     ethnicity?: string;
```
attwad, line 60: sometimes we have this ?: string; and sometimes string | null. Can we be consistent please?

allysonjp715: Yea I agree it's not great. I chose either undefined or null for each based on what the defaults are in the field selector when you remove the field and what the backend accepted for "not-populated". I'll look again now and see if those can be aligned and do it as it's own PR if so.

PR #299, opened 2020-06-19
Add CaseReference object, per spec, to our case schema/model

axmb:
I'll chat with Kelly next week, but for the time being, this adds the `CaseReference` object defined in the official line list spec to our data service schema/model. This is necessary to perform upsert for the purpose of deduplication, which I'll send subsequently.

I suspect we'll want to somewhat merge this with the `Sources` field; something for me to pick Kelly/Stephen's brain on next week.

### DIFF
--- a/data-serving/data-service/schemas/cases.schema.json
+++ b/data-serving/data-service/schemas/cases.schema.json
@@ -16,10 +16,10 @@
           "_id": {
             "bsonType": "objectId"
           },
-          "dataSourceId": {
+          "sourceId": {
             "bsonType": "string"
           },
-          "dataEntryId": {
+          "sourceEntryId": {
             "bsonType": "string"
           }
         }
--- a/data-serving/data-service/schemas/cases.schema.json
+++ b/data-serving/data-service/schemas/cases.schema.json
@@ -9,6 +9,21 @@
       "__v": {
         "bsonType": "int"
       },
+      "caseReference": {
+        "bsonType": "object",
+        "additionalProperties": false,
+        "properties": {
+          "_id": {
+            "bsonType": "objectId"
+          },
+          "dataSourceId": {
+            "bsonType": "string"
+          },
+          "dataEntryId": {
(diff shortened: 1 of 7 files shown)


axmb:

data-serving/data-service/schemas/cases.schema.json
```diff
@@ -9,6 +9,21 @@
...
         "bsonType": "int"
       },
+      "caseReference": {
+        "bsonType": "object",
+        "additionalProperties": false,
+        "properties": {
+          "_id": {
+            "bsonType": "objectId"
+          },
+          "dataSourceId": {
```
allysonjp715, line 19: Will this be a foreign key to the _id from the sources collection? If so, do we need to keep the "sources" field in this schema? (Not suggesting deletion in this PR, should be it's own if necessary)

data-serving/data-service/schemas/cases.schema.json
```diff
@@ -9,6 +9,21 @@
...
+            "bsonType": "string"
+          },
+          "dataEntryId": {
```
allysonjp715, line 22: nit: I'm not sure if dataSourceId and dataEntryId are the most clear names. I looked at this PR before I read your deduping doc and was confused by what they represented. Maybe "sourceId" and "sourceEntryId"? Especially for dataEntryId I didn't understand that it was the id that came from the source.

data-serving/data-service/schemas/cases.schema.json
axmb: Yep, certainly a good question. This is related to my comment in the PR description re: `Sources`. Let's catch up with Kelly on this next week -- I'm not sure what her plans were for `Sources` now that the spec is finalized.

data-serving/data-service/schemas/cases.schema.json
axmb: Makes sense to me -- done.

PR #300, opened 2020-06-19
Add the case upsert API to the data service

axmb:
I'd originally envisioned implementing this with `Mongoose.Model.findOneAndUpdate([...], {upsert=true})`, but it turns out that behavior's pretty thorny. I've used some straight-forward base calls instead, which seems to be the simplest solution here.

One thing I'm lukewarm on is the fact that we're allowing partial updates here. But, best to leave that consistent with the PUT update API until we introduce a PATCH endpoint.

### DIFF
--- a/data-serving/data-service/src/controllers/case.ts
+++ b/data-serving/data-service/src/controllers/case.ts
@@ -124,7 +124,13 @@ export const update = async (req: Request, res: Response): Promise<void> => {
 };
 
 /**
- * Upserts a case based on a compound index of { sourceId, entryId }.
+ * Upserts a case based on a compound index of
+ * caseReference.{dataSourceId, dataEntryId}.
+ *
+ * On success, the returned status code indicates whether than item was created
+ * (201) or updated (200).
+ *
+ * Handles HTTP PUT /api/cases.
  */
 export const upsert = async (req: Request, res: Response): Promise<void> => {
     try {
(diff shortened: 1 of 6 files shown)


axmb:

data-serving/data-service/src/controllers/case.ts
```diff
@@ -123,6 +123,42 @@ export const update = async (req: Request, res: Response): Promise<void> => {
...
+ * caseReference.{dataSourceId, dataEntryId}.
+ *
+ * On success, the returned status code indicates whether than item was created
+ * (201) or updated (200).
+ *
+ * Handles HTTP PUT /api/cases.
+ */
+export const upsert = async (req: Request, res: Response): Promise<void> => {
+    try {
+        const c = await Case.findOne({
```
attwad, line 137: you'll want a transaction to do the fetch + save here now?

https://docs.mongodb.com/manual/core/transactions/#transactions-api

axmb:
> This PR also contains the changes from #301 , bad git merge?

I (diff)based that one on this branch, so I think this is expected. If you merge this one first, GitHub automatically rebases the "child" feature onto master.

image

data-serving/data-service/src/controllers/case.ts
axmb: Totally, this isn't atomic. A transaction would be better, but for now we're at least protected from doing anything "unsafe" by [Mongoose versioning](https://mongoosejs.com/docs/guide.html#versionKey). It would be better to do this in a transaction, but given the expected QPS (to a given document), I figured it was OK to revisit.

PR #301, opened 2020-06-19
Add a proxy upsert API to the curator service

axmb:
Only non-triviality here is geocoding. I've just extracted the biz logic to be shared by create and upsert mechanisms.

### DIFF
--- a/verification/curator-service/api/src/controllers/cases.ts
+++ b/verification/curator-service/api/src/controllers/cases.ts
@@ -62,27 +62,31 @@ export default class CasesController {
         }
     };
 
+    upsert = async (req: Request, res: Response): Promise<void> => {
+        if (!(await this.geocode(req))) {
+            res.status(404).send(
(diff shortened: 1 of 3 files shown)


axmb:

verification/curator-service/api/src/controllers/cases.ts
```diff
@@ -62,27 +62,31 @@ export default class CasesController {
...
+                `no geolocation found for ${req.body['location']?.query}`,
+            );
+            return;
+        }
+        try {
+            const response = await axios.put(
+                this.dataServerURL + '/api' + req.url,
+                req.body,
+            );
+            res.json(response.data);
```
attwad, line 77: Not necessary now but it seems that returning 200 when updated and 201 when created is standard procedure. To keep in mind I guess.

axmb: Yep -- wasn't rocking the boat here. We should fix that at some point, though.

PR #307, opened 2020-06-22
Better looking place suggest

attwad:
Use the `place_name` field instead of the `text` which is more suitable for suggest and overall.
Add an icon next to the suggestions.

New suggest UI:
Screen Shot 2020-06-22 at 12 51 31 PM


attwad:

PR #309, opened 2020-06-22
Remove ID column and display age properly

attwad:
A bit more in line with the figma mocks now. Chose the interval display for age `[,...)` instead of  `xx-yy` not sure what's best yet.

image

Also disable reordering of columns and put actions at the end as per the mocks (we'll want to switch to a menu instead perhaps)

### DIFF
--- a/verification/curator-service/ui/src/components/LinelistTable.tsx
+++ b/verification/curator-service/ui/src/components/LinelistTable.tsx
@@ -83,7 +83,7 @@ interface TableRow {
     id: string;
     // demographics
     sex: string;
-    age: number;
+    age: [number, number]; // start, end.
     ethnicity: string;
     // Represents a list as a comma and space separated string e.g. 'Afghan, Albanian'
     nationalities: string;
@@ -141,7 +141,8 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
             demographics: {
                 sex: rowData.sex,
                 age: {
-                    start: rowData.age,
+                    start: rowData.age[0],
+                    end: rowData.age[1],
                 },
                 ethnicity: rowData.ethnicity,
                 nationalities: rowData.nationalities?.split(', '),
@@ -245,12 +246,6 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
                     <MaterialTable
                         columns={[
                             {
-                                title: 'ID',
-                                field: 'id',
-                                filtering: false,
-                                editable: 'never',
-                            },
-                            {
                                 title: 'Sex',
                                 field: 'sex',
                                 filtering: false,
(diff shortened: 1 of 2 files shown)


verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -244,12 +248,6 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
                 <Paper>
                     <MaterialTable
                         columns={[
-                            {
```
allysonjp715, line 247: IIRC the last discussion we had was to keep the ID field in to give curators a way to name a case when discussing e.g. duplicates.

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -260,7 +258,11 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
                                 title: 'Age',
                                 field: 'age',
                                 filtering: false,
-                                type: 'numeric',
+                                editable: 'never',
+                                render: (rowData) =>
+                                    rowData.age[0] === rowData.age[1]
+                                        ? rowData.age[0]
+                                        : `[${rowData.age[0]},${rowData.age[1]})`,
```
allysonjp715, line 265: The mocks have "0-1" and is more readable than this syntax IMO.

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -140,9 +140,13 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
         return {
             demographics: {
                 sex: rowData.sex,
-                age: {
-                    start: rowData.age,
-                },
+                ageRange:
+                    rowData.age[0] !== null || rowData.age[1] !== null
```
allysonjp715, line 144: Is this check needed? Does the following work?
```ts
ageRange: {
    start: rowData.age[0] ?? undefined;
    end: rowDate.age[1] ?? undefined;
}
```

verification/curator-service/ui/src/components/LinelistTable.tsx
attwad: we could show it in the details page (or in the details pane feature of this table that we're not usng yet) but here it takes some big real estate for no good reason imo, let's discuss with Shru.

verification/curator-service/ui/src/components/LinelistTable.tsx
attwad: ok changed to xx-yy

verification/curator-service/ui/src/components/LinelistTable.tsx
attwad: ah yeah I forgot about the ?? :) done.

attwad:

PR #310, opened 2020-06-22
Only show hospital admission date if the case has been admitted to th…

allysonjp715:
…e hospital

### DIFF
--- a/verification/curator-service/ui/src/components/new-case-form-fields/Events.tsx
+++ b/verification/curator-service/ui/src/components/new-case-form-fields/Events.tsx
@@ -1,5 +1,6 @@
+import { Field, useFormikContext } from 'formik';
+
 import DateFnsUtils from '@date-io/date-fns';
-import { Field } from 'formik';
 import FormControl from '@material-ui/core/FormControl';
 import InputLabel from '@material-ui/core/InputLabel';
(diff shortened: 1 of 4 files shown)


allysonjp715:

verification/curator-service/ui/src/components/new-case-form-fields/Events.tsx
```diff
@@ -36,90 +85,60 @@ const methodsOfConfirmation = [
...
+                    <DateField
+                        name="selfIsolationDate"
+                        label="Self isolation date"
+                    ></DateField>
+                    <SelectField
+                        name="admittedToHospital"
+                        label="Hospital admission"
+                        values={yesNoUndefined}
+                    ></SelectField>
+                    {(values as any).admittedToHospital === 'Yes' && (
```
attwad, line 121: maybe leave a todo that we should export our values interface defined in NewCaseForm to be able to use it here and in other places?
useFormikContext takes a type argument to support that: `useFormikContext<SomeValuesInterface>()`

allysonjp715: Went ahead and did it in this PR by moving the form values interface into its own file.

PR #314, opened 2020-06-22
Display only location name and country in linelist table instead of various geo fields

attwad:
Conforms to mocks.

image


attwad:

PR #315, opened 2020-06-22
Fix example case.json

khmoran:
- Fix a bug in the example case from a schema change last week
- While I was in there, did a quick cleanup (removing geo_resolution from importedCase)

### DIFF
--- a/data-serving/samples/cases.json
+++ b/data-serving/samples/cases.json
@@ -7,7 +7,9 @@
             },
             "sex": "Female",
             "profession": "Horse trainer",
-            "nationality": "Swedish",
+            "nationalities": [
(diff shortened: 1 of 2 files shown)


data-serving/samples/cases.json
```diff
@@ -2,16 +2,14 @@
     {
         "demographics": {
             "ageRange": {
-                "start": {
-                    "$numberDouble": "50"
```
axmb, line 6: What's the impact of removing this?

khmoran: it used to be that start/end had to be doubles, but @allysonjp715 changed that last week and now it can be an int or double

PR #316, opened 2020-06-22
Finish bringing travelHistory field up to spec

khmoran:
I recommend going through by commits, but the two major changes are:

- Add "method" field to travel history
- Remove 'additionalInformation' field from travel history
- There are two new properties that apply to the whole travel history field, so we need to introduce a new schema that has these properties + the current array of Travel documents

TESTED=Import of sample case works; full conversion + import of csv data works


khmoran:

data-serving/data-service/schemas/cases.schema.json
```diff
@@ -213,110 +213,125 @@
...
-        "items": {
-          "bsonType": "object",
-          "additionalProperties": false,
-          "properties": {
-            "_id": {
-              "bsonType": "objectId"
-            },
-            "location": {
+        "bsonType": "object",
+        "additionalProperties": false,
```
axmb, line 217: Is this true anywhere? If not, can we inherit from the top-level schema?

data-serving/data-service/schemas/cases.schema.json
```diff
@@ -213,110 +213,125 @@
...
+        "properties": {
+          "_id": {
```
axmb, line 219: Slightly orthogonal thought, but WDYT about `_id` in all of our subdocs? Does it serve any purpose if we don't plan to query the documents by ID?

data-serving/data-service/schemas/cases.schema.json
```diff
@@ -213,110 +213,125 @@
...
-                "Unknown"
-              ]
             }
+          },
+          "traveledPrior30Days": {
+            "bsonType": "bool"
+          },
+          "numLocations": {
+            "bsonType": "int"
           }
```
axmb, line 335: Are these two fields derivable from the travel field? We should be cautious, IMO, about adding "convenience" fields to the core dataset.

data-serving/data-service/src/model/travel-history.ts
```diff
@@ -0,0 +1,16 @@
+import { TravelDocument, travelSchema } from './travel';
+
+import mongoose from 'mongoose';
+import { positiveIntFieldInfo } from './positive-int';
+
+export const travelHistorySchema = new mongoose.Schema({
+    travel: [travelSchema],
+    traveledPrior30Days: Boolean,
+    numLocations: positiveIntFieldInfo,
```
axmb, line 9: hyper-nit, but these seem alphabetized most elsewhere

data-serving/data-service/schemas/cases.schema.json
khmoran: I agree. I have an open comment on the spec about this -- I think the idea might be that some datasets only have values for the "convenience" fields, and we don't want to lose that data -- but I'm not sure. I will definitely come back and clean it up if we change the spec.

data-serving/data-service/schemas/cases.schema.json
khmoran: MongoDB adds this by default :\ If we want to use additionalProperties: false, we have to include it in the schema (AFAICT).

data-serving/data-service/schemas/cases.schema.json
khmoran: unfortunately, it doesn't seem to inherit :\ I tested removing it from a sub-schema and adding a random prop to that schema and it validates. I can't find any documentation on it though :(

data-serving/data-service/src/model/travel-history.ts
khmoran: done

data-serving/data-service/schemas/cases.schema.json
axmb: There's [this option in Mongoose](https://mongoosejs.com/docs/guide.html#_id) to disable it. I'm not totally sure what it means by it "will refuse to save a document," but I can't imagine they'd have this for sub-docs, then prevent that parent doc from being saved. :thinking:

khmoran: many dates late but here it is! https://github.com/open-covid-data/healthmap-gdo-temp/pull/411

PR #317, opened 2020-06-22
Add transmission to schema per spec

khmoran:
I'm still waiting on clarity on whether route & place should be lists, but will send a new PR if I need to change it once I have answers.


khmoran:

data-serving/data-service/schemas/cases.schema.json
```diff
@@ -355,6 +355,33 @@
...
+            "Iatrogenic",
+            "Vector borne",
+            "Other",
+            "Unknown"
+          ]
+        },
+        "place": {
+          "bsonType": "string"
+        },
+        "linkedCases": {
```
axmb, line 377: Will this be the IDs of the cases? If so, could we clarify this in the field name (linkedCaseIds)?

PR #318, opened 2020-06-22
Add genomeSequences field to schema per spec

khmoran:
The GenomeSequences field is an array of GenomeSequence objs, which have sample collection date, repository url, sequence id, sequence name, sequence length, and notes fields.

AFACIT, none of these fields is required.


khmoran:

PR #319, opened 2020-06-23
Update retrieval/India parsing functions: (1) based on case schema changes; (2) to use the upsert API

axmb:
Couple parsing updates based on the data spec, and on what we're now showing in the linelist UI. Not extracting every possible field to fit the schema, since I imagine a couple things might still change, and this isn't in the first wave of data we're focused on productionizing.


axmb:
[Update the India parser to the ~finalized line list schema](https://trello.com/c/dOJQbFJC/205-update-the-india-parser-to-the-finalized-line-list-schema)

PR #320, opened 2020-06-23
Add method of confirmation to linelist

attwad:


attwad:

verification/curator-service/ui/cypress/integration/components/LinelistTableTest.spec.ts
```diff
@@ -9,6 +9,15 @@ describe('Linelist table', function () {
         cy.clearSeededLocations();
     });
 
+    it('Display case properly', function () {
+        cy.addCase('France', 'some notes', 'www.example.com');
+        cy.visit('/cases');
+        cy.contains('PCR test');
```
allysonjp715, line 15: It's not the best to test for something that wasn't explicitly set in the setup. WDYT about adding this to the addCase parameters? We could change the params to opts so people only have to set what they are explicitly testing for.

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -389,9 +397,13 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
                                                 geoResolution:
                                                     c.location?.geoResolution,
                                                 locationName: c.location?.name,
-                                                confirmedDate: confirmedDate
-                                                    ? new Date(confirmedDate)
+                                                confirmedDate: confirmedEvent
```
allysonjp715, line 400: confirmedEvent?.dateRange?.start ? new Date...

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -389,9 +397,13 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
...
+                                                    ? new Date(
+                                                          confirmedEvent.dateRange.start,
+                                                      )
                                                     : null,
+                                                confirmationMethod:
+                                                    confirmedEvent?.value || '',
```
allysonjp715, line 406: Is the `|| ''` needed?

verification/curator-service/ui/src/components/LinelistTable.tsx
attwad: done

verification/curator-service/ui/src/components/LinelistTable.tsx
attwad: ah no sorry that was before I used confirmedDate?: string as a type

verification/curator-service/ui/cypress/integration/components/LinelistTableTest.spec.ts
attwad: good call, done.

PR #321, opened 2020-06-23
Add left side collapsible drawer for navigation

attwad:
This brings us closer to the mocks and is in general nice to have imo.
Went for collapsible vs permanent as real estate is limited on my macbook when viewing the linelist.

Screen Recording 2020-06-23 at 11 37 03 AM mov

Changed the theme to be primary white as per mocks and secondary blue as per latest brand deck (we can fine tune those once we get real values).

I haven't moved all the navigation from <Home> to the new drawer as it was quite big of a PR already, I will do that next. One change is that charts are accessible as is, there's no need to log in to view them.

The Navbar component is merged into the main App one as it was awkward to handle open/close of the drawer from another component and it made easy to make the style of the main content consistent (has to be pushed right when drawer is open).

I moved the `User` into its oen file to share its definition between components.

Heavily inspired from https://material-ui.com/components/drawers/#persistent-drawer.


attwad:
ok thanks, yeah I followed the material example, I don't see any easy to match gmail's behavior with the given API but we can see to that later.

PR #322, opened 2020-06-23
Add transmission fields to NewCaseForm

allysonjp715:


allysonjp715:

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -179,6 +190,11 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
             symptoms: {
                 provided: rowData.symptoms?.split(', '),
             },
+            transmission: {
+                route: rowData.transmissionRoute,
+                place: rowData.transmissionPlace,
+                linkedCaseIds: rowData.transmissionLinkedCaseIds?.split(', '),
```
attwad, line 196: if you allow editing of these here then it is very brittle, we should split on , and then trim the remaining of the ID string otherwise someone can just remove a space and it won't be split.

verification/curator-service/ui/src/components/new-case-form-fields/Transmission.tsx
```diff
@@ -0,0 +1,128 @@
...
+                ></SelectField>
+                <SelectField
+                    name="transmissionPlace"
+                    label="Place of transmission"
+                    values={transmissionPlaces}
+                ></SelectField>
+                <ChipInput
+                    fullWidth
+                    alwaysShowPlaceholder
+                    placeholder="Contacted case IDs"
```
attwad, line 114: maybe we can validate that those are valid IDs here?
[a-z0-9]{24} is the regexp we use on the server side.

allysonjp715: Done.

verification/curator-service/ui/src/components/LinelistTable.tsx

PR #327, opened 2020-06-23
Catch geocode errors

attwad:
And just send the error message not the whole error as the http error, now it shows up properly, it didn't before.


verification/curator-service/ui/src/components/NewCaseForm.tsx
```diff
@@ -198,8 +198,9 @@ class NewCaseForm extends React.Component<Props, NewCaseFormState> {
             });
             this.setState({ errorMessage: '' });
         } catch (e) {
+            console.log(e);
```
allysonjp715, line 201: one more to remove :)

attwad: yes done ;)

PR #333, opened 2020-06-23
Bring symptoms and preexisting conditions up to spec with addition of status

khmoran:
- Adds status field (previously missed this in the spec!)
- Renames 'provided' to 'values'
- Removes 'imputed' (we can add later if we ever need it... ontology dreams are looking far away)

### DIFF
--- a/data-serving/data-service/schemas/cases.schema.json
+++ b/data-serving/data-service/schemas/cases.schema.json
@@ -204,19 +204,20 @@
           "_id": {
             "bsonType": "objectId"
           },
-          "provided": {
+          "values": {
             "bsonType": "array",
             "uniqueItems": true,
             "items": {
               "bsonType": "string"
             }
           },
-          "imputed": {
-            "bsonType": "array",
-            "uniqueItems": true,
-            "items": {
-              "bsonType": "string"
-            }
+          "status": {
+            "enum": [
+              "Asymptomatic",
+              "Presymptomatic",
+              "Symptomatic",
+              "Unknown"
+            ]
           }
         }
       },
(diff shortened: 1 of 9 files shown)


khmoran:
[Sprint #5: Updates to schema based on standard spec](https://trello.com/c/uimr21R4/207-updates-to-schema-based-on-standard-spec)

data-serving/data-service/schemas/cases.schema.json
```diff
@@ -204,19 +204,20 @@
...
-            "uniqueItems": true,
-            "items": {
-              "bsonType": "string"
-            }
+          "status": {
+            "enum": [
+              "Asymptomatic",
+              "Presymptomatic",
+              "Symptomatic",
+              "Unknown"
```
allysonjp715, line 219: Do we want Unknown as a specific value? Alternative would be to not populate this field. I think we've gone in the latter direction for other fields.

data-serving/data-service/schemas/cases.schema.json
```diff
@@ -227,19 +228,19 @@
...
-            "bsonType": "array",
-            "uniqueItems": true,
-            "items": {
-              "bsonType": "string"
-            }
+          "status": {
+            "enum": [
+              "Yes",
+              "No",
+              "Unknown"
```
allysonjp715, line 242: Same ? as above

data-serving/data-service/schemas/cases.schema.json
khmoran: I believe Stephen wants to make these fields mandatory with Unknown as the default, so there's less room for ambiguity -- there was some back and forth on the spec about this: https://docs.google.com/spreadsheets/d/1qOnGJQg1gtbmNxAW8c9hC1TvFpIQmSnORD-x-iXrpJ0/edit?disco=AAAAJn4cUhc

khmoran:
[Sprint #5: Updates to schema based on standard spec](https://trello.com/c/uimr21R4/207-sprint-5-updates-to-schema-based-on-standard-spec)

PR #336, opened 2020-06-24
Basic bulk case entry page

axmb:
Couple notes:

- Large-ish change, but commits are structured to be intelligible units (there's a clean-up commit at the end; please don't shame my typos/logging).
- The form of the "standard" CSV isn't defined yet, so just using something ballpark for now. Not testing geocoding -- let's consider exercising that in a Cypress case when more of the flow nuance is set in stone.


verification/curator-service/ui/cypress/fixtures/bad_bulk_data.csv
```diff
@@ -0,0 +1,2 @@
+sourceId,sourceEntryId,sex,ageRangeStart,ageRangeEnd,country,locationQuery,latitude,longitude,geoResolution,locationName,dateConfirmedStart,dateConfirmedEnd,url
```
allysonjp715, line 1: Could you delete the examples file now that we have real fixtures in this folder? 
https://github.com/open-covid-data/healthmap-gdo-temp/blob/master/verification/curator-service/ui/cypress/fixtures/example.json

verification/curator-service/ui/cypress/fixtures/bad_bulk_data.csv
```diff
@@ -0,0 +1,2 @@
...
+abc,def,InvalidSex,42,42,Canada,Canada,42,84,Country,Canada,06/23/2020,06/23/2020,foo.bar
```
allysonjp715, line 2: Should we have more than one line of data to test the bulk-ness of it? Same below.

verification/curator-service/ui/cypress/support/commands.ts
```diff
@@ -1,3 +1,5 @@
+import 'cypress-file-upload';
```
allysonjp715, line 1: What is this import for?

verification/curator-service/ui/src/components/BulkCaseForm.tsx
```diff
@@ -0,0 +1,144 @@
...
+                } else {
+                    this.setState({ statusMessage: e.message });
+                }
+            }
+        }
+    }
+
+    render(): JSX.Element {
+        const { classes } = this.props;
+        const papaparseOptions = {
```
allysonjp715, line 119: hmm can't tell if this is a typo or intentional :P what is the 'papa' prefix for? It sounds like a sports announcer introducing parseOptions 😁

verification/curator-service/ui/cypress/support/commands.ts
axmb: The `attachFile()` command comes from this package.

verification/curator-service/ui/src/components/BulkCaseForm.tsx
axmb: Ha, yeah -- the react-csv-reader package uses a popular file parsing package called Papa Parse. Seems pretty canonical to name the options for react-csv-reader this way.

verification/curator-service/ui/cypress/fixtures/bad_bulk_data.csv
axmb: Good idea, will add.

verification/curator-service/ui/cypress/fixtures/bad_bulk_data.csv
axmb: Will do!

PR #337, opened 2020-06-24
Add param to limit geocode to a specific resolution

attwad:
The /api/cases geocoding requests now accept a `location.limitToResolution` param which allows to filter the geocode results.
This should allow parsers that know they are about a region/district/country to limit geocodes to upper regions and not end up geocoding admin3 when they wanted admin2 for example.

Verified with a modified UI that hardcoded Admin2 as the resolution that geocoding of Puri returns the region instead of the city.

With `Admin2` restriction:
image

Without:
image

Kind of hard to ensure that properly in unit tests as the fake geocoder ignores options in order to keep it dead simple, happy to make it more complex (accept options and store and expose requests for example via another http handler) if you think it's worth it.

### DIFF
--- a/verification/curator-service/api/src/controllers/cases.ts
+++ b/verification/curator-service/api/src/controllers/cases.ts
@@ -118,36 +118,35 @@ export default class CasesController {
     private async geocode(req: Request): Promise<boolean> {
         // Geocode query if no lat lng were provided.
         const location = req.body['location'];
-        if (!location?.geometry?.latitude || !location.geometry?.longitude) {
-            for (const geocoder of this.geocoders) {
-                const opts: GeocodeOptions = {};
-                if (location['limitToResolution']) {
-                    opts.limitToResolution =
-                        Resolution[
-                            location[
-                                'limitToResolution'
-                            ] as keyof typeof Resolution
-                        ];
-                    if (!opts.limitToResolution) {
-                        throw new InvalidParamError(
-                            `invalid limitToResolution: ${location['limitToResolution']}`,
-                        );
-                    }
-                }
-                const features = await geocoder.geocode(location?.query, opts);
-                if (features.length === 0) {
-                    continue;
-                }
-                // Currently a 1:1 match between the GeocodeResult and the data service API.
-                req.body['location'] = features[0];
-                return true;
-            }
-            return false;
-        } else {
-            // Remove any left over field as the schema doesn't allow unknown fields.
-            delete location.query;
-            delete location.limitToResolution;
+        if (location?.geometry?.latitude & location.geometry?.longitude) {
+            return true;
         }
-        return true;
+        if (location?.query) {
+            throw new InvalidParamError(
+                'location.query must be specified to be able to geocode',
+            );
+        }
+        for (const geocoder of this.geocoders) {
+            const opts: GeocodeOptions = {};
+            if (location['limitToResolution']) {
+                opts.limitToResolution =
+                    Resolution[
+                        location['limitToResolution'] as keyof typeof Resolution
+                    ];
+                if (!opts.limitToResolution) {
+                    throw new InvalidParamError(
+                        `invalid limitToResolution: ${location['limitToResolution']}`,
+                    );
+                }
+            }
+            const features = await geocoder.geocode(location?.query, opts);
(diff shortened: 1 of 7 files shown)


verification/curator-service/api/src/geocoding/geocoder.ts
```diff
@@ -27,7 +27,13 @@ export enum Resolution {
     Country = 'Country',
 }
 
+export interface GeocodeOptions {
```
axmb, line 30: Do we anticipate adding more, here? On the subject of adding this interface, vs. using a single param.

verification/curator-service/api/src/controllers/cases.ts
```diff
@@ -120,8 +144,9 @@ export default class CasesController {
             }
             return false;
         } else {
-            // Remove any query as we shouldn't store it.
+            // Remove any left over field as the schema doesn't allow unknown fields.
```
axmb, line 147: For now, is this strictly necessary? Seems like something we'd added because we got an error, but with Mongoose `strict=true` (rather than `throw`), I thought we just silently dropped these.

verification/curator-service/api/src/geocoding/mapbox.ts
```diff
@@ -56,20 +67,40 @@ export default class MapboxGeocoder {
         });
     }
 
-    async geocode(query: string): Promise<GeocodeResult[]> {
+    async geocode(
+        query: string,
+        opts?: GeocodeOptions,
+    ): Promise<GeocodeResult[]> {
         query = query.trim();
```
axmb, line 74: We're calling this with `location?.query` in the other file. Won't this line throw if it's undefined?

verification/curator-service/api/src/controllers/cases.ts
```diff
@@ -110,7 +120,21 @@ export default class CasesController {
...
+                    opts.limitToResolution =
+                        Resolution[
+                            location[
+                                'limitToResolution'
+                            ] as keyof typeof Resolution
+                        ];
+                    if (!opts.limitToResolution) {
+                        throw new InvalidParamError(
+                            `invalid limitToResolution: ${location['limitToResolution']}`,
+                        );
```
axmb, line 134: DYT this logic should be here or in the geocoder?

verification/curator-service/api/src/geocoding/mapbox.ts
```diff
@@ -44,6 +46,15 @@ function getResolution(features: GeocodeFeature[]): Resolution {
     return Resolution.Country;
 }
 
+// Get the mapbox types for a given Resolution.
+const resolutionToGeocodeQueryType = new Map<Resolution, GeocodeQueryType>([
+    [Resolution.Country, 'country'],
+    [Resolution.Admin1, 'region'],
+    [Resolution.Admin2, 'district'],
+    [Resolution.Admin3, 'place'],
+    [Resolution.Point, 'poi'],
```
axmb, line 55: Did we call this (our version of mapbox's poi) place in this data service def?

verification/curator-service/api/src/geocoding/geocoder.ts
attwad: seeing how many params the mapbox GeocodeRequest has I suspect we will add more yes, thus the options struct

verification/curator-service/api/src/geocoding/mapbox.ts
attwad: good call on that bad call, I am now checking for the presence of query in the cases controller before calling this.

verification/curator-service/api/src/controllers/cases.ts
attwad: Both really, the cases controller deal with API types conversion and the mapbox geocoder deal with mapbox types conversion.

verification/curator-service/api/src/geocoding/mapbox.ts
attwad: There is no "place" resolution in the schema, it's "point".
Happy to rename if we prefer this.

axmb: Gotcha -- couldn't remember. All good.

PR #339, opened 2020-06-24
new form UI nit fixes to make it look better and more mobile friendly

attwad:
Somehow with a small screen the demographics for was fine but all the other fieldsets were using 60% of the demographics size, removing the 60% width fixed that and the form is of a reasonable size by default.

Also make notes multiline because it is really.

image


attwad:
[Sprint #5: Implement "add case" UI flow structure](https://trello.com/c/FCXJ2XgW/208-sprint-5-implement-add-case-ui-flow-structure)

PR #340, opened 2020-06-24
Add a script to safely update all deps

attwad:
This broke all unit tests somehow, the UI ones were expecting stuff that's happening faster now and as such were flaky, this is fixed by waiting on proper text to appear upon edit/deletes.

All api tests are now using an in memory mongo DB so that our local stack isn't impacted when we run unit tests.


attwad:
The *safe* bit was a lie! ... fixed it hopefully.

attwad:
@axmb FYI this is adding google ID to bearer token holders.

PR #341, opened 2020-06-24
Fix incorrect path in README

khmoran:
(my b)


attwad:
ah sorry I was running this from the scripts/ directory usually, lgtm.

khmoran:
> ah sorry I was running this from the scripts/ directory usually, lgtm.

this was totally my b

PR #342, opened 2020-06-24
Add travel history to NewCaseForm

allysonjp715:
Removes locationQuery which is no longer used.

This is the beginning of the travel history section to get the framework in place. We still need to add the other travel history fields, and add a way to delete travel history sections once added.

### DIFF
--- a/verification/curator-service/ui/src/components/NewCaseForm.tsx
+++ b/verification/curator-service/ui/src/components/NewCaseForm.tsx
@@ -117,9 +117,7 @@ class NewCaseForm extends React.Component<Props, NewCaseFormState> {
                     nationalities: values.nationalities,
                     profession: values.profession,
                 },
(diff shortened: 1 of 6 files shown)


allysonjp715:

verification/curator-service/ui/src/components/NewCaseForm.tsx
```diff
@@ -118,7 +119,6 @@ class NewCaseForm extends React.Component<Props, NewCaseFormState> {
                 },
                 location: {
                     ...values.location,
-                    ...{ query: values.locationQuery },
                 },
```
attwad, line 122: this can simply be `location: values.location` now

allysonjp715: Done

PR #349, opened 2020-06-25
Find existing sources when creating/updating/upserting case and fill them in properly

attwad:
I'm curious what you all think we should do here. This is going to cost us a bit as we do a round trip to fetch any existing source by URL before touching a case but this way we get consistent sub documents which was not the case before.

https://mongoosejs.com/docs/subdocs.html

I've tried to use mongoose pre save hooks (https://mongoosejs.com/docs/middleware.html#order) as it felt cleaner at first but there are a few issues with that approach:
- There is a circular dependency between the schema and its model when before saving a source we're trying to fetch existing sources. This is doable by using the globally accessible mongoose.schemas['Source'] but feels ugly and brittle.
- Subdocuments save hooks are called before top level documents save hooks so when a case is saved the sub sources are already saved so there's no easy way to know if the sub sources were just created or already existed before.

### DIFF
--- a/data-serving/data-service/src/controllers/case.ts
+++ b/data-serving/data-service/src/controllers/case.ts
@@ -1,6 +1,7 @@
 import { Request, Response } from 'express';
 
 import { Case } from '../model/case';
+import { Source } from '../model/source';
 
 /**
  * Get a specific case.
@@ -75,6 +76,29 @@ export const list = async (req: Request, res: Response): Promise<void> => {
     }
 };
 
+// Finds sources with an existing URL in the request body and replace them
+// for the ones that already exist.
+// TODO: Consider calling find() with all the IDs at once if multiple sources
+// per case become the norm to avoid roundtrips to mongo.
+// TODO: Support more than just URL to fetch existing sources.
+async function fillInExistingSources(req: Request): Promise<void> {
+    const sources = req.body.sources;
+    // No sources will automatically fail at validation time.
+    if (!sources) {
+        return;
+    }
+    for (let i = 0; i < sources.length; i++) {
+        if (sources[i].id) {
+            continue;
+        }
+        const source = await Source.findOne({ url: sources[i].url });
+        // If we have an existing source with the same URL, use it.
+        if (source) {
+            req.body.sources[i] = source;
+        }
+    }
+}
+
 /**
  * Create a case.
  *
@@ -82,6 +106,7 @@ export const list = async (req: Request, res: Response): Promise<void> => {
  */
 export const create = async (req: Request, res: Response): Promise<void> => {
     try {
+        await fillInExistingSources(req);
         // TODO: Don't consume req.body directly; add layer between API and
         // storage.
         const c = new Case(req.body);
@@ -104,6 +129,7 @@ export const create = async (req: Request, res: Response): Promise<void> => {
  */
 export const update = async (req: Request, res: Response): Promise<void> => {
     try {
+        await fillInExistingSources(req);
         const c = await Case.findByIdAndUpdate(req.params.id, req.body, {
             new: true,
             runValidators: true,
@@ -134,6 +160,7 @@ export const update = async (req: Request, res: Response): Promise<void> => {
  */
 export const upsert = async (req: Request, res: Response): Promise<void> => {
     try {
+        await fillInExistingSources(req);
         const c = await Case.findOne({
             'caseReference.sourceId': req.body.caseReference?.sourceId,
             'caseReference.sourceEntryId':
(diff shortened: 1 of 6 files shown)


attwad:
[Design/implement retrieval and storage of raw sources for manual entries](https://trello.com/c/k97tK7DJ/250-design-implement-retrieval-and-storage-of-raw-sources-for-manual-entries)

data-serving/data-service/src/controllers/case.ts
```diff
@@ -75,13 +76,37 @@ export const list = async (req: Request, res: Response): Promise<void> => {
...
+// TODO: Consider calling find() with all the IDs at once if multiple sources
+// per case become the norm to avoid roundtrips to mongo.
+// TODO: Support more than just URL to fetch existing sources.
+async function fillInExistingSources(req: Request): Promise<void> {
+    const sources = req.body.sources;
+    // No sources will automatically fail at validation time.
+    if (!sources) {
+        return;
+    }
+    for (let i = 0; i < sources.length; i++) {
```
axmb, line 90: nit: `for (let requestSource of sources)`?

data-serving/data-service/src/controllers/case.ts
```diff
@@ -75,13 +76,37 @@ export const list = async (req: Request, res: Response): Promise<void> => {
     }
 };
 
+// Finds sources with an existing URL in the request body and replace them
+// for the ones that already exist.
+// TODO: Consider calling find() with all the IDs at once if multiple sources
+// per case become the norm to avoid roundtrips to mongo.
+// TODO: Support more than just URL to fetch existing sources.
+async function fillInExistingSources(req: Request): Promise<void> {
```
axmb, line 84: Overall thought on the approach -- do we want to just require the source ID, and match on that?

For manual and bulk entry, we can provide a search-by-name/URL dialogue that allows source creation if an appropriate one doesn't exist. For automated ingestion, we have the source ID.

data-serving/data-service/test/controllers/case.test.ts
```diff
@@ -206,3 +208,44 @@ describe('DELETE', () => {
...
     });
 });
+
+describe('Sources', () => {
+    it('are created if needs be', async () => {
+        const c = await new Case(minimalCase).save();
+        expect(c.sources).toHaveLength(1);
+    });
+    it('are reused if existing on create', async () => {
+        const source = await new Source({ url: 'cdc.gov' }).save();
```
axmb, line 218: Might be clearer to use `minimalCase.url`?

data-serving/data-service/test/controllers/case.test.ts
```diff
@@ -206,3 +208,44 @@ describe('DELETE', () => {
...
+        console.log('created source with id', source.id);
```
axmb, line 219: Axe remaining `log`s?

data-serving/data-service/src/controllers/case.ts
khmoran: Yeah I guess I'd assumed that we'd fetch the sources in the curator portal and pass the id to the data service

data-serving/data-service/src/model/source.ts
```diff
@@ -1,22 +1,13 @@
...
+        return this != null && this.url == null && this.other == null;
     },
     'Source must specify id, url, or other',
 ];
 
 export const sourceSchema = new mongoose.Schema({
-    id: {
-        type: String,
-        required: fieldRequiredValidator,
-    },
```
khmoran, line 19: ooc, why not store the source id?

data-serving/data-service/src/controllers/case.ts
attwad: well I need to update the req.body.sources[i], if I don't get an index I can't update it right?
unless there's something like *requestSource = *source like in Go that does a deep copy of the pointed variable.

data-serving/data-service/test/controllers/case.test.ts
attwad: good point, done.

data-serving/data-service/test/controllers/case.test.ts
attwad: done.

data-serving/data-service/src/model/source.ts
attwad: _id/id is automatically added by mongoose to all documents

data-serving/data-service/src/controllers/case.ts
attwad: Right now the curator portal doesn't fetch any sources, the input for source is a textfield for a single URL, even if we implement a fetch->attach complete source from curator portal, having this check in here would allow us to make sure the cases we create actually has the proper denormalized sources. (Say, suggest field fails on curator portal somehow, a new source is going to be created if the case is submitted as is).
No strong opinion, I can see how it is cleaner to get rid of this PR, I can also see how things can go wrong without it :)

PR #350, opened 2020-06-25
Add more travel history fields

allysonjp715:
- Removes unnecessary ".clear()" calls in cypress test
- Consolidates formik fields into it's own file
- Unfortunately couldn't use a material date *range* picker since it is not stable yet: https://dev.material-ui-pickers.dev/demo/daterangepicker So I went with two date pickers

Screen Shot 2020-06-25 at 5 40 29 PM


allysonjp715:

PR #353, opened 2020-06-26
Add genome sequencing to NewCaseForm

allysonjp715:
Screen Shot 2020-06-26 at 10 11 41 AM

### DIFF
--- a/verification/curator-service/ui/src/components/new-case-form-fields/GenomeSequences.tsx
+++ b/verification/curator-service/ui/src/components/new-case-form-fields/GenomeSequences.tsx
@@ -31,65 +31,70 @@ export default function GenomeSequences(): JSX.Element {
                         return (
                             <div>
                                 {values.genomeSequences &&
-                                    values.genomeSequences.map((_, index) => (
-                                        <fieldset
-                                            key={shortId.generate()}
-                                            className={
-                                                classes.genomeSequenceSection
-                                            }
-                                        >
-                                            <DateField
-                                                name={`genomeSequences[${index}].sampleCollectionDate`}
-                                                label="Sample collection date"
-                                            ></DateField>
-                                            <Field
-                                                className={classes.field}
-                                                name={`genomeSequences[${index}].repositoryUrl`}
-                                                type="text"
-                                                label="Repository URL"
-                                                fullWidth
-                                                component={TextField}
-                                            ></Field>
-                                            <Field
-                                                className={classes.field}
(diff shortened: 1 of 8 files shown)


verification/curator-service/ui/src/components/new-case-form-fields/GenomeSequences.tsx
```diff
@@ -0,0 +1,104 @@
...
+            <fieldset>
+                <legend>Genome Sequences</legend>
+                <FieldArray name="genomeSequences">
+                    {({ push }): JSX.Element => {
+                        return (
+                            <div>
+                                {values.genomeSequences &&
+                                    values.genomeSequences.map((_, index) => (
+                                        <fieldset
+                                            key={index}
```
attwad, line 35: we shouldn't use the index as key in react lists apparently this is wrong, slows down the app and can show the wrong data :(
https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

verification/curator-service/ui/src/components/new-case-form-fields/GenomeSequences.tsx
```diff
@@ -0,0 +1,104 @@
...
+                                                fullWidth
+                                                component={TextField}
+                                            ></Field>
+                                            <Field
+                                                name={`genomeSequences[${index}].notes`}
+                                                type="text"
+                                                label="Notes"
+                                                multiline={true}
+                                                fullWidth
+                                                rows="3"
```
attwad, line 82: you can remove that for now, I know I added that to the other notes but after talking to Shruu the right behavior for textareas is to start on one line and grow as needed, I am not sure how to implement that but I agree that's a better pattern than a fixed number of row.

verification/curator-service/ui/src/components/new-case-form-fields/GenomeSequences.tsx
allysonjp715: Unfortunate that formik says to use the index in their example :/ https://jaredpalmer.com/formik/docs/api/fieldarray

Done.

verification/curator-service/ui/src/components/new-case-form-fields/GenomeSequences.tsx
allysonjp715: Done. It does grow when you hit enter.

allysonjp715:

PR #354, opened 2020-06-26
Add edit case component

attwad:
Moved the case definition that we get from the /api/cases calls to somewhere common to share it.
This caught a lot of undefined checks that needed to be done as well.

### DIFF
--- a/verification/curator-service/ui/cypress/integration/components/EditCase.spec.ts
+++ b/verification/curator-service/ui/cypress/integration/components/EditCase.spec.ts
@@ -1,0 +1,47 @@
+/* eslint-disable no-undef */
+describe('Edit case', function () {
+    beforeEach(() => {
+        cy.task('clearCasesDB', {});
+        cy.login();
+    });
+
+    afterEach(() => {
+        cy.clearSeededLocations();
+    });
+
+    it('Errors when case does not exist', function () {
+        cy.visit('/cases/edit/foo');
+        cy.contains('Request failed');
+    });
+
+    it('can edit an existing case', function () {
+        cy.addCase({
+            country: 'France',
+            notes: 'some notes',
+            sourceUrl: 'www.example.com',
+        });
+        cy.request({ method: 'GET', url: '/api/cases' }).then((resp) => {
+            console.log(resp.body);
+            expect(resp.body.cases).to.have.lengthOf(1);
+            cy.visit(`/cases/edit/${resp.body.cases[0]._id}`);
+            // Check that we have something from the original case.
+            cy.contains('France');
+            // Change a few things.
+            cy.get('div[data-testid="sex"]').click();
+            cy.get('li[data-value="Female"').click();
+            cy.get('input[name="age"]').type('21');
+            // Submit the changes.
+            cy.server();
+            cy.route('PUT', `/api/cases/${resp.body.cases[0]._id}`).as(
+                'editCase',
+            );
+            cy.get('button[data-testid="submit"]').click();
+            cy.wait('@editCase');
+            // Updated info should be there.
+            cy.visit('/cases');
+            cy.contains('No records to display').should('not.exist');
+            cy.contains('Female');
+            cy.contains('21');
+        });
+    });
+});
--- a/verification/curator-service/ui/cypress/integration/components/EditCase.spec.ts
+++ b/verification/curator-service/ui/cypress/integration/components/EditCase.spec.ts
@@ -21,11 +21,12 @@ describe('Edit case', function () {
             sourceUrl: 'www.example.com',
         });
         cy.request({ method: 'GET', url: '/api/cases' }).then((resp) => {
-            console.log(resp.body);
             expect(resp.body.cases).to.have.lengthOf(1);
             cy.visit(`/cases/edit/${resp.body.cases[0]._id}`);
             // Check that we have something from the original case.
             cy.contains('France');
+            cy.contains('Female').should('not.exist');
+            cy.contains('21').should('not.exist');
             // Change a few things.
             cy.get('div[data-testid="sex"]').click();
             cy.get('li[data-value="Female"').click();
--- a/verification/curator-service/ui/src/components/App.tsx
+++ b/verification/curator-service/ui/src/components/App.tsx
@@ -358,7 +358,7 @@ class App extends React.Component<Props, State> {
                     >
                         <div className={classes.drawerHeader} />
                         <Switch>
-                            {this.hasAnyRole(['curator', 'reader']) && (
+                            {this.hasAnyRole(['curator']) && (
                                 <Route
                                     path="/cases/edit/:id"
                                     render={({ match }) => {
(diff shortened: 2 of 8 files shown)


verification/curator-service/ui/cypress/integration/components/EditCase.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+    });
+
+    it('can edit an existing case', function () {
+        cy.addCase({
+            country: 'France',
+            notes: 'some notes',
+            sourceUrl: 'www.example.com',
+        });
+        cy.request({ method: 'GET', url: '/api/cases' }).then((resp) => {
+            console.log(resp.body);
```
allysonjp715, line 24: Remove

verification/curator-service/ui/cypress/integration/components/EditCase.spec.ts
```diff
@@ -0,0 +1,47 @@
...
+            expect(resp.body.cases).to.have.lengthOf(1);
+            cy.visit(`/cases/edit/${resp.body.cases[0]._id}`);
+            // Check that we have something from the original case.
+            cy.contains('France');
```
allysonjp715, line 28: Could also check 'Female' and '21' don't exist, to ensure the edit is changing something

verification/curator-service/ui/src/components/App.tsx
```diff
@@ -357,6 +358,19 @@ class App extends React.Component<Props, State> {
                     >
                         <div className={classes.drawerHeader} />
                         <Switch>
+                            {this.hasAnyRole(['curator', 'reader']) && (
```
allysonjp715, line 361: Should only be curator

verification/curator-service/ui/src/components/NewCaseForm.tsx
```diff
@@ -47,8 +43,97 @@ const styles = () =>
...
+                ?.start || null,
+        firstClinicalConsultationDate:
+            c.events.find((event) => event.name === 'firstClinicalConsultation')
+                ?.dateRange?.start || null,
+        selfIsolationDate:
+            c.events.find((event) => event.name === 'selfIsolation')?.dateRange
+                ?.start || null,
+        admittedToHospital: c.events.find(
+            (event) => event.name === 'hospitalAdmission',
+        )?.dateRange?.start
```
allysonjp715, line 111: This is the value param of the hospitalAdmission event, so this can be:
```ts
admittedToHospital: c.events.find(
            (event) => event.name === 'hospitalAdmission',
        )?.value
```

verification/curator-service/ui/src/components/NewCaseForm.tsx
```diff
@@ -107,104 +192,107 @@ class NewCaseForm extends React.Component<Props, NewCaseFormState> {
...
-            await axios.post('/api/cases', {
-                demographics: {
-                    sex: values.sex,
-                    ageRange: ageRange,
-                    ethnicity: values.ethnicity,
-                    nationalities: values.nationalities,
-                    profession: values.profession,
+        const newCase = {
+            demographics: {
+                sex: values.sex || '',
```
allysonjp715: why was it needed to add `|| ''` ? I would think it's better to leave it as undefined than to set a value of the empty string. Otherwise doesn't the DB value become empty string rather than not set?

verification/curator-service/ui/src/components/NewCaseForm.tsx
```diff
@@ -298,13 +360,13 @@ class NewCaseForm extends React.Component<Props, NewCaseFormState> {
                                     isChecked:
                                         values.sex !== undefined ||
                                         (values.age !== undefined &&
-                                            values.age !== '') ||
+                                            values.age.toString() !== '') ||
```
allysonjp715, line 363: When is it not a string? i.e. why is `toString()` necessary?

verification/curator-service/ui/src/components/NewCaseForm.tsx
```diff
@@ -373,7 +435,7 @@ class NewCaseForm extends React.Component<Props, NewCaseFormState> {
                                 onClick={(): void => this.scrollTo('symptoms')}
                             >
                                 {this.tableOfContentsIcon({
-                                    isChecked: values.symptoms.length > 0,
+                                    isChecked: values.symptoms?.length > 0,
```
allysonjp715, line 438: Does this statement have problems when symptoms is null? Or is there some ts craziness where `null > 0` is a valid check
If there are problems it could be `(values.symptoms?.length ?? 0) > 0`

verification/curator-service/ui/src/components/NewCaseForm.tsx
attwad: that was a mistake and I removed it already actually

verification/curator-service/ui/src/components/NewCaseForm.tsx
attwad: values.age is a number, so this check was always false. An eslint warning pointed it out.

verification/curator-service/ui/src/components/NewCaseForm.tsx
attwad: This was throwing a "can't read length property of undefined" when passing in an existing case without symptoms so I added the necessary ? safeguard. I don't expect any issue with null symptoms. Added a unit test to check that.

verification/curator-service/ui/src/components/NewCaseForm.tsx
attwad: cool, simpler.

verification/curator-service/ui/src/components/App.tsx
attwad: absolutely!

verification/curator-service/ui/cypress/integration/components/EditCase.spec.ts
attwad: good call, done.

verification/curator-service/ui/cypress/integration/components/EditCase.spec.ts
attwad: done.

verification/curator-service/ui/src/components/App.tsx
attwad: done

verification/curator-service/ui/src/components/NewCaseForm.tsx
```diff
@@ -107,104 +190,107 @@ class NewCaseForm extends React.Component<Props, NewCaseFormState> {
...
-                    sex: values.sex,
-                    ageRange: ageRange,
-                    ethnicity: values.ethnicity,
-                    nationalities: values.nationalities,
-                    profession: values.profession,
+        const newCase = {
+            demographics: {
+                sex: values.sex,
+                ageRange: ageRange,
+                ethnicity: values.ethnicity || '',
```
allysonjp715, line 197: remove `|| ''` here and for professions too?

attwad: yup, done.

attwad:
[Implement "edit case" UI flow structure](https://trello.com/c/5pAOPwxX/215-implement-edit-case-ui-flow-structure)

PR #355, opened 2020-06-26
Use common location interface

attwad:


verification/curator-service/ui/src/components/new-case-form-fields/Location.tsx
```diff
@@ -1,5 +1,6 @@
 import { Typography, withStyles } from '@material-ui/core';
 
+import { Location as CaseLocation } from '../Case';
```
allysonjp715, line 3: Maybe `Location as Loc` to keep in sync with other files

attwad: sounds good, done.

PR #356, opened 2020-06-26
Edit in full page instead of inline in linelist

attwad:
This simplifies the edit code as it is in one place instead of two and allows full edit of the case.
Also moved the errors in the linelist at the top of the page otherwise one would have to scroll further down the table to see that there was an error.

### DIFF
--- a/verification/curator-service/ui/cypress/integration/components/LinelistTableTest.spec.ts
+++ b/verification/curator-service/ui/cypress/integration/components/LinelistTableTest.spec.ts
@@ -23,7 +23,7 @@ describe('Linelist table', function () {
         cy.contains('www.example.com');
     });
 
-    it.only('Can go to the edit page', function () {
+    it('Can go to the edit page', function () {
         cy.addCase({
             country: 'France',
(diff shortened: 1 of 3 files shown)


attwad:
[Implement "edit case" UI flow structure](https://trello.com/c/5pAOPwxX/215-implement-edit-case-ui-flow-structure)

verification/curator-service/ui/cypress/integration/components/LinelistTableTest.spec.ts
```diff
@@ -23,7 +23,7 @@ describe('Linelist table', function () {
         cy.contains('www.example.com');
     });
 
-    it('Can edit a case', function () {
+    it.only('Can go to the edit page', function () {
```
allysonjp715, line 26: Oh cool does it.only run only that test? Good to know, but should be removed :)

attwad: yeah it does :) thanks for spotting! removed...

PR #358, opened 2020-06-26
Make appbar background white

attwad:
This doesn't break the other parts of the website as it did when I changed the primary color of the theme.


allysonjp715:

PR #361, opened 2020-06-26
Last-last hail mary to get the update workflow to work

khmoran:
The real magic of this PR is in the different connection string, which is a github secret. It follows the < 3.4 connection string format here: https://www.mongodb.com/blog/post/mongodb-3-6-here-to-SRV-you-with-easier-replica-set-connections

PERHAPS without the +srv URL, it won't do a DNS lookup, and magic will happen.

Note that this format of URL requires the database name in the conn string, hence the change in the data-pipeline script.

TESTED=It works locally... but so did so many other things.

### DIFF
--- a/.github/workflows/case-data-update.yml
+++ b/.github/workflows/case-data-update.yml
@@ -15,7 +15,7 @@ on:
 
 jobs:
   update-case-data-prod:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
 
     steps:
       - name: Checkout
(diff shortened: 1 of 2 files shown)


khmoran:
[Fix bug preventing automated import of data from workflow](https://trello.com/c/FNUfwVz2/238-fix-bug-preventing-automated-import-of-data-from-workflow)

khmoran:
[Fix bug preventing automated import of data from workflow](https://trello.com/c/FNUfwVz2/238-fix-bug-preventing-automated-import-of-data-from-workflow)

PR #363, opened 2020-06-29
Bulk formik upload component

axmb:
This is based to branch `bulk-formik` (#362).

For now, don't use the data from the Source component. But, once we update that to provide source ID, we'll use the information.


axmb:
[Finalize initial UI elements in bulk entry page](https://trello.com/c/d529ZOFq/245-finalize-initial-ui-elements-in-bulk-entry-page)

PR #365, opened 2020-06-29
Add text indexes to relevant fields

attwad:


attwad:
[Implement basic text search functionality](https://trello.com/c/r8dZnF6A/261-implement-basic-text-search-functionality)

data-serving/data-service/src/model/case.ts
```diff
@@ -39,12 +39,16 @@ const caseSchema = new mongoose.Schema(
...
+        notes: {
+            type: String,
+            text: true,
+        },
         pathogens: [pathogenSchema],
         preexistingConditions: dictionarySchema,
         sources: {
             type: [sourceSchema],
             required: true,
+            text: true,
```
khmoran, line 51: by defining this at this level, does it apply to all fields of the sources sub-schema?

attwad:
> is the overall approach to add text indexes to every text field?

Yes I believe so. At least that's the mongoose-way.

data-serving/data-service/src/model/case.ts
attwad: by defining this at this level, I've shown the world that I am stupid. This was a mistake.

attwad:
Well this was tricky, I thought this was good so I wrote tests, and it turns out no, mongoose will pick any of the text: true fields and make it the ONLY full text searhc index of ALL collections... which contradicts about everything the documentation says... I figured out that one needs to call index() on the schema and specify all the fields that should be in the full text search index instead. This way the tests work.
Will send the PR to implement search next so you'll be able to see.

attwad:
Alright, I moved the index creation to the setup-db script as the integration tests were resetting the DB and killing the indexes that mongoose was setting up, not worth fighting that battle imo.

data-serving/scripts/setup-db/src/index.ts
```diff
@@ -43,6 +48,14 @@ const setupDatabase = async ({
 
         const collection = await database.getCollection(collectionName);
 
+        print('Dropping indexes👇');
```
khmoran, line 51: love the emojis, definitely elevating the vibe

data-serving/scripts/setup-db/src/index.ts
```diff
@@ -43,6 +48,14 @@ const setupDatabase = async ({
...
+        const indexName = `${collectionName}Idx`;
+        await collection.dropIndex(indexName);
```
khmoran, line 53: I don't think you need to drop the indexes because we drop the collection if it exists but I guess it can't hurt

data-serving/scripts/setup-db/src/index.ts
attwad: 👍

data-serving/scripts/setup-db/src/index.ts
attwad: yeah I figured it wouldn't hurt.

PR #366, opened 2020-06-29
Add 'Unknown' display to undefined select fields

### LINKED ISSUE #347: Curator portal, manual entry form. Sex field drop down
**Describe the bug**
- Curator portal, manual entry form. Demographics, Sex field
- Sex field has small unknown selection area & untitled

**Expected behavior**
- Same size area as male / female and titled should be titled unknown

**Screenshots**
Screen Shot 2020-06-25 at 11 32 24 AM

attwad:
Fixes #347

### DIFF
--- a/verification/curator-service/ui/src/components/new-case-form-fields/Demographics.tsx
+++ b/verification/curator-service/ui/src/components/new-case-form-fields/Demographics.tsx
@@ -66,7 +66,7 @@ class Demographics extends React.Component<DemographicsProps, {}> {
                             >
                                 {sexValues.map((sex) => (
                                     <MenuItem
-                                        key={sex ?? undefined}
+                                        key={sex ?? 'undefined'}
                                         value={sex}
                                     >
                                         {sex ?? 'Unknown'}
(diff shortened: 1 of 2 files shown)


allysonjp715:
One question we need to sort out is if we should be storing 'Unknown' in the DB, or not storing a value for that field. The schema includes 'Unknown' in the value list, e.g.:
```
"sex": {
            "enum": [
              "Male",
              "Female",
              "Other",
              "Unknown"
            ]
          },
```
But I think it probably makes sense to leave the field unset (as this PR has it). Thoughts? @khmoran

attwad:
I think not storing it is a better approach than storing "unknown", the meaning is the same and it's less data to process.
I think removing the "Unknown" from the json schema is fine if @khmoran approves as well.
We don't have "other" as well in the dropdown in the UI, should we add it now?

attwad:

khmoran:
> I think not storing it is a better approach than storing "unknown", the meaning is the same and it's less data to process.
> I think removing the "Unknown" from the json schema is fine if @khmoran approves as well.
> We don't have "other" as well in the dropdown in the UI, should we add it now?

Stephen felt strongly that we should have Unknown as the default and actually specify it, to be super clear about when data is missing. I think personally I agree with you all that then we end up with two "empty" values that mean the same thing -- let's bring this up with him?

Maybe in the schema we just have it as empty, but we export it/display it as unknown?

allysonjp715:
"Maybe in the schema we just have it as empty, but we export it/display it as unknown?"

SGTM

PR #368, opened 2020-06-29
Add remove button to travel history section

allysonjp715:


allysonjp715:
> But, but, the (integration) tests? 😕 where are they?? 🧪🧪🧪

Added a test for adding and removing a travel history section

allysonjp715:

PR #370, opened 2020-06-29
Add required fields to NewCaseForm

allysonjp715:
I removed two unit tests that checked that requests were sent because those requests are no longer sent until all the required fields are filled out. That unit test didn't provide any added coverage compared to the integration tests, so I don't think they are worth keeping around. Let me know if you disagree.

In future PRs we can look into using <FastField>s to speed up the I/O on the form: https://github.com/formik/formik/issues/671

### DIFF
--- a/verification/curator-service/ui/src/components/NewCaseForm.test.tsx
+++ b/verification/curator-service/ui/src/components/NewCaseForm.test.tsx
@@ -2,10 +2,6 @@ import { fireEvent, render, wait } from '@testing-library/react';
 
 import NewCaseForm from './NewCaseForm';
 import React from 'react';
-import axios from 'axios';
-
-jest.mock('axios');
-const mockedAxios = axios as jest.Mocked<typeof axios>;
 
 const user = {
     _id: 'testUser',
@@ -13,10 +9,6 @@ const user = {
     email: 'foo@bar.com',
     roles: ['admin', 'curator'],
 };
-
-afterEach(() => {
-    jest.clearAllMocks();
-});
 
 it('renders form', () => {
     const { getByText, getAllByText } = render(<NewCaseForm user={user} />);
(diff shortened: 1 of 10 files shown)


verification/curator-service/ui/src/components/NewCaseForm.test.tsx
```diff
@@ -27,38 +19,3 @@ it('renders form', () => {
...
-    const { getByText } = render(<NewCaseForm user={user} />);
-    expect(getByText(/Submit case/i)).toBeInTheDocument();
-
-    await wait(() => {
-        fireEvent.click(getByText(/Submit case/));
-    });
-    expect(mockedAxios.post).toHaveBeenCalledTimes(1);
-});
-
-it('submits case not ok', async () => {
```
attwad, line 50: we should be testing for a POST vs PUT really in those tests depending on if we had an initial case or not, I thought I added that but no?

verification/curator-service/ui/src/components/new-case-form-fields/Events.tsx
```diff
@@ -27,11 +27,13 @@ export default function Events(): JSX.Element {
                 <DateField
                     name="confirmedDate"
                     label="Confirmed case date"
+                    required={true}
```
attwad, line 30: nit. can we just say `required` ? if no no big deal.

allysonjp715: Done

verification/curator-service/ui/src/components/NewCaseForm.test.tsx
allysonjp715: Those checks aren't in this test file in the master branch.

We do check in cypress tests if it's a POST or PUT by which route we wait on, comparing NewCaseFormTest and EditCaseTest

allysonjp715:

verification/curator-service/ui/src/components/NewCaseForm.test.tsx
attwad: ok nm then, thanks

PR #371, opened 2020-06-29
3 schema removals per various conversations

khmoran:
Each commit represents a different small schema  removal:

- 'Unknown' options from enums (decided on another PR that an empty field implies this, and we can export or display absence of data as 'Unknown' as necessary)
- travelHistory.numLocations (decided Friday during design review)
- genomeSequence.notes: Consolidating into the top-level notes field (already exists)


khmoran:

PR #373, opened 2020-06-30
Conversion: A couple more abbreviations for locations

khmoran:
Total travel history errors are down to 128! (From 2121)


khmoran:
[Converter: Re-run line list conversion and resolve remaining errors](https://trello.com/c/JZIfHyZt/230-converter-re-run-line-list-conversion-and-resolve-remaining-errors)

data-serving/scripts/convert-data/constants.py
```diff
@@ -54,18 +54,24 @@
     ('bergamo', 'lombardia', 'italy'): ('Bergamo', 'Lombardia', 'Italy'),
     ('harbin',): ('Harbin City', 'Heilongjiang', 'China'),
     ('lombardy', 'italy'): (None, 'Lombardia', 'Italy'),
-    ('lombardy italy',): (None, 'Lombardia', 'Italy'),
     ('london',): ('London', 'England', 'United Kingdom'),
     ('madrid',): (None, 'Madrid', 'Spain'),
+    ('madrid', 'spain'): (None, 'Madrid', 'Spain'),
+    ('miami', 'florida', 'united states'): ('Miami-Dade County', 'Florida', 'United States'),
```
axmb, line 60: What does this data represent? The comment says `city, province, country` -- but we're using Miama-Dade County in lieu of Miami? And, totally without context, I'd guess that `miami` is most often a reference to the city as opposed to the county.

data-serving/scripts/convert-data/constants.py
```diff
@@ -54,18 +54,24 @@
...
     ('milan', 'italy'): ('Milan', 'Lombardia', 'Italy'),
     ('nanning',): ('Nanning City', 'Guangxi', 'China'),
+    ('naples',): ('Naples', 'Campania', 'Italy'),
+    ('rio de janeiro', 'brazil'): ('Rio de Janeiro', 'Rio de Janeiro', 'Brazil'),
     ('sao paulo',): (None, 'Sao Paulo', 'Brazil'),
-    ('south tyrol',): ('Trentino-Alto Adige', 'Italy'),
-    ('south tyrol', 'italy'): ('Trentino-Alto Adige', 'Italy'),
+    ('south tyrol',): (None, 'Trentino-Alto Adige', 'Italy'),
+    ('south tyrol', 'italy'): (None, 'Trentino-Alto Adige', 'Italy'),
+    ('taiwan',): (None, 'Taiwan', None),
```
axmb, line 69: Now that I see this row, I'm guessing there's some sort of conditional flexibility on how these tuples are interpreted?

data-serving/scripts/convert-data/constants.py
khmoran: that's fair -- then miami the city needs to be added to the geo tsv. I'll do that instead

data-serving/scripts/convert-data/constants.py
khmoran: a lot of these are to disambugate when they match multiple admin regions, ex. madrid is the city and the county.

PR #374, opened 2020-06-30
Replace filter capability of linelist with full text search

attwad:
To keep things simpler I replaced the filtering of columns with full text search, support is pretty good, terms are OR'ed by default but oyu can also negate them so you can do stuff like: "India -twitter" to get all the India cases that are not coming out of twitter for example. It also supports "sentence match" where the exact terms are searched for. We might consider adding a AND checkbox or something to make it easy to do searches like  "curator@email India" for example for a curator to the cases they entered in india.

Also use `lean()` to return plain JSON objects when we don't need to return a document (GET calls mostly), this is apparently much more efficient.

### DIFF
--- a/data-serving/data-service/src/controllers/case.ts
+++ b/data-serving/data-service/src/controllers/case.ts
@@ -8,7 +8,7 @@ import { Case } from '../model/case';
  * Handles HTTP GET /api/cases/:id.
  */
 export const get = async (req: Request, res: Response): Promise<void> => {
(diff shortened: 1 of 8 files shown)


attwad:
[Implement basic text search functionality](https://trello.com/c/r8dZnF6A/261-implement-basic-text-search-functionality)

attwad:
[Implement basic text search functionality](https://trello.com/c/r8dZnF6A/261-implement-basic-text-search-functionality)

PR #378, opened 2020-06-30
Use FastField where possible to speed up the form

allysonjp715:
https://jaredpalmer.com/formik/docs/api/fastfield


allysonjp715:

PR #381, opened 2020-06-30
Make table scroll rather than entire page scroll

allysonjp715:
Screen Shot 2020-06-30 at 11 35 09 AM


attwad:

PR #384, opened 2020-06-30
add location and event details to details page

attwad:


verification/curator-service/ui/src/components/ViewCase.tsx
```diff
@@ -81,6 +81,15 @@ function ageRange(range?: { start: number; end: number }): string {
         : `${range.start}-${range.end}`;
 }
 
+function dateRange(range?: { start: string; end: string }): string {
```
allysonjp715, line 84: You could merge this and the ageRange function into one by having the inputs be string | number.

attwad: no because the dates ar ecurrently returned with "-" as their separators and joining them with "-" without space as we do for ages is awkward, that's why I added a space in the formatting below, hard to catch I admit :)

allysonjp715: I was thinking the age could also be separated with " - ", but not a big deal.

attwad:

PR #385, opened 2020-06-30
Update travel history and genome sequencing to spec

allysonjp715:
Screen Shot 2020-06-30 at 1 36 49 PM

### DIFF
--- a/verification/curator-service/ui/src/components/NewCaseForm.tsx
+++ b/verification/curator-service/ui/src/components/NewCaseForm.tsx
@@ -75,6 +75,7 @@ function initialValuesFromCase(c?: Case): NewCaseFormValues {
             transmissionRoutes: [],
             transmissionPlaces: [],
             transmissionLinkedCaseIds: [],
+            traveledPrior30Days: undefined,
             travelHistory: [],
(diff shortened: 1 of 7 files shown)


allysonjp715:

verification/curator-service/ui/src/components/NewCaseForm.tsx
```diff
@@ -533,9 +546,15 @@ class NewCaseForm extends React.Component<Props, NewCaseFormState> {
                                 }
                             >
                                 {this.tableOfContentsIcon({
-                                    isChecked: values.travelHistory?.length > 0,
+                                    isChecked:
+                                        values.travelHistory?.length > 0 ||
```
attwad, line 550: I don't think you need the .length > 0 here if you don't access the array after that

allysonjp715: The initial value of travelHistory is [] so we want to check if there are any entries in the array. This is currently an unnecessary check since the array will always be [] if `traveledPrior30Days === undefined`, but there is a bug out to change the criteria for a check in the table of contents, so I will leave it here until that changes.

PR #389, opened 2020-06-30
Add pathogens to NewCaseForm

allysonjp715:


allysonjp715:

verification/curator-service/ui/src/components/new-case-form-fields/Pathogens.tsx
```diff
@@ -0,0 +1,114 @@
...
+            const resp = await axios.get<string>(
+                'https://raw.githubusercontent.com/open-covid-data/healthmap-gdo-temp/master/suggest/pathogens.csv',
+            );
+            // CSV lines are of the form '123,Disease name' and we want to
+            // map that to [{'Disease name', 123}]
+            const retrievedOptions = new Map(
+                resp.data
+                    .split('\n')
+                    .map((option) => [
+                        option.substring(option.indexOf(',') + 1),
```
attwad, line 49: use string.split method with destructuration perhaps for better readability?
const id, value = {option.split(",")}

not sure if we can use destructuration with those constructs but if we can that's nice, otherwise at least split() would be nice as we don't have to deal with indexes.

verification/curator-service/ui/src/components/new-case-form-fields/Pathogens.tsx
```diff
@@ -0,0 +1,114 @@
...
+            itemType="string"
+            open={open}
+            onOpen={(): void => {
+                setOpen(true);
+            }}
+            onClose={(): void => {
+                setOpen(false);
+            }}
+            options={Array.from(options.keys())}
+            loading={loading}
```
attwad, line 83: you're missing initialValue for when we edit a case

verification/curator-service/ui/src/components/Case.tsx
```diff
@@ -95,7 +100,7 @@ export interface Case {
     sources: Source[];
     travelHistory: TravelHistory;
     genomeSequences: GenomeSequence[];
+    pathogens: Pathogen[];
```
attwad, line 108: pathogens?: Pathogen[];

allysonjp715: We haven't used ? for any other array fields like sources or genomeSequences. Any reason for it here in particular?

verification/curator-service/ui/src/components/new-case-form-fields/Pathogens.tsx
allysonjp715: That's defaultValue a couple lines down

verification/curator-service/ui/src/components/new-case-form-fields/Pathogens.tsx
allysonjp715: Done.

verification/curator-service/ui/src/components/Case.tsx
attwad: we should have been using those, if a case has no pathogen it won't be returned by the server os we should mark it as such to make sure that everywhere we access it we do check that it is present, this is what you're doing in this PR as well, you're using ? to access the field even though you didn't add it here it's inconsistent so let's add ? here.

allysonjp715: We use this Case object for EditCase and we're able to load that page with a partial case without a problem, but happy to add for general defensive code. Done.

allysonjp715: After a couple attempts I actually don't think it's a good idea to do this. It requires adding ? or undefined in loads of places. We don't have any issues with the server returning partial cases right now, so I don't think this helps us.

attwad: Well yes if you say this can be undefined you have to check for undefinedess everywhere that's WAI. If you say the server ALWAYS returns the pathogens field even when empty then we're fine and we don't need "?", otherwise we'll see stuff fail down the road.

allysonjp715: I just checked loading the EditCase page with a partial case (and found a couple missing ? along the way which I'll commit). Interestingly, when I printed the Case returned from the server it didn't have a pathogens field, however there were no errors either. Perhaps an object can always have a field not set, and that's different from setting an undefined value.

PR #390, opened 2020-06-30
Add a case data update script for dev

khmoran:
Apparently there's no way to do workflow composition or inheritance so I've duplicated the logic to keep a clean dev/prod separation -- and because they need to be triggered on different things.

Right now, *prod* mainly updates when the *data* should change: nightly after the expected CSV upload, and whenever the conversion script changes.

Now *dev* will mainly update when the *schema* should change, because we don't care about the data in dev being fresh -- but we don't want our dev instance to go down because the schema doesn't match the dataserver.

Ideally, we'd really trigger these things when we trigger the binary pushes, but I believe that's blocked ATM.

TESTED=Ran it locally with `act`

### DIFF
--- a/.github/workflows/case-data-update-dev.yml
+++ b/.github/workflows/case-data-update-dev.yml
@@ -7,7 +7,7 @@ on:
     paths: ["data-serving/data-service/schemas/**"]
 
 jobs:
-  # NOTE: This job should bee kept in sync with the prod version.
+  # NOTE: This job should be kept in sync with the prod version.
   update-case-data-dev:
     runs-on: ubuntu-latest
 
--- a/.github/workflows/case-data-update-dev.yml
+++ b/.github/workflows/case-data-update-dev.yml
@@ -1,0 +1,40 @@
+name: Update case data in dev
+
+on:
+  push:
+    # Update the data when the schema changes.
+    branches: [master]
+    paths: ["data-serving/data-service/schemas/**"]
+
+jobs:
+  # NOTE: This job should bee kept in sync with the prod version.
+  update-case-data-dev:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install MongoDB CLIs
+        run: |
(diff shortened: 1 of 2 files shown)


khmoran:
[Update case data update workflow to also update dev with a subset of the data](https://trello.com/c/5iuCz8zp/219-update-case-data-update-workflow-to-also-update-dev-with-a-subset-of-the-data)

.github/workflows/case-data-update-dev.yml
```diff
@@ -0,0 +1,40 @@
...
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Install MongoDB CLIs
+        run: |
+          wget -qO - https://www.mongodb.org/static/pgp/server-4.2.asc | sudo apt-key add -
+          echo "deb [ arch=amd64,arm64 ] https://repo.mongodb.org/apt/ubuntu xenial/mongodb-org/4.2 multiverse" | sudo tee /etc/apt/sources.list.d/mongodb-org-4.2.list
+          sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 5DC22404A6F9F1CA 656408E390CFB1F5
```
attwad, line 22: 👀 
/me what is this madness

.github/workflows/case-data-update-prod.yml
```diff
@@ -14,6 +14,7 @@ on:
     - cron: "0 5 * * *"
 
 jobs:
+  # NOTE: This job should bee kept in sync with the dev version.
```
attwad, line 17: git pull 🐝
aka. 🐝 kept in sync

.github/workflows/case-data-update-dev.yml
khmoran: NO QUESTIONS

ok fine, it was necessary because the script was having trouble looking up the keyserver for mongodb packages... it was weird... it might have just been a local problem but i'm not taking any chances

.github/workflows/case-data-update-prod.yml
khmoran: oooops. but it is more fun that way

🐝 our guest, 🐝 our guest, put our workjobs to the test!

PR #391, opened 2020-06-30
Delete unused fields in travel history and genome sequences

allysonjp715:


allysonjp715:

PR #392, opened 2020-06-30
add travel history to details page

attwad:


verification/curator-service/ui/src/components/fixtures/fullCase.json
```diff
@@ -110,16 +110,16 @@
         "travel": [
             {
                 "dateRange": {
-                    "start": "2020-01-10",
+                    "start": "2020-02-10",
```
allysonjp715, line 113: Why this change? It makes it an invalid date range

verification/curator-service/ui/src/components/Case.tsx
```diff
@@ -72,7 +73,7 @@ export interface Travel {
         end: string;
     };
     purpose?: string;
-    method?: string;
+    methods?: string[];
```
allysonjp715, line 76: I didn't know there could be multiple methods for one travel object - just saw that in the schema. We need to update the NewCaseForm for that. (Mostly a note to self, don't need to add it in this PR)

verification/curator-service/ui/src/components/fixtures/fullCase.json
attwad: because 2020-01-10 was already used somewhere else and findByText was complaining that it now found two occurences, we don't really check that start < end but yes it makes sense to have it valid, fixed.

PR #393, opened 2020-06-30
By default, drop only imported data as part of setup-db script

khmoran:
As of this change, `dev/setup_db.sh` will still drop all data from the local db, but other callers of this `setup-db` script (namely our update db workflows) will only drop imported data. This is to protect any data created via the portal, dev or prod (though we can discuss whether we want this for dev).

TESTED=I imported the sample data below, which I've modified to have 1 record that simulates coming from the new curator portal, i.e. it has no `importedCase` record. The script successfully drops the 99 "imported" documents and keeps the 

```
npm run setup-db

Read index from ./../../data-service/schemas/cases.index.json
Dropped imported documents (99 total) 🗎
Applied schema to existing collection "cases" 📑
Dropping indexes 👇
Creating indexes 👆
Done 👍
{ "collection" : "covid19.cases", "documents" : 1, "indexes" : 2 }
```

Now with `deleteAllDocuments`:
```
DELETE_ALL_DOCUMENTS=true npm run setup-cases

Read index from ./../../data-service/schemas/cases.index.json
Dropped all documents (1 total) 🗎
Applied schema to existing collection "cases" 📑
Dropping indexes 👇
Creating indexes 👆
Done 👍
{ "collection" : "covid19.cases", "documents" : 0, "indexes" : 2 }
Thank you for joining us 💾
```

### DIFF
--- a/data-serving/scripts/setup-db/src/index.ts
+++ b/data-serving/scripts/setup-db/src/index.ts
@@ -6,6 +6,8 @@ interface SetupDatabaseParameters {
     collectionName: string;
     schemaPath: string;
     indexPath: string;
+    /** If not specified, deletes only imported documents. Defaults to false. */
+    deleteAllDocuments: boolean;
 }
 
 const setupDatabase = async ({
@@ -14,6 +16,7 @@ const setupDatabase = async ({
     collectionName,
     schemaPath,
     indexPath,
+    deleteAllDocuments = false,
 }: SetupDatabaseParameters): Promise<void> => {
     try {
         const schema = JSON.parse(await cat(schemaPath));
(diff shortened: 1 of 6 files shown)


khmoran:
[Db update script should only drop imported data](https://trello.com/c/iprezrd1/260-db-update-script-should-only-drop-imported-data)

data-serving/scripts/setup-db/src/index.ts
```diff
@@ -30,31 +33,47 @@ const setupDatabase = async ({
...
-        print(`Created collection "${collectionName}" with schema`);
+            collection = await database.getCollection(collectionName);
+            const query = deleteAllDocuments
+                ? {}
+                : { importedCase: { $exists: true, $ne: null } };
+            const results = await collection.remove(query);
+            print(
+                `Dropped ${
+                    deleteAllDocuments ? 'all' : 'imported'
+                } documents (${results.nRemoved} total) 🗎`,
```
attwad, line 52: I am missing that emoji :(
image

khmoran: EEK! it was not a helpful one anyway, hopefully you can see this awesome trash can

PR #397, opened 2020-07-01
Add notes to details page

attwad:
They are not in the mocks, I left a question but for now this seems reasonable.


attwad:

PR #398, opened 2020-07-01
Finish details page

attwad:
Okay not really, there's the scroll-to-section panel left to do, will do it next ;)


attwad:

PR #400, opened 2020-07-01
add map to location components

attwad:
In new case form:
image

In details page:
image


attwad:

PR #402, opened 2020-07-01
Finalize base field format and parsing for bulk upload

axmb:


verification/curator-service/ui/src/components/BulkCaseForm.tsx
```diff
@@ -53,54 +92,120 @@ class BulkCaseForm extends React.Component<
...
+        } else if (c.admin1) {
+            return 'Admin1';
+        } else {
+            return 'Country';
+        }
+    }
+
+    createLocationQuery(c: ParsedCase): string {
+        return [c.admin3, c.admin2, c.admin1, c.country]
+            .filter((field) => field)
```
khmoran, line 146: does this filter do anything? does it filter out null/undefined somehow?

axmb: Yep:

image

If there's a cleaner way to do this, LMK!

verification/curator-service/ui/src/components/BulkCaseForm.tsx
```diff
@@ -43,6 +44,44 @@ interface BulkCaseFormValues {
...
+    // Demographics
+    sex?: string;
+    ageRangeStart?: number;
+    ageRangeEnd?: number;
+
+    // Events
+    dateConfirmed: string;
+    hospitalized?: boolean;
+    dateHospitalized?: string;
+    outcome?: string;
```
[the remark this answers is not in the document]
axmb: I'm using the `Case` type declared in this folder, where ~date fields are strings. I'd initially had these as Dates, but changed to strings to align with the NewCaseForm.

To your question, yes -- we should probably handle these as Dates where appropriate throughout. I can try to sneak it into the validation PR(s).

PR #404, opened 2020-07-01
add unknown check for genome sequence details

attwad:
Otherwise we'd get
```
TypeError: Cannot read property 'map' of undefined
    at (ViewCase.tsx:365)
...
```


attwad:

Issue #394, opened 2020-06-30
Define API to provide authenticated curator information to data service


attwad:
We need to define an API for the curator service to pass down the authentication details of the curator talking to it to the data service so that revisionMetadata can be set properly.

Notably right now it is impossible to increment the `revisionMetadata.revisionNumber` or set the `revisionMetadata.updateMetadata` field properly.

**Strawman proposal:**

- data service accepts a new `curator_email` field in the request body of create/update/upsert requests.
- curator service drops any `revisionMetadata` that clients send to it (to not trust them) and fills the `curator_email` field when talking to the data service based on the currently authenticated user.
- data service can then fetch the case, update the `revisionMetadata` accordingly, perform the udpate/upsert/create with the proper `curator` and `date` fields set in the `revisionMetadata.(creationMetadata|updateMetadata)` fields.

khmoran:
![](https://github.trello.services/images/mini-trello-icon.png) [Dataserver should populate (most) version data](https://trello.com/c/X0icPhIz/263-dataserver-should-populate-most-version-data)

PR #406, opened 2020-07-01
Remove "New" from NewCaseForm(Values) and send curator edits

attwad:
Sorry got git-snipped here where I got two PRs in one somehow :(

- one does s/NewCaseForm/CaseForm everywhere
- one sets the udpateMetadata upon edits, waiting for #394 to be properly implemented to show the last editor in the details page.


attwad:
> Is the NewCaseForm cypress test in this PR?

ah, actually no it's not been renamed but it only tests the "New" case... as we have the EditCase tests as well in a separate file... no big deal to leave it as is imo.

allysonjp715:
> > Is the NewCaseForm cypress test in this PR?
> 
> ah, actually no it's not been renamed but it only tests the "New" case... as we have the EditCase tests as well in a separate file... no big deal to leave it as is imo.

Ah true, sg.

PR #408, opened 2020-07-01
Separate symptoms and preexistingConditions mongoose schemas

khmoran:
Motivated by changing the `status` field in `preexistingConditions` to a bool (`hasPreexistingConditions`) -- now these two fields don't share the same structure anymore


khmoran:

data-serving/data-service/src/model/preexisting-conditions.ts
```diff
@@ -0,0 +1,12 @@
+import mongoose from 'mongoose';
+import { uniqueStringsArrayFieldInfo } from './unique-strings-array';
+
+export const preexistingConditionsSchema = new mongoose.Schema({
+    values: uniqueStringsArrayFieldInfo,
+    status: Boolean,
```
allysonjp715, line 6: s/status/hasPreexistingConditions

khmoran: ah good catch, thanks!

PR #411, opened 2020-07-01
Remove Mongoose's auto-added _id prop from subdocs

khmoran:


khmoran:

data-serving/data-service/schemas/cases.schema.json
```diff
@@ -13,9 +13,6 @@
         "bsonType": "object",
         "additionalProperties": false,
         "properties": {
-          "_id": {
```
axmb, line 16: For fields that we'll be searching Cases by, does this make any difference, with the assumption that we _won't_ be searching by `_id`? Off the top of my head, this (`caseReference.sourceUrl/Id`) might be the only one.

khmoran: I think the thing is that _id is indexed by default -- so we should just make sure to add an index on the caseReference.sourceId (or maybe URL) field

PR #414, opened 2020-07-02
Use remote autocomplete for the source UI form component

axmb:
Couple things we'll want to do subsequently:

- Add the `add source` dialogue. That'll enable us to remove `freeSolo` from the field, store the value as a full-fledged `Source` (in lieu of simply the URL), and then:
- Use this in the bulk entry UI.

source_ac


verification/curator-service/api/src/controllers/sources.ts
```diff
@@ -26,9 +26,14 @@ export default class SourcesController {
             res.status(422).json('limit must be > 0');
             return;
         }
+        const filter = req.query.url
```
attwad, line 29: would be nice to augment the open api spec with this new query param (fine in another PR)

verification/curator-service/ui/src/components/common-form-fields/Source.tsx
```diff
@@ -1,27 +1,151 @@
...
+        };
+    }, [value, inputValue, fetch]);
+
+    return (
+        <Autocomplete
+            // TODO: Replace freeSolo with offering an 'Add source' dialogue.
+            freeSolo
+            itemType="SourceData"
+            options={options}
+            value={value}
```
attwad, line 109: the other autocompletes have `defaultValue={props.initialValue}`, don't you need this to display a case source in the edit component?

verification/curator-service/ui/src/components/common-form-fields/Source.tsx
```diff
@@ -1,27 +1,151 @@
...
+    _id: string;
+    origin: OriginData;
+}
+
+interface ListSourcesResponse {
+    sources: SourceData[];
+}
+
+interface SourcesAutocompleteProps {
+    name: string;
```
allysonjp715, line 39: optional: If these are always going to be set as the same thing for the SourcesAutocomplete we don't have to pass them in as props.

verification/curator-service/api/src/controllers/sources.ts
axmb: Good call, done.

verification/curator-service/ui/src/components/common-form-fields/Source.tsx
axmb: Thanks for pointing that out -- indeed, the existing value won't be displayed in the edit component. Need to use a slightly different approach, here, since this component is shared by multiple forms (with different Value interfaces). Might be best to tackle this in a separate PR.

verification/curator-service/ui/src/components/common-form-fields/Source.tsx
axmb: I'd originally done this to allow us to avoid assumptions about the URL `Value` field name/whether this was required in the forms in which it appears, but you're right -- I'll scrap this now and can wire it back up if we need it.

axmb:
[Create a remote-autocomplete source UI component](https://trello.com/c/JiHQVVm2/265-create-a-remote-autocomplete-source-ui-component)

PR #417, opened 2020-07-02
pass down initial location nameto autocomplete field when editing a case

### LINKED ISSUE #416: Location name isn't shown in location input field when editing the case
**Describe the bug**
cf title

**To Reproduce**
1. add a case
2. edit the case
3. observe that the location field doesn't have the initial value set

attwad:
When something exists:

image

When you click on the field:
image

When you change:
image

Fixes #416

### DIFF
--- a/verification/curator-service/ui/src/components/new-case-form-fields/LocationForm.tsx
+++ b/verification/curator-service/ui/src/components/new-case-form-fields/LocationForm.tsx
@@ -2,6 +2,7 @@ import { Field, useFormikContext } from 'formik';
 import { Typography, makeStyles } from '@material-ui/core';
 
 import { Autocomplete } from '@material-ui/lab';
+import CaseFormValues from './CaseFormValues';
 import { Location as Loc } from '../Case';
 import Location from './Location';
 import LocationOnIcon from '@material-ui/icons/LocationOn';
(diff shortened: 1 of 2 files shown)


verification/curator-service/ui/src/components/new-case-form-fields/LocationForm.tsx
```diff
@@ -122,7 +128,7 @@ export function PlacesAutocomplete(
                 setInputValue(newInputValue);
             }}
             renderInput={(params): JSX.Element => (
-                <div>
+                <>
```
allysonjp715, line 131: Interesting. This can be used to group the fields into one Element to return?

verification/curator-service/ui/src/components/new-case-form-fields/LocationForm.tsx
```diff
@@ -132,7 +138,13 @@ export function PlacesAutocomplete(
                         name="unused"
                         required={props.required}
                         data-testid={props.name}
-                        label="Location"
+                        // Use the initial valuelocation name as a hint when untouched
```
allysonjp715, line 141: Is there not a defaultValue that can be used rather than changing the label?

attwad: nope, tried hard but {value}, {text}, {initialValue}, {defaultValue} etc nothing worked except label.

verification/curator-service/ui/src/components/new-case-form-fields/LocationForm.tsx
attwad: yes, I've been using it plenty, super useful :)

verification/curator-service/ui/src/components/new-case-form-fields/LocationForm.tsx
allysonjp715: Hm that's a shame. The material UI TextField component has a defaultValue field: https://material-ui.com/api/text-field/ Maybe we could try using that TextField rather than the formik-material-ui one? I thought all the formik-material-ui components passed through props to the material-ui component... shame it's not working as expected.

attwad: good idea, let me try with that field now and merge if I don't get anywhere...

attwad: nopenopenope considering this a _good enough_ fix for now to not not lose too much time on it...

FTR got a mixed bags of errors like
```
Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
```

or `TypeError: Cannot read property 'isSubmitting' of undefined` when trying to use something else than the current TextField :(

PR #422, opened 2020-07-02
Curator CUJ cypress test

attwad:
This basically gets rid of the fake data seeded in `addFullCase` and allows us to check that a case that was inputted from the NewCase form can be edited and viewed fully.

Test runtime is the same because most of the stuff added are checks that are instant, most of the test time is on the setup and typing in the form which is done once as before.


verification/curator-service/ui/cypress/integration/components/Curator.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+        cy.contains('2020-01-02');
+        cy.contains('2020-01-03');
+        cy.contains('2020-01-04');
+        cy.contains('2020-01-05');
+        cy.contains('2020-01-06');
+        cy.contains('PCR test');
+        cy.contains('Recovered');
+        cy.contains('Yes');
+        //  Symptoms.
+        cy.contains('dry cough, mild fever');
```
allysonjp715, line 285: Could you add `cy.contains('Presymptomatic');`

verification/curator-service/ui/cypress/integration/components/Curator.spec.ts
```diff
@@ -0,0 +1,306 @@
...
+        cy.contains('Germany');
+        cy.contains('Bus');
+        cy.contains('United Kingdom');
+        // Pathogens and genome.
+        cy.contains('Bartonella (232), Ebola (41)');
+        cy.contains('www.example2.com');
+        cy.contains('test sequence name');
+        cy.contains('33000');
+        cy.contains('testSequenceId');
+    });
```
allysonjp715, line 305: You could add deleting the case for full CRUD flow

verification/curator-service/ui/cypress/integration/components/Curator.spec.ts
attwad: done! sorry I forgot it

verification/curator-service/ui/cypress/integration/components/Curator.spec.ts
attwad: in that case I am not convinced this gives us anything more than the existing test that deletes a case as it doesn't matter whether the case was created with the form or with fake data.

PR #425, opened 2020-07-02
rename sequence id to accession as per moritz request

attwad:


verification/curator-service/ui/src/components/new-case-form-fields/GenomeSequences.tsx
```diff
@@ -87,7 +87,7 @@ export default function GenomeSequences(): JSX.Element {
                                                     className={classes.field}
                                                     name={`genomeSequences[${index}].sequenceId`}
```
allysonjp715, line 88: I wonder if we should change the name throughout the schema. I realize that is a lot more work... but it could get confusing if they are different. For example on bulk upload are the csv column headers the same name as the schema field?

attwad: whoa I hope they are not, and... confirmed that they are not.

I don't want to change the schema for that, considering this too risky just before we both go out and in general not worth it. It's still just an "ID" really (https://www.ncbi.nlm.nih.gov/genbank/sequenceids/), "accession" is more of a specialist term afaiu but in a schema it's an "ID".

PR #431, opened 2020-07-02
Don't validate on change

allysonjp715:
Validating the entire form on every change was making the form very laggy. Now it will validate on blur and form submission, and the form is responsive again.


allysonjp715:

PR #433, opened 2020-07-02
Some date picker enhancements

allysonjp715:
- Dialog closes when you select a date
 - Date pickers open up to show the confirmed date, hopefully saving them from having to scroll through the months multiple times.


allysonjp715:

PR #436, opened 2020-07-03
material table for users

attwad:
This allows for pagination.
I considered moving the tag approach as in https://material-ui.com/components/selects/#multiple-select because it looks good but figured that'd be in another PR.


verification/curator-service/ui/src/components/Users.test.tsx
```diff
@@ -136,14 +135,9 @@ test('updates roles on selection', async () => {
     expect(mockedAxios.put).toHaveBeenCalledWith('/api/users/abc123', {
         roles: ['admin', 'reader', 'curator'],
     });
-    expect(await findByText('Alice Smith')).toBeInTheDocument();
```
allysonjp715, line 139: Why remove these checks?

verification/curator-service/ui/src/components/Users.test.tsx
```diff
@@ -188,7 +178,7 @@ test('calls callback when user is changed', async () => {
         headers: {},
     };
     mockedAxios.put.mockResolvedValueOnce(axiosPutResponse);
-    expect(functionCalledCounter).toBe(0);
+    expect(mockCallback).toHaveBeenCalledTimes(0);
```
allysonjp715, line 181: cool!

verification/curator-service/ui/src/components/Users.tsx
```diff
@@ -13,42 +15,22 @@ interface ListResponse {
...
-    name: string;
-    email: string;
-    roles: string[];
-}
-
 interface UsersState {
     users: User[];
     availableRoles: string[];
     url: string;
+
```
allysonjp715, line 22: nit: remove empty line

verification/curator-service/ui/src/components/Users.tsx
```diff
@@ -85,113 +66,145 @@ class Users extends React.Component<Props, UsersState> {
...
+                        })
+                    }
+                    title="Users"
+                    options={{
+                        search: false,
+                        filtering: false,
+                        sorting: false,
+                        padding: 'dense',
+                        draggable: false, // No need to be able to drag and drop headers.
+                        pageSize: 10,
```
allysonjp715, line 142: I had to make some changes to get choosing a pageSize to work in the other tables, could you add that here? https://github.com/open-covid-data/healthmap-gdo-temp/pull/548/files

verification/curator-service/ui/src/components/Users.tsx
```diff
@@ -85,113 +66,145 @@ class Users extends React.Component<Props, UsersState> {
...
-        updatedUser: User,
+        userId: string,
     ): void {
         axios
-            .put(this.state.url + updatedUser._id, {
+            .put(this.state.url + userId, {
                 roles: event.target.value,
             })
             .then(() => {
                 const updatedUsers = this.state.users.slice();
```
allysonjp715, line 159: Now that it's a table can we get rid of lines 159 -> 166 (RHS) and let `this.tableRef.current.onQueryChange();` do the updating?

verification/curator-service/ui/src/components/Users.test.tsx
attwad: contents are checked in the cypress tests, didn't like the redundancy and the fact that we have to mock again and again the axios responses here, we're just checking that the document contains whatever we put in our mocks in unit tests which feels useless to me.

verification/curator-service/ui/src/components/Users.tsx
attwad: done.

verification/curator-service/ui/src/components/Users.tsx
attwad: ah nice, thanks for pointing it out. Done.

verification/curator-service/ui/src/components/Users.tsx
attwad: you are right, much simpler now, thanks

PR #451, opened 2020-07-07
Fill source URL when a case is edited.

axmb:


axmb:
[Create a remote-autocomplete source UI component](https://trello.com/c/JiHQVVm2/265-create-a-remote-autocomplete-source-ui-component)

PR #461, opened 2020-07-08
Use the source component (in lieu of the CSV) to derive source ID/URL for bulk upload

axmb:
Couple of changes made consequently:

- Make the source component required.
- Add Yup form validation.

I'm also enacting the same test strategy we have for the `CaseForm` component: make the unit test check bare bones render logic, and do all the interactive stuff in Cypress.


axmb:
[S#6: Populate CaseReference from Source form component using autocomplete/creation dialog](https://trello.com/c/Jq7ZsO22/267-s6-populate-casereference-from-source-form-component-using-autocomplete-creation-dialog)

verification/curator-service/ui/cypress/integration/utils/enterSource.ts
```diff
@@ -1,8 +1,16 @@
...
+export default function enterSource(url: string, existingSource = false): void {
+    if (existingSource) {
+        cy.get('div[data-testid="caseReference"]').type(
+            `${url}{downarrow}{enter}`,
+        );
+    } else {
+        cy.get('div[data-testid="caseReference"]').type(
+            `${url}{downarrow}{enter}`,
+        );
+        cy.get('input[id="name"]').type('New source');
```
allysonjp715, line 10: Do we need unique names across URLs?

verification/curator-service/ui/src/components/BulkCaseForm.tsx
```diff
@@ -250,6 +265,7 @@ class BulkCaseForm extends React.Component<
         const { classes } = this.props;
         return (
             <Formik
+                validationSchema={BulkFormSchema}
```
allysonjp715, line 268: You may also want to include `validateOnChange={false}` That way the form validation only kicks in on blur and form submission. It's a nicer UX (doesn't start showing error messages as you are typing) and also greatly decreased lag on the CaseForm.

verification/curator-service/ui/cypress/integration/utils/enterSource.ts
axmb: Nope, the name doesn't need to be unique (which is reasonable, IMO). What we'll likely want to implement, though, is a safeguard against duplicate URLs. I'll chat with Kelly about this next week -- I think ensuring the uniqueness at the data layer makes sense.

verification/curator-service/ui/src/components/BulkCaseForm.tsx
axmb: Good call, I'll add.

PR #465, opened 2020-07-08
Don't require method of confirmation

allysonjp715:
It can have value "Unknown" which we are currently saving as undefined in the schema.


allysonjp715:

allysonjp715:
> Is method of confirmation an item we're enum'ing in the UI (via suggestions)? I don't see an enum in our schema for it. I imagine it'd be a little non-trivial, given we'd have to enforce the validation specifically on the `value` of the `confirmed` event. WDYT?

No we don't have an enum in the schema, but it is a dropdown selection in the UI.
Screen Shot 2020-07-08 at 6 01 19 PM
Yea the schema validation may be tricky, but it could be an API check. I do have a TODO in the UI to get the dropdown values from the backend.

PR #473, opened 2020-07-09
Add a validate_only param to the case creation API

axmb:
This allows bulk ingestion to check all cases prior to saving them to the DB, to avoid partial uploads.

I'm using a query param in the existing creation API, which is a [relatively canonical technique](https://google.aip.dev/163).

Spotted a conflict between our eslint config and prettier in the controller file. So including a tiny fix to our eslint configs that properly delegates indentation formatting to prettier.


axmb:

PR #474, opened 2020-07-09
Create an API-ready case interface in BulkCaseForm to facilitate forthcoming validation

axmb:
Primarily need to define a type in the component to hold case data such that we can cleanly construct each case once, and reuse the object for validation and eventual submission. Cutting this as a separate PR so that the eventual surfacing of validation errors isn't too massive/dense.

See the note on the partial type I've created. I'm hoping that we eventually find cycles to label fields optional in `Case` if they're optional in the API case document.

### DIFF
--- a/verification/curator-service/ui/src/components/BulkCaseForm.tsx
+++ b/verification/curator-service/ui/src/components/BulkCaseForm.tsx
@@ -214,17 +214,6 @@ class BulkCaseForm extends React.Component<
         return axios.put('/api/cases', c);
     }
 
-    async validateParsedCases(
-        results: ParseResult<RawParsedCase>,
-        caseReference: CaseReference,
-    ): Promise<Array<string>> {
-        return [];
-    }
-
-    surfaceErrorMessage(errors: Array<string>): void {
-        return;
-    }
-
     async uploadData(
         results: ParseResult<RawParsedCase>,
         caseReference: CaseReference,
--- a/verification/curator-service/ui/src/components/BulkCaseForm.tsx
+++ b/verification/curator-service/ui/src/components/BulkCaseForm.tsx
@@ -87,6 +87,14 @@ interface RawParsedCase {
     caseCount?: number;
 }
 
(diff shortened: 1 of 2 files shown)


axmb:

verification/curator-service/ui/src/components/BulkCaseForm.tsx
```diff
@@ -79,14 +79,27 @@ interface ParsedCase {
...
+    latitude?: number;
+    longitude?: number;
     locationName?: string;
 
     // Bulk upload specific data
     caseCount?: number;
 }
 
+type RecursivePartial<T> = {
+    [P in keyof T]?: RecursivePartial<T[P]>;
```
allysonjp715, line 91: 🤔  This is pretty confusing, could you add a comment to the type describing it's use?

verification/curator-service/ui/src/components/Case.tsx
```diff
@@ -45,6 +45,7 @@ export interface Location {
     geometry: Geometry;
     name: string;
     place: string;
+    query?: string;
```
allysonjp715, line 48: Is this needed? The type description says `// Case definitions as returned by the /api/cases endpoint.` and I don't think query is ever returned from the backend, is it? If it's needed perhaps change the type description.

axmb: Good call out, I'll change the type description. This is required to POST cases without a prior geocode. In the new case form, we use the geocode `suggest` endpoint, which takes care of completing the `location` object prior to use POSTing the new case. But for non-human-interactive usage, we're currently requiring a `query` field in `location` [in order to geocode](https://github.com/open-covid-data/healthmap-gdo-temp/blob/master/verification/curator-service/api/src/controllers/cases.ts#L124).

verification/curator-service/ui/src/components/BulkCaseForm.tsx
axmb: Yeah, sorry about that. Done.

PR #522, opened 2020-07-10
Preserve proxied error status/data in curator case API.

axmb:
Ran into this while implementing bulk upload errors.

Previously, the curator service would completely swallow the nature of the error. E.g., for validation issues it'd return something along the lines of:

```
{
  code: 500
  message: 'Request failed with code 422.'
}
```

Which has the downside of (1) removing helpful information as to the nature of the error (say, a specific validation issue), and (2) casts all errors (except geocoding 422s) to 500. What I've added isn't totally perfect, and assumes (probably safely) that we're satisfied with just catching errors made using axios and which received a response from the case server. But, that seems fair enough for now.


axmb:

Issue #475, opened 2020-07-09
Batched UX Behaviour Bugs


shroothless:
**Describe the bug**
Multiple UX Behaviour Bugs

**To Reproduce**
Noted in deck

**Expected behavior**
https://docs.google.com/presentation/d/1Fg-UYsIB8oq03-VIHR0CP8G6ym02HxjCpyDcroheFzc/edit?usp=sharing

**Screenshots**
Noted in deck

**Environment (please complete the following information):**
 - OS: OS Mojave
 - Browser Chrome

allysonjp715:
Thanks for filing Shru! I'll post updates in the deck as things are fixed.

allysonjp715:
The only bug left in the deck is adding the tooltips to the sections which is blocked on copy. @shroothless who created the copy for the fields? Could they also add some for the sections? Thanks!

allysonjp715:
I see what I asked for is covered in #562. I'll close this bug out and track there. Feel free to reopen or create another bug for more UX issues.

PR #523, opened 2020-07-10
Have CaseForm show up as a modal

allysonjp715:
Also adds the create new menu.

Screenshots because the gif file size is too big :( 

Screen Shot 2020-07-10 at 5 06 38 PM
Screen Shot 2020-07-10 at 5 06 28 PM

### DIFF
--- a/verification/curator-service/ui/cypress/integration/components/EditCase.spec.ts
+++ b/verification/curator-service/ui/cypress/integration/components/EditCase.spec.ts
@@ -9,11 +9,6 @@ describe('Edit case', function () {
         cy.clearSeededLocations();
     });
 
-    it('Errors when case does not exist', function () {
-        cy.visit('/cases/edit/foo');
-        cy.contains('Request failed');
-    });
-
     // Full case edit is covered in the curator tests.
     it('can edit a case', function () {
         cy.addCase({
@@ -26,6 +21,8 @@ describe('Edit case', function () {
         cy.request({ method: 'GET', url: '/api/cases' }).then((resp) => {
             expect(resp.body.cases).to.have.lengthOf(1);
             cy.visit(`/cases/edit/${resp.body.cases[0]._id}`);
+            cy.visit('cases');
+            cy.get('button[title="Edit this case"]').click();
             // Check that we have something from the original case.
             cy.contains('France');
             cy.contains('Andorrean');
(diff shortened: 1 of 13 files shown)


allysonjp715:

allysonjp715:
Addresses some of #475

verification/curator-service/ui/cypress/integration/components/EditCase.spec.ts
```diff
@@ -9,11 +9,6 @@ describe('Edit case', function () {
         cy.clearSeededLocations();
     });
 
-    it('Errors when case does not exist', function () {
```
axmb, line 12: What's the new behavior, here?

verification/curator-service/ui/src/components/App.tsx
```diff
@@ -216,7 +241,53 @@ class App extends React.Component<Props, State> {
                         }}
                     >
                         <div className={classes.drawerHeader}></div>
-                        <Divider />
+                        {this.hasAnyRole(['curator']) && (
```
axmb, line 244: Should this be a separate component?

verification/curator-service/ui/src/components/CaseForm.tsx
```diff
@@ -458,338 +483,459 @@ class CaseForm extends React.Component<Props, CaseFormState> {
...
-                                {isSubmitting && <LinearProgress />}
-                                <br />
-                                <Button
-                                    variant="contained"
-                                    color="primary"
-                                    data-testid="submit"
-                                    disabled={isSubmitting}
-                                    onClick={submitForm}
+            <>
+                <Modal open={true}>
```
axmb, line 494: The bulk case form is also mocked as a modal.

Is there any way we can share the modal logic as a discrete component that contains the contents of either form? Or partially share it?

verification/curator-service/ui/cypress/integration/components/EditCase.spec.ts
allysonjp715: There's no more URL for edit case so we can no longer trigger the edit case with an invalid ID.

verification/curator-service/ui/src/components/CaseForm.tsx
allysonjp715: Definitely worth considering. I kept the bulk case change out of this PR since it was already pretty big. Let's keep this in mind when we change the bulk case to a modal.

verification/curator-service/ui/src/components/App.tsx
allysonjp715: Perhaps we can pull it out if we use another menu button somewhere else in the UI in the future

PR #524, opened 2020-07-10
Create a class to store bulk case validation errors.

axmb:


axmb:

verification/curator-service/ui/src/components/bulk-case-form-fields/CaseValidationError.test.ts
```diff
@@ -0,0 +1,29 @@
+import CaseValidationError from './CaseValidationError';
```
allysonjp715, line 1: Perhaps we should also have an integration test so that if the real error messages change we know that we need to change our parsing logic. Though I'm not sure we can trigger backend error messages right now with the UI validation we have in place...

verification/curator-service/ui/src/components/bulk-case-form-fields/CaseValidationError.ts
```diff
@@ -0,0 +1,33 @@
...
+                    message: parts.slice(1).join().trim(),
+                };
+            })
+            .sort((a, b) => {
+                return a.field > b.field ? 1 : -1;
+            });
+        this.formattedIssues = sortedErrorsByField
+            .filter(
+                (a, index) =>
+                    !sortedErrorsByField
```
allysonjp715, line 27: I'm not quite following what this filtering is doing, could you add a comment?

verification/curator-service/ui/src/components/bulk-case-form-fields/CaseValidationError.test.ts
axmb: Great suggestion. I can add this to the Cypress test edited in a subsequent PR, or, we should be able to check this in the case API supertest tests.

verification/curator-service/ui/src/components/bulk-case-form-fields/CaseValidationError.ts
axmb: Yep, done. Added a top-level comment detailing the creation logic.

PR #527, opened 2020-07-13
Move reusable parsing subroutines into a common lib file/Lambda Layer

axmb:
Diff appears large, but it's mostly just moving code around.

The code being moved isn't optimal -- there are a number of things (in the API, and in Pythonicisms) that should be changed. As a Python expert I'm sure you'll have comments on it! Feel free to comment, or, it might simply be better to sync offline and for us to make changes in subsequent PR(s). And of course please do comment on anything that isn't a drag-and-drop.

There are a couple oddities (like the way parsers import the lib), that are mostly a function of the way AWS Lambda Layers work. The import shenanigans are specifically so that we can run pytest unit tests in addition to invoking the function locally via `sam local invoke`. [This post is a pretty good summary of what this change does](https://aws.amazon.com/blogs/compute/working-with-aws-lambda-and-lambda-layers-in-aws-sam/), albeit with node as the example runtime. In separate PRs, there are some things I'd like to tweak, including:

- Seeing if we can shove the dependencies (as in `requirements.txt`) to the Layer level (in lieu of per-parser).
- Seeing if we can shove os env params (e.g. params configured in `template.yaml`) to the Layer level.

On my to-do today is poking some folks to try and sort the question around AWS credentials and running locally. In the meantime, I'd be happy to set you up an account so that you can run the SAM commands, if you'd like.

### DIFF
--- a/ingestion/functions/parsing/common/python/parsing_lib.py
+++ b/ingestion/functions/parsing/common/python/parsing_lib.py
@@ -1,0 +1,87 @@
+import boto3
+import os
+import requests
+import google.auth.transport.requests
+from google.oauth2 import service_account
+
+LOCAL_DATA_FILE = "/tmp/data.json"
+METADATA_BUCKET = "epid-ingestion"
+SERVICE_ACCOUNT_CRED_FILE = "covid-19-map-277002-0943eeb6776b.json"
+SOURCE_URL_FIELD = "sourceUrl"
+S3_BUCKET_FIELD = "s3Bucket"
+S3_KEY_FIELD = "s3Key"
+S3_KEY_PATH_SEPERATOR = "/"
+
+s3_client = boto3.client("s3")
+
+
+def extract_event_fields(event):
+    if any(
+            field not in event
+            for field in [SOURCE_URL_FIELD, S3_BUCKET_FIELD, S3_KEY_FIELD]):
+        error_message = (
+            f"Required fields {SOURCE_URL_FIELD}; {S3_BUCKET_FIELD}; "
+            f"{S3_KEY_FIELD} not found in input event json.")
+        print(error_message)
+        raise ValueError(error_message)
+    return event[SOURCE_URL_FIELD], event[S3_BUCKET_FIELD], event[S3_KEY_FIELD]
+
+
+def retrieve_raw_data_file(s3_bucket, s3_key):
+    try:
+        local_data_file = LOCAL_DATA_FILE
+        print(f"Retrieving raw data from s3://{s3_bucket}/{s3_key}")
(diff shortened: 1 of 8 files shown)


ingestion/functions/parsing/india/india.py
```diff
@@ -1,44 +1,17 @@
...
-        raise e
+# Layer code, like parsing_lib, is added to the path by AWS.
+# To test locally (e.g. via pytest), we have to modify sys.path.
+# pylint: disable=import-error
+if ('lambda' not in sys.argv[0]):
+    sys.path.append(
+        os.path.join(
+            os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
+            'common/python'))
+import parsing_lib
```
iamleeg, line 14: Like I said, I have other problems getting this to work so I haven't tried the following change. You should be able to import a relative file (which should work both locally and on AWS) with `import ..parsing.parsing_lib`, unless there's something about the AWS environment that doesn't allow that.

iamleeg: Ignore this, I managed to fix my venv (actually I don't know what I did, it just started working AFAICT) and relative imports can't be above the top-level module in the filesystem.

ingestion/functions/parsing/common/python/parsing_lib.py
```diff
@@ -0,0 +1,87 @@
...
+        print(e)
+        raise e
+
+
+def write_to_server(cases, headers):
+    """
+    Upserts the provided cases via the G.h Case API.
+
+    TODO: Parallelize these requests.
+    """
```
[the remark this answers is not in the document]
axmb: Yeah, fully agree. Let me slap another TODO on, and update this after Kelly submits the API resource documentation.

ingestion/functions/parsing/common/python/parsing_lib.py
```diff
@@ -0,0 +1,87 @@
+import boto3
```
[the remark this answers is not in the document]
axmb: Heh, SG. Assuming resolved re: comment below, but do reopen if there's a persistent issue.

ingestion/functions/parsing/india/india.py
axmb: Yep, ran into that error as well (I believe due to Python file "meta" naming -- as in the argv value -- when running a file directly via `__main__`, as is done with pytest).

axmb:
[Prep ingestion parsing library for contributions from exo-G.h developers](https://trello.com/c/4tBvIME2/249-prep-ingestion-parsing-library-for-contributions-from-exo-gh-developers)

PR #535, opened 2020-07-13
Collapse a Cypress test chain.

axmb:
Missed this commit in #526.


axmb:

PR #538, opened 2020-07-13
Fix required fields UI

allysonjp715:
- Use "Required" instead of "Required field"
- Have "Required" always be present, and change text color to red when error is triggered
- Remove asterisks

### DIFF
--- a/verification/curator-service/ui/src/components/common-form-fields/FormikFields.tsx
+++ b/verification/curator-service/ui/src/components/common-form-fields/FormikFields.tsx
@@ -178,7 +178,6 @@ export function DateField(props: DateFieldProps): JSX.Element {
                     name={props.name}
                     label={props.label}
                     format="yyyy/MM/dd"
-                    required={props.required}
                     minDate={new Date('2019/12/01')}
                     disableFuture
                     autoOk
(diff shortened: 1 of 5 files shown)


allysonjp715:
Fixes https://docs.google.com/presentation/d/1Fg-UYsIB8oq03-VIHR0CP8G6ym02HxjCpyDcroheFzc/edit#slide=id.g8b4f04bb4d_0_23 and addresses some of #475

allysonjp715:

verification/curator-service/ui/src/components/common-form-fields/FormikFields.tsx
```diff
@@ -205,14 +204,18 @@ export function RequiredHelperText(
...
+                error={
+                    hasKey(touched, props.name) &&
+                    touched[props.name] &&
+                    hasKey(values, props.name) &&
+                    (values[props.name] === undefined ||
+                        values[props.name] === null ||
+                        values[props.name] === '')
+                }
+            >
+                Required
```
axmb, line 217: Can you remind me what the distinction is between the function of this text, and the Yup `required` string?

allysonjp715: I don't think the Yup required string is technically used anywhere yet. Formik keeps track of all the form errors in an object called errors. The Yup required string dictates what shows up in that Formik object. Though that isn't shown in the UI anywhere currently, it may be useful in the future and is also useful when debugging (I frequently print the contents of Formik's `error` to see what's going on in the form).

PR #540, opened 2020-07-13
Remove enums from the mongodb json schema in favor of strings

khmoran:
Note that these are still validated by Mongoose in the data server so this is a no-op

### DIFF
--- a/data-serving/data-service/schemas/cases.schema.json
+++ b/data-serving/data-service/schemas/cases.schema.json
@@ -58,11 +58,7 @@
             }
           },
           "sex": {
-            "enum": [
-              "Male",
-              "Female",
-              "Other"
-            ]
+            "bsonType": "string"
           },
           "profession": {
             "bsonType": "string"
@@ -127,13 +123,7 @@
             "bsonType": "string"
           },
           "geoResolution": {
-            "enum": [
-              "Point",
-              "Admin3",
-              "Admin2",
-              "Admin1",
-              "Country"
-            ]
+            "bsonType": "string"
           },
           "geometry": {
             "bsonType": "object",
@@ -197,11 +187,7 @@
             }
           },
           "status": {
-            "enum": [
-              "Asymptomatic",
-              "Presymptomatic",
-              "Symptomatic"
-            ]
+            "bsonType": "string"
           }
         }
       },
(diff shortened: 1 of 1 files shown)


khmoran:
[Allow free-form text entry for "Other" fields](https://trello.com/c/BhMc3HjT/275-allow-free-form-text-entry-for-other-fields)

data-serving/data-service/schemas/cases.schema.json
```diff
@@ -197,11 +187,7 @@
             }
           },
           "status": {
-            "enum": [
-              "Asymptomatic",
-              "Presymptomatic",
-              "Symptomatic"
-            ]
+            "bsonType": "string"
```
allysonjp715, line 190: I don't think this one will have an Other field

data-serving/data-service/schemas/cases.schema.json
```diff
@@ -127,13 +123,7 @@
...
           },
           "geoResolution": {
-            "enum": [
-              "Point",
-              "Admin3",
-              "Admin2",
-              "Admin1",
-              "Country"
-            ]
+            "bsonType": "string"
```
allysonjp715, line 126: We can keep the enum here

data-serving/data-service/schemas/cases.schema.json
```diff
@@ -255,13 +241,7 @@
...
                     },
                     "geoResolution": {
-                      "enum": [
-                        "Point",
-                        "Admin3",
-                        "Admin2",
-                        "Admin1",
-                        "Country"
-                      ]
+                      "bsonType": "string"
```
allysonjp715, line 244: Same as above

data-serving/data-service/schemas/cases.schema.json
khmoran: Per discussion documented [here](https://docs.google.com/document/d/1u0FUPZ6vG00irfPaw7CLwekWxpeE7n70WlRRnMQAi-g/edit#heading=h.rijffgfpldij), we're going to remove all enums from the schema in order to separate the schema from the data; data values will be validated at the API layer instead

data-serving/data-service/schemas/cases.schema.json
khmoran: See above

data-serving/data-service/schemas/cases.schema.json
khmoran: See above

PR #553, opened 2020-07-14
Add vertical scroll to table as opposed to window

allysonjp715:
Before:
Screen Shot 2020-07-14 at 3 51 06 PM

After:
Screen Shot 2020-07-14 at 3 51 19 PM


allysonjp715:

allysonjp715:
The cypress tests don't seem to want to use the table's built in scroll, so had to add some `{force:true}`s

allysonjp715:
> Should we make the height of the table a little larger? Now that it's fit to the window, it feels a little vertically cramped in the `After` picture.

I think it's best to fit it to the window. Otherwise there will be two scroll bars right next to each other and the pagination will be out of the initial visible window

PR #556, opened 2020-07-14
s/sex/gender and add more options

khmoran:


data-serving/data-service/test/model/data/demographics.full.json
```diff
@@ -3,7 +3,7 @@
         "start": 20,
         "end": 29
     },
-    "sex": "Female",
+    "gender": "Female",
```
axmb, line 6: Now that this is gender, rather than sex, should some test data cover values aside from "Male"/"Female"?

khmoran: Good call, changed up 3 tests (one for Non-binary/Third gender and one for Other)

PR #557, opened 2020-07-14
s/profession/occupation

khmoran:


data-serving/data-service/src/model/demographics.ts
```diff
@@ -17,8 +17,7 @@ export const demographicsSchema = new mongoose.Schema(
             _id: false,
         },
         sex: String,
-        // TODO: The below 3 fields should be data dictionaries.
```
axmb, line 20: Intentionally removed?

khmoran: yeah since I don't think this layer is going to handle the dictionaries at all (will be in API layer)

PR #558, opened 2020-07-14
s/ethnicity/race \/ ethnicity and make it free-form text field

### LINKED ISSUE #555: Ethnicity field change
Request to make 2 changes to the ethnicity field

1. Change the ethnicity field title to: Race / Ethnicity
2. Change the input to free text field only, no drop down or pre selects. 

Rationale: Want to allow curators the most inclusive and broadest set of options for data entry. 

_**Researcher feedback: People would prefer an inclusive large selection of options which needs manual work, rather than an exclusive subset of the totality of what is out there**_

khmoran:
Closes #555

### DIFF
--- a/verification/curator-service/ui/src/components/new-case-form-fields/Demographics.tsx
+++ b/verification/curator-service/ui/src/components/new-case-form-fields/Demographics.tsx
@@ -38,16 +38,6 @@ type DemographicsProps = WithStyles<typeof styles>;
 // TODO: get values from DB.
 const sexValues = [undefined, 'Male', 'Female'];
 
-const ethnicityValues = [
-    undefined,
-    'Asian',
-    'Black',
-    'Latino',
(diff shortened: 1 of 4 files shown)


verification/curator-service/ui/src/components/new-case-form-fields/Demographics.tsx
```diff
@@ -100,26 +90,15 @@ function Demographics(props: DemographicsProps): JSX.Element {
...
-                                </MenuItem>
-                            ))}
-                        </FastField>
-                    </FormControl>
+                    <FastField
+                        label="Race / Ethnicity"
+                        name="ethnicity"
+                        type="text"
+                        data-testid="ethnicity"
+                        multiline={true}
```
allysonjp715, line 98: Probably don't want multiline here

Issue #530, opened 2020-07-13
Bulk CSV ingestion: Admin level 2 data selecting specific road location 


sratcliffe118:
Steps to recreate:
1. Adding Florida data which provide location data to county level
2. Add county information to level 3 data: Broward
3. Populates Broward road, jacksonville in the line list entry
See attached

Expected: Admin level 3 should be county, city or town level, should not go down to specific road depth

![Screen Shot 2020-07-13 at 1 19 52 PM](https://user-images.githubusercontent.com/63205311/87303762-9565a480-c50b-11ea-90b4-3b5fc9419923.png)

sratcliffe118:
Another example is case: 5f0c8630a17f74e5a82fb77b
This admin level 3 data was: CORONEL FABRICIANO/TIMOTEO
The location added to the case: Rua Coronel Fabriciano, Centro Sul, Timóteo - Minas Gerais, 35182, Brazil

axmb:
Thoughts on the results:

For Florida, I think that's a subset of the existing issue around US county identification.

Brazil is somewhat more complicated. To start, that admin3 example data seems to be compound -- Coronel Fabriciano and Timóteo are both municipalities. So I suppose the slash indicates they're joint reporting, or the data is from one of those two locations. For that data, I think we should be entering:

country: Brazil
admin1: Minas Gerais
admin2: Valo do Aço
admin3: Either blank (CF and Timóteo are two of four municipalities in VofA), Coronel Fabriciano, or Timóteo. Granted this field isn't in the bulk format, yet (I'll add it so we can do this kind of testing).

Thoughts on fixes:

We can't "force" the geocoding resolution until the county bug is resolved. If we do, we get 404 for queries like "Broward, Florida, United States."

Do we want to leave a P1 here to limit geocoding resolution, blocked on figuring out our issue with counties?

axmb:
I've added an option to tweak the way we invoke the mapbox API (see the above two PRs).

Does the behavior I've gif'd in #561 seem reasonable?

(Separately, I've documented that `admin3` can be supplied as a field to bulk upload.)

sratcliffe118:
Alex, reviewed and this approach LGTM

Think this example highlights that no matter how automated we make the process, as the data is reported in a fragmented way curators will still be required to make some judgement calls and data manipulation before bulk uploading.

PR #560, opened 2020-07-14
Allow specifying multiple geocode limits/types.

axmb:
Adding a feature to limit the impact of issues like #530 while we wait on resolving admin2 geocoding.

### DIFF
--- a/verification/curator-service/api/src/geocoding/mapbox.ts
+++ b/verification/curator-service/api/src/geocoding/mapbox.ts
@@ -91,12 +91,20 @@ export default class MapboxGeocoder {
             // with features of that specific type.
             // The context of the feature can still has other types
             // (if you ask for a region, the context will still contain the country).
-            if (opts?.limitToResolution) {
-                const type = resolutionToGeocodeQueryType.get(
-                    opts?.limitToResolution,
-                );
-                if (type) {
-                    req.types = [type];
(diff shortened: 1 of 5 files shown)


verification/curator-service/api/src/geocoding/mapbox.ts
```diff
@@ -91,12 +91,20 @@ export default class MapboxGeocoder {
...
-                );
-                if (type) {
-                    req.types = [type];
+            const requestedResolutions = opts?.limitToResolution;
+            if (requestedResolutions) {
+                const types = requestedResolutions
+                    .map((r: Resolution) => resolutionToGeocodeQueryType.get(r))
+                    .filter((t) => t);
+                if (types) {
+                    req.types = [];
```
khmoran, line 100: can you do the filter & assignment all in one again here `req.types = types.filter((t) => t)` or does it still give you type issues?

axmb:

verification/curator-service/api/src/geocoding/mapbox.ts
axmb: I'd started as one line and expanded until Typescript took the hint. It seems that `(t) => t` (or even, strangely, `t !== undefined`) isn't sufficient to register the content of the array as `string`.

PR #565, opened 2020-07-15
Add top-level NewCase properties to API docs & use to validate create requests

khmoran:
Note: Also needed to move the express handlers for API endpoints to be registered after the open api validator is installed -- without this, it wasn't actually validating

Next few PRs will be fleshing out the sub-props

image


khmoran:

PR #566, opened 2020-07-15
Enable request & response validation in data server and fix resulting issues

khmoran:
There was a bug in `index.ts` on dataserver leading to validation not actually taking place -- this fixes it


khmoran:

PR #568, opened 2020-07-15
Add the top-level properties to the case requests & responses in data service

khmoran:
These requests & responses are now validated against the `Case` object definition in the API YAML -- and these show up in the API docs

Note that the `create` request has some required properties that the responses don't have (since legacy data might not have these properties)

Next PRs will be to flesh out the sub-props of the `Case` object

api-docs

### DIFF
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -60,7 +60,13 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/NewCase"
+              allOf:
+                - $ref: "#/components/schemas/Case"
+              required:
+                - caseReference
+                - location
+                - events
+                - revisionMetadata
       parameters:
         - name: validate_only
           in: query
@@ -94,7 +100,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/NewCase"
+              $ref: "#/components/schemas/Case"
       responses:
         "200":
           $ref: "#/components/responses/200Case"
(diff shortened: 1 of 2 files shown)


khmoran:

data-serving/data-service/test/controllers/case.test.ts
```diff
@@ -138,11 +138,11 @@ describe('POST', () => {
...
             .expect(201);
     });
-    it('create with invalid input and validate_only should return 422', async () => {
+    it('create with input missing required properties and validate_only should return 400', async () => {
         return request(app)
             .post('/api/cases?validate_only=true')
             .send({})
-            .expect(422);
+            .expect(400);
     });
```
axmb, line 146: Is there a test that _has_ required parameters, but there's an invalid value, that returns 422?

data-serving/data-service/api/openapi.yaml
```diff
@@ -94,7 +100,7 @@ paths:
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/NewCase"
+              $ref: "#/components/schemas/Case"
```
axmb, line 103: Aren't the same values `required` for upsert?

data-serving/data-service/test/controllers/case.test.ts
khmoran: added one -- the distinction is that 400's come from the API validation, and 422's come from Mongoose validation

so added a test that (at least for now) we're only checking at the mongoose layer (age range)

data-serving/data-service/api/openapi.yaml
khmoran: AFAICT, from the tests, no -- since it doesn't require a full `case` object -- LMK if that's incorrect though and I can fix the tests

I did add the `caseReference` field as required though since I think that *is* required for upsert

data-serving/data-service/test/controllers/case.test.ts
axmb: Cool, SG. We should reflect on which error is raised when afterwards, but this is great for now.

data-serving/data-service/api/openapi.yaml
axmb: You're right on about that -- erroneous comment on my part. The mechanics are a little complicated:

1. Upsert looks for a case by the provided `cR.sourceId` and `cR.sourceEntryId`.
2. If it finds one, it performs an update, potentially partial, with the remainder of the request body.
3. If it doesn't find one, it creates an item with the provided request body; and in this case, `cR` must be present as you'd expect.

The result is that `cR` is effectively required, but depending on whether a case is found, other fields are semantically required (or not).

khmoran: makes sense, thank you!

PR #575, opened 2020-07-16
Add API docs for caseReference, demographics, and location props

khmoran:
Open to different names for `Case` vs. `NewCase` although I've seen this naming convention in OpenAPI examples. `NewCase` only modifies what's required, not the actual property defs.

image
image

### DIFF
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -223,10 +223,20 @@ components:
               type: string
             place:
               type: string
-            query:
+            name:
               type: string
             geoResolution: 
               type: string
+            geometry:
+              type: object
+              properties:
+                latitude:
+                  type: number
+                longitude:
+                  type: number
+              required:
+                - latitude
+                - longitude
         events:
           type: array
           items:
@@ -270,6 +280,8 @@ components:
         location:
           required:
             - country
+            - geoResolution
+            - geometry
       required:
         - caseReference
         - location
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -60,13 +60,7 @@ paths:
         content:
           application/json:
             schema:
-              allOf:
-                - $ref: "#/components/schemas/Case"
-              required:
-                - caseReference
-                - location
-                - events
-                - revisionMetadata
+              $ref: "#/components/schemas/NewCase"
       parameters:
         - name: validate_only
           in: query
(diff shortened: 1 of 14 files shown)


data-serving/data-service/api/openapi.yaml
```diff
@@ -221,6 +258,23 @@ components:
...
+      description: A "#/components/schemas/Case" with additional required fields for newly-created cases
+      allOf:
+        - $ref: "#/components/schemas/Case"
+      properties:
+        caseReference:
+          required:
+            - sourceId
+            - sourceUrl
+        location:
+          required:
```
axmb, line 271: This is another area in which there's a conditional requirement. If lat/long aren't provided, then `query` is required in order to geocode. So there's this sort of `required: $or: { $and: { lat, lng } }, query}` relationship going on.

WDYT is the best way to handle this?

khmoran:

data-serving/data-service/api/openapi.yaml
```diff
@@ -221,6 +258,23 @@ components:
...
+            - country
+      required:
+        - caseReference
+        - location
+        - events
```
axmb, line 276: I can't imagine there's a good way to encode requiring an event with a specific `name` field. Is there a description portion where it'd be easy to clarify this?

khmoran: Yeah fair, I will add it to the `description` for this property in the next PR when I flesh out the `events` prop

data-serving/data-service/api/openapi.yaml
khmoran: I think that applies to the curator service, not data service -- that said, 2 things:

1. When I start doing the API for the curator service, we'll run into this. My proposal is to add a request interceptor for create (and maybe upsert?) that runs the more complex validation logic that we can't express in the OpenAPI spec, and also document these requirements in the `description` field for the property in the OpenAPI spec.

2. You made me realize I did kinda mess up location here though -- s/query/name and missed `geometry` completely. I've updated the location spec

axmb: Oof -- completely glossed that this was for the data service, sorry. Thanks for the catch and SGTM.

data-serving/data-service/src/model/location.ts
```diff
@@ -1,37 +1,32 @@
...
-            type: String,
-            required: fieldRequiredValidator,
+            required: true,
         },
+        administrativeAreaLevel1: String,
+        administrativeAreaLevel2: String,
+        administrativeAreaLevel3: String,
         // Place represents a precise location, such as an establishment or POI.
         place: String,
         // A human-readable name of the location.
```
axmb, line 32: Do we rely on `name` to the extent that it should be required? It'll be present in all geocoded results, IIRC, but I'm not sure if there's a use case to not have it.

data-serving/data-service/test/model/data/location.minimal.json
```diff
@@ -1,4 +1,8 @@
 {
     "place": "Kings County Hospital Center",
```
axmb, line 2: Isn't this invalid for lack of country?

khmoran: oops, good catch, thanks! was missing test for minimal case

data-serving/data-service/src/model/location.ts
khmoran: yeah, good call, it should be -- added it

PR #576, opened 2020-07-16
Navigate to /cases after submitting the CaseForm

allysonjp715:
- After hitting submit the CaseForm is closed and the LinelistTable data is refreshed
 - Removes the add button on the LinelistTable
 - Shows the added/edited case row as highlighted in the LinelistTable
 - Shows success message above LinelistTable

Before adding a row:
Screen Shot 2020-07-16 at 3 54 35 PM

After adding a row:
Screen Shot 2020-07-16 at 3 53 47 PM

### DIFF
--- a/verification/curator-service/ui/src/components/LinelistTable.tsx
+++ b/verification/curator-service/ui/src/components/LinelistTable.tsx
@@ -72,7 +72,7 @@ interface LocationState {
     editedCaseIds: [string];
 }
 
-interface Props extends RouteComponentProps<any, any, LocationState> {
+interface Props extends RouteComponentProps<never, never, LocationState> {
     user: User;
 }
 
--- a/verification/curator-service/ui/src/components/LinelistTable.tsx
+++ b/verification/curator-service/ui/src/components/LinelistTable.tsx
@@ -1,8 +1,8 @@
 import { Case, Pathogen, Travel, TravelHistory } from './Case';
 import MaterialTable, { QueryResult } from 'material-table';
 import React, { RefObject } from 'react';
+import { RouteComponentProps, withRouter } from 'react-router-dom';
 
-import AddIcon from '@material-ui/icons/AddOutlined';
 import CaseForm from './CaseForm';
 import DeleteIcon from '@material-ui/icons/DeleteOutline';
 import EditCase from './EditCase';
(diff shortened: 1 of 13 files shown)


allysonjp715:

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -67,15 +67,18 @@ interface TableRow {
     admittedToHospital: string;
 }
 
-interface Props {
+interface LocationState {
+    newCaseIds: [string];
+    editedCaseIds: [string];
+}
+
+interface Props extends RouteComponentProps<any, any, LocationState> {
```
axmb, line 75: Do you get a warning in VSCode for these `any`s? What about if you run `npm run lint`?

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -67,15 +67,18 @@ interface TableRow {
     admittedToHospital: string;
 }
 
-interface Props {
+interface LocationState {
+    newCaseIds: [string];
```
axmb, line 71: Intentional `[tuple]` in lieu of `array[]`?

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -89,6 +92,15 @@ export default class LinelistTable extends React.Component<
             pageSize: 50,
         };
         this.closeCaseForm = this.closeCaseForm.bind(this);
+        // history.location.state can be updated with newCaseIds, on which we
+        // must refresh the table
+        this.unlisten = this.props.history.listen((_, __) =>
+            this.tableRef.current?.onQueryChange(),
+        );
```
axmb, line 99: Neat solution!

verification/curator-service/ui/src/components/LinelistTable.tsx
allysonjp715: Yes, I thought we were okay with ignoring warnings for small things like that. But in quick fixes I saw an option that said "Use 'never' instead, this is useful when instantiating generic type parameters that you don't need to know the type of." so switched to that instead.

verification/curator-service/ui/src/components/LinelistTable.tsx
allysonjp715: Nope, thanks, changed.

PR #579, opened 2020-07-16
Document API objects for events, symptoms, pre-existing conditions, and transmissions properties

khmoran:
dateRange.start and dateRange.end should be strings with the date-time format, but apparently we call it with all sorts of things so leaving it pretty loosey-goosey for now


khmoran:

Issue #573, opened 2020-07-16
Bulk CSV upload in progress: Progress bar / UI template


sratcliffe118:
Feature: UI pattern to show that a CSV bulk upload is in progress, e.g. upload progress bar / spinner / time left to upload. 

Currently: There is no indication that anything has happened or is taking place which caused me to click multiple times and think the process was broken.

Where: On the bulk upload CSV screen once the file has been submitted

sratcliffe118:
Current UI has no indication as to state of the upload.

![Screen Shot 2020-07-16 at 5 57 10 PM](https://user-images.githubusercontent.com/63205311/87700159-cd7c1a00-c78d-11ea-9f5e-87d8ec7981e9.png)

shroothless:
![image](https://user-images.githubusercontent.com/67452108/87727691-6fe8cd00-c78f-11ea-89f5-1e244619f2c9.png)
 Can we do this?

axmb:
Yep! I implemented something similar earlier today. I'll submit that, and tweak/move that into position when I move the buttons.

axmb:
To update: I'm working through #571 and once done will swing back to bring the bulk upload UI to spec (including this specific issue).

PR #580, opened 2020-07-17
Add/use an updating file upload progress spinner in the bulk upload form

axmb:
Re: #573.

Wrote this earlier today, prior to the mock, but it should be quite simple to tweak/move this into position when we move the upload button(s). Including a bit of testing here, but since our server requests are real rather than stubbed in Cypress, it's a bit non-trivial.

spinner

Includes a small change to awaiting form submission results/handling `isSubmitted` in order to get the right behavior during upload.

### DIFF
--- a/verification/curator-service/ui/cypress/integration/components/BulkCaseForm.spec.ts
+++ b/verification/curator-service/ui/cypress/integration/components/BulkCaseForm.spec.ts
@@ -33,7 +33,7 @@ describe('Bulk upload form', function () {
         cy.wait('@upsertCase');
         cy.wait('@upsertCase');
         cy.contains('Success! Created 2 new rows.');
-        cy.get('[data-testid="progress"]').should('not.be.visible');
+        cy.get('[data-testid="progress"]').should('not.exist');
         cy.get('button[aria-label="close overlay"').click();
 
         // Common data.
@@ -67,7 +67,7 @@ describe('Bulk upload form', function () {
         cy.get('button[data-testid="submit"]').click();
         cy.wait('@upsertCases');
         cy.contains('Success! Created 2 new rows.');
-        cy.get('[data-testid="progress"]').should('not.be.visible');
+        cy.get('[data-testid="progress"]').should('not.exist');
         cy.get('button[aria-label="close overlay"').click();
 
         cy.visit('/cases');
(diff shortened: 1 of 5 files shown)


axmb:

verification/curator-service/ui/cypress/integration/components/BulkCaseForm.spec.ts
```diff
@@ -33,6 +33,7 @@ describe('Bulk upload form', function () {
         cy.wait('@upsertCase');
         cy.wait('@upsertCase');
         cy.contains('Success! Created 2 new rows.');
+        cy.get('[data-testid="progress"]').should('not.be.visible');
```
allysonjp715, line 36: Could you also check that it is visible at some point?

allysonjp715, line 36: Perhaps `not.exist` ?

verification/curator-service/ui/src/components/bulk-case-form-fields/FileUploadProgress.tsx
```diff
@@ -0,0 +1,33 @@
+import CircularProgress, {
+    CircularProgressProps,
+} from '@material-ui/core/CircularProgress';
+
+import Box from '@material-ui/core/Box';
+import React from 'react';
+import Typography from '@material-ui/core/Typography';
+
+export default function FileUploadProgress(
```
allysonjp715, line 9: Nit: this isn't specific to Files so we could name it just UploadProgress in case it's needed somewhere else in the future.

verification/curator-service/ui/cypress/integration/components/BulkCaseForm.spec.ts
axmb: Not in this test, since we're making real remote calls. But I'll add a case to the unit test for this!

axmb: I've typically used that, but thought that `visible` was a better description of the user behavior we were testing, here.

I can switch to `exist` for now, since it's used elsewhere. I'll take a look at which might be better to use throughout.

verification/curator-service/ui/src/components/bulk-case-form-fields/FileUploadProgress.tsx
axmb: SG, done.

PR #582, opened 2020-07-17
Update error message on age range fields

### LINKED ISSUE #459: how do we currently accommodate for example age >60
If data provided says >60. How do we enable that to be entered unambiguously. For now the only option I see is enter 60 - 120. But that's not ideal.

Option could be to allow to enter min age '>60' and autofill to max age (120)?

allysonjp715:
Fixes #459


verification/curator-service/ui/src/components/CaseForm.tsx
```diff
@@ -202,7 +202,9 @@ const NewCaseValidation = Yup.object().shape(
             .max(120, 'Age must be between 0 and 120')
```
khmoran, line 202: at some point maybe we want to make 0 and 120 consts since they're referred to several times in here (and probably in bulk upload too)

allysonjp715: Yes good point, added TODO.

PR #584, opened 2020-07-17
Add a batch validate API to the data service

axmb:


data-serving/data-service/api/openapi.yaml
```diff
@@ -117,6 +117,22 @@ paths:
           $ref: "#/components/responses/422"
         "500":
           $ref: "#/components/responses/500"
+  /cases/batchValidate:
```
khmoran, line 120: will we be able to remove the validate_only param on the other endpoint once we're using this?

axmb: Yep, should -- the only current user will be migrating to the batch API.

axmb:
Retroactively taggin #584

axmb:

PR #585, opened 2020-07-17
Finish fleshing out the OpenAPI case request/response object in data service

khmoran:


khmoran:

data-serving/data-service/api/openapi.yaml
```diff
@@ -298,25 +262,124 @@ components:
...
+      properties:
+        curator:
+          type: string
+        date:
+          $ref: "#/components/schemas/Date"
+        notes:
+          type: string
+    Location:
+      type: object
+      properties:
```
axmb, line 354: Should more properties (here and elsewhere) be labeled as required? E.g., location.country, metadata.curator.

khmoran: believe it or not, some of the imported data doesn't have even that. so those fields are required in NewCase, but not in Case (except metadata.curator -- trying to work through that now...)

PR #593, opened 2020-07-20
Set revision metadata on data server

khmoran:
Major points:

- Any create/update/upsert requests will need to provide a new `curator` param in the request, with the email address of the currently logged-in user
- The create/update/upsert routes have a new middleware function that sets the correct revision metadata (and removes the extra `curator` request param); the handlers downstream stay the same


data-serving/data-service/src/controllers/preprocessor.ts
```diff
@@ -0,0 +1,82 @@
...
+    };
+};
+
+const createUpdateMetadata = (
+    c: CaseDocument,
+    curatorEmail: string,
+    updateNotes: string,
+) => {
+    return {
+        revisionNumber: ++c.revisionMetadata.revisionNumber,
```
axmb, line 23: For some reason I had in my mind that this was a property of `updateMetadata` (which itself was a repeated field). But that level of audit detail probably isn't necessary.

data-serving/data-service/src/controllers/preprocessor.ts
```diff
@@ -0,0 +1,82 @@
...
+        });
+    }
+
+    // Set the correct, server-generated revisionMetadata for subsequent
+    // processors to use.
+    const revisionMetadata = c
+        ? createUpdateMetadata(
+              c,
+              curatorEmail,
+              request.body?.revisionMetadata?.updateMetadata?.notes,
```
axmb, line 68: Hm -- so we're now supplying a separate field for `curatorEmail`, but if there are curator notes, we're still expecting them to be populated in `revisionMetadata`? It's probably clearer _not_ to have `revisionMetadata` in the request type, here. Otherwise there are multiple places one might expect curator email to be drawn from.

khmoran:
[Dataserver should populate (most) version data](https://trello.com/c/X0icPhIz/263-dataserver-should-populate-most-version-data)

data-serving/data-service/src/controllers/preprocessor.ts
khmoran: `updateMetadata` isn't repeated -- might be strange, but the `revisionMetadata` field isn't meant to capture the full edit log, only the first edit (i.e. `creationMetadata`) and most recent (`updateMetadata`), so that we can display these easily in the UI. (You can imagine wanting to see who initially curated the case and when it was most recently updated & why.)

The full log in theory would be available through the versioning pattern which has not yet been implemented :)

Does that make sense?

khmoran:
> Approach looks good, for your note I don't think you can call next with the data you want to avoid multiple reads, the pattern here is to out what you fetched in the request itself and get it from it in the next handler.

Got it. For now I think I'll just leave as is to keep things self-contained -- does you feel OK with doing double reads?

data-serving/data-service/src/controllers/preprocessor.ts
khmoran: I knowww it's horrible.

We're not populating these notes so I can drop this piece of the PR until we figure out a better pattern, but see my comment above for why it is the way it is. LMK what you think.

attwad:
> > Approach looks good, for your note I don't think you can call next with the data you want to avoid multiple reads, the pattern here is to out what you fetched in the request itself and get it from it in the next handler.
> 
> Got it. For now I think I'll just leave as is to keep things self-contained -- do you feel OK with doing double reads?

feeling uncomfortably excited, yes. 👍

data-serving/data-service/src/controllers/preprocessor.ts
axmb: Totally; mostly just musing. That was approximately my understanding.

data-serving/data-service/src/controllers/preprocessor.ts
```diff
@@ -0,0 +1,82 @@
...
+        : createNewMetadata(
+              curatorEmail,
+              request.body?.revisionMetadata?.creationMetadata?.notes,
+          );
+    request.body.revisionMetadata = revisionMetadata;
+
+    // Clean up the additional metadata.
+    delete request.body.curator;
+
+    // NTS: Can you call `next` with the case and save ourselves from having to
```
axmb: Should this be removed?

data-serving/data-service/src/controllers/preprocessor.ts
axmb: Dropping the notes is probably cool for now (don't think it has/will imminently come up in testing), but I think we ultimately want to have the field. Would it be workable just to put `notes` into `curator`, so that there's no expectation of `revisionMetadata` fields being necessary for a write? Sorry if I've mis-grok'd the comments above.

data-serving/data-service/test/controllers/preprocessor.test.ts
```diff
@@ -0,0 +1,221 @@
...
+    return Case.deleteMany({});
+});
+
+afterAll(() => {
+    dateNowSpy.mockRestore();
+    return mongoServer.stop();
+});
+
+describe('create', () => {
+    it('create sets create metadata', async () => {
```
axmb, line 31: Super great tests, one hyper-nit: don't need to repeat expressions from the `describe` block in `it` blocks (I think it's preferred for them to [read fluently](https://jestjs.io/docs/en/api#describename-fn)).

data-serving/data-service/src/controllers/preprocessor.ts
```diff
@@ -0,0 +1,70 @@
...
+    };
+};
+
+export const setRevisionMetadata = async (
+    request: Request,
+    response: Response,
+    next: NextFunction,
+) => {
+    const curatorEmail = request.body.curator.email;
+    const caseReference = request.body?.caseReference;
```
attwad, line 36: why body? here and not above?

data-serving/data-service/test/controllers/preprocessor.test.ts
```diff
@@ -0,0 +1,221 @@
...
+        const requestBody = {
+            ...minimalCase,
+            curator: { email: 'creator@gmail.com' },
+        };
+        await setRevisionMetadata(
+            { body: requestBody, method: 'POST' } as Request,
+            {} as Response,
+            /* next= */ jest.fn,
```
attwad, line 39: nit, you could expect that this mock fn got called, up to you.

data-serving/data-service/src/controllers/preprocessor.ts
khmoran: 👍

data-serving/data-service/src/controllers/preprocessor.ts
khmoran: I think that make sense. I've added a Trello card to circle back to these revision notes, but I think your suggestion is a good one. I think I need to understand where in the UX this will be surfaced (if anywhere).

data-serving/data-service/src/controllers/preprocessor.ts
khmoran: just terrible inconsistency

it should always have a body at this point; otherwise openAPI would have rejected it

data-serving/data-service/src/controllers/preprocessor.ts
khmoran: I think maybe you were looking at an old snapshot of this file -- LMK if you want to re-review (though it's pretty much the same, sans notes)

data-serving/data-service/test/controllers/preprocessor.test.ts
khmoran: oops totally right, thanks!

data-serving/data-service/test/controllers/preprocessor.test.ts
khmoran: good call! done

Issue #571, opened 2020-07-16
Bulk CSV upload slow for files with many rows


sratcliffe118:
Attempt to bulk upload file CSV and and seeing this error

<img width="547" alt="Screen Shot 2020-07-16 at 5 07 46 PM" src="https://user-images.githubusercontent.com/63205311/87695068-ee8d3c80-c786-11ea-859c-8886d785110c.png">

axmb:
It's working, but it's working rather slowly. I think there are two issues to address, here:

1. There needs to be a progress indicator (#573).
2. Upload of large files (the one in question was ~40k rows) _should_ be faster. I'll repurpose this issue for that.

axmb:
Retroactively tagging #584, which is the first change to speed this up.

PR #594, opened 2020-07-20
Add a batchUpsert endpoint in the curator service case API.

axmb:
Re: #571 

This was a little tricky, due to geocoding happening on the curator service, but I think this is the best balance of clarity and efficiency. It might be somewhat better if we moved geocoding to the data service, but that's a separate thread. For now, orchestrating this work in the curator service API, taking advantage of the batchValidate endpoint I'd added to the data service API.

The plan is to migrate the bulk upload UI to use this endpoint, and likely to use it from parsing functions as well. I think this should more or less fix the aforementioned issue (since the bulk of the slowdown seemed to be network latency between clients/our server), but there are a couple TODOs with improvements should we need them.

### DIFF
--- a/verification/curator-service/api/src/controllers/cases.ts
+++ b/verification/curator-service/api/src/controllers/cases.ts
@@ -151,6 +151,8 @@ export default class CasesController {
             }
 
             // 3. Upsert each case.
+            // Consider adding a batchUpsert endpoint on the data service if
+            // this slows us down too much.
             for (let index = 0; index < req.body.cases.length; index++) {
                 const c = req.body.cases[index];
                 await axios.put(this.dataServerURL + '/api/cases', c);
--- a/verification/curator-service/api/src/controllers/cases.ts
+++ b/verification/curator-service/api/src/controllers/cases.ts
@@ -107,6 +107,71 @@ export default class CasesController {
         }
     };
 
+    /**
+     * Validates and upserts the provided cases.
+     *
+     * Executes three operations on the supplied data:
+     *
+     *   1. Geocodes provided cases as required. If there are any issues
+     *      geocoding (which is done serially -- though we should consider
+     *      taking advantage of a batch geocode API), return the results now
+     *      without proceeding to the validation stage.
+     *   2. Performs validation of all provided cases via the data service
+     *      batchValidate API. If any validation issues are found, return the
+     *      results now without proceeding to the upsert stage.
+     *   3. Upserts the data via the data service upsert API.
+     */
+    batchUpsert = async (req: Request, res: Response): Promise<void> => {
+        try {
+            // 1. Geocode each case.
+            const geocodeErrors = await this.batchGeocode(req);
+            if (geocodeErrors.length > 0) {
+                res.status(207).send({
+                    phase: 'GEOCODE',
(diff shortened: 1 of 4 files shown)


axmb:

verification/curator-service/api/src/controllers/cases.ts
```diff
@@ -107,6 +107,73 @@ export default class CasesController {
         }
     };
 
+    /**
+     * Validates and upserts the provided cases.
+     *
+     * Executes three operations on the supplied data:
+     *
+     *   1. Geocodes provided cases as required. If there are any issues
+     *      geocoding (which is done serially -- though we should consider
```
attwad, line 116: FYI in case you didn't know we're using the endpoint that supports batch geocoding with mapbox (https://docs.mapbox.com/api/search/#batch-geocoding), we could rely on it to geocode. Their docs do make it clear what are the limits though... I assume the queries must fit in a URL, if so that limits seems pretty low...

verification/curator-service/api/src/controllers/cases.ts
```diff
@@ -107,6 +107,73 @@ export default class CasesController {
...
+     *      taking advantage of a batch geocode API), return the results now
+     *      without proceeding to the validation stage.
+     *   2. Performs validation of all provided cases via the data service
```
attwad, line 119: shouldn't we assume geocode is going to be valid and validate the rest before geocoding? validation is free, geocoding is a paid and much slower service.

verification/curator-service/api/src/controllers/cases.ts
axmb: Yeah, I wasn't sure how to read that:

> Batch geocoding is only available using the **mapbox.places-permanent** endpoint. If you are interested in using **mapbox.places-permanent** to perform batch geocodes, contact Mapbox sales.

Is that a typo, or are they saying that it's available, but you should contact sales first? I wasn't clear.

attwad: you have to contact sales to enable the endpoint, I think batch comes with it, we are already whitelisted for that endpoint so batch should be available to us.

verification/curator-service/api/src/controllers/cases.ts
axmb: Considered this for the reasons you listed. Also considered failing the geocoding stage fast (stopping on the first error, rather than doing all cases) for similar reasons.

I'm sort of on the fence for this and could go either way. I think our caching is likely to have a high hit rate for bulk geocoding, which abates the biggest downsides. The downside to validating first is that it requires setting dummy location values, and so makes this marginally more complex.

WDYT?

attwad: I also hope the LRU cache does its job well enough so it's probably not worth the added complexity of dummy location values, fine as is.

verification/curator-service/api/src/controllers/cases.ts
axmb: Got it, SG. One thing that's occurred to me is that we might not want to do this for the sake of caching requests (since each geocode in the batch is billed as a request).

Issue #426, opened 2020-07-02
Location field does not support county-level lookup in the US.


sratcliffe118:
Location field does not support county-level lookup for the US.
This is a standard expectation to be able to enter the county (in the US) in which a case is occurring.
“In the United States, having county level data is critical”

attwad:
Still waiting Jared's for this, he sent an email to their mapbox contact and we're awaiting a reply.

JaredHawkins:
Hey all!  I haven't heard back from Mapbox.  I have a follow-up email set to go out this afternoon if I still haven't heard back.

sratcliffe118:
Hi Jared, any update here, this is a key blocker for our first milestone

JaredHawkins:
They said: "Sounds like we do not currently have counties as a returnable data type in the US. If you have a counties layer then you could do a spatial join to extract that information for your current dataset."  I know that Mapbox has county layer(s), and followed up asking if we can speak to an engineer on their recommended work around - but I haven't heard back yet.

JaredHawkins:
Mapbox says:

> If you are using the points returned from the geocoder AND you have a county tileset then I would recommend that you try using our Tilequery API. You can pass the lat/long coordinates to our Tilequery API to retrieve the county name.
> 
> Tilequery information
> 
> - Documentation: https://docs.mapbox.com/api/maps/#tilequery
> - Tutorials:
> 
> - https://docs.mapbox.com/help/tutorials/create-a-timezone-finder-with-mapbox-tilequery-api/
> 
> - https://docs.mapbox.com/help/tutorials/tilequery-healthy-food-finder/#add-the-tilequery-api
> 
> Hope this helps!

I've asked them if there is an existing Mapbox tileset for county (I believe there is), and am awaiting more info.  They have confirmed that county is not returned for US, and I'm also waiting on some more clarity as to why that is the case.

JaredHawkins:
OK, I just had a nice talk with Mapbox.  This sums it up:

[...]

PR #596, opened 2020-07-21
Make location fields editable

attwad:
This enables curators to manually add missing admin levels for example.

Temporary unblock #426.

This PR follows #595.

image


attwad:
> Could you add a test for changing the fields?

ah yeah definitely, done.

PR #601, opened 2020-07-21
Navigate to /cases after successful bulk upload

allysonjp715:
Highlights all added and updated row. Shows success message on /cases.
bulkgif

Updated message style after taking that screen recording:
Screen Shot 2020-07-21 at 1 39 36 PM


allysonjp715:

PR #608, opened 2020-07-23
Add admins fetcher using mapbox boundaries API

attwad:
This is step 1/2: it gets the admins but not their name so not filling this in just yet in the geocode results.
Getting the names is going to be slightly complex due to the way mapbox gives it to us (not via an API but via static files (!)) so I'll do that in a next PR.

For #426

### DIFF
--- a/verification/curator-service/api/src/geocoding/adminsFetcher.ts
+++ b/verification/curator-service/api/src/geocoding/adminsFetcher.ts
@@ -40,9 +40,8 @@ export default class MapboxAdminsFetcher {
             geocode.geometry.latitude
         }.json?access_token=${this.accessToken}`;
         try {
-            const resp = await axios.get(url);
-            const features = (resp.data as BoundariesResponse).features;
-            features.forEach((feature) => {
+            const resp = await axios.get<BoundariesResponse>(url);
+            resp.data.features.forEach((feature) => {
                 switch (feature.properties.tilequery.layer) {
                     case 'boundaries_admin_1':
                     // TODO: Get name from prop.id and put in geocode.administrativeAreaLevel1;
--- a/verification/curator-service/api/src/geocoding/adminsFetcher.ts
+++ b/verification/curator-service/api/src/geocoding/adminsFetcher.ts
@@ -1,0 +1,95 @@
+import { GeocodeResult, Resolution } from './geocoder';
+
+import axios from 'axios';
+
+/**
+ * Mapbox administrative area fetcher using the boundaries API.
+ * https://www.mapbox.com/boundaries/.
+ *
+ * Some geocoding results do not contain the full administrative area details,
+ * such details can be subsequently fetched using the boundaries API by doing a
+ * tilequery fetch of the center of the geocode result with the administrative
+ * boundaries layers.
+ * Cf. https://docs.mapbox.com/help/glossary/tilequery-api/
+ *
+ * TODO: Explain the custom local fetches of names once implemented.
+ */
+export default class MapboxAdminsFetcher {
+    constructor(private readonly accessToken: string) {}
+
+    // Fill in missing admin levels for the given GeocodeResult.
+    async fillAdmins(geocode: GeocodeResult): Promise<void> {
+        // Get missing admins from geocode result.
+        const missingAdmins = new Set<Resolution>();
+        if (geocode.administrativeAreaLevel1 === '') {
+            missingAdmins.add(Resolution.Admin1);
+        }
+        if (geocode.administrativeAreaLevel2 === '') {
+            missingAdmins.add(Resolution.Admin2);
+        }
+        if (geocode.administrativeAreaLevel3 === '') {
+            missingAdmins.add(Resolution.Admin3);
+        }
+        if (missingAdmins.size === 0) {
+            return;
+        }
+        // Fetch all missing admins in one query.
+        const url = `https://api.mapbox.com/v4/${getLayersParam(
+            missingAdmins,
+        )}/tilequery/${geocode.geometry.longitude},${
+            geocode.geometry.latitude
+        }.json?access_token=${this.accessToken}`;
+        try {
+            const resp = await axios.get(url);
+            const features = (resp.data as BoundariesResponse).features;
+            features.forEach((feature) => {
+                switch (feature.properties.tilequery.layer) {
+                    case 'boundaries_admin_1':
+                    // TODO: Get name from prop.id and put in geocode.administrativeAreaLevel1;
+                    case 'boundaries_admin_2':
+                    // TODO: Get name from prop.id and put in geocode.administrativeAreaLevel2;
+                    case 'boundaries_admin_3':
+                    // TODO: Get name from prop.id and put in geocode.administrativeAreaLevel3;
+                }
+            });
+        } catch (e) {
+            // Fail gracefully, not being able to fetch all admins isn't a huge deal.
+            console.error(`Retrieving admins from url: ${url}:`, e);
+            return;
+        }
+    }
+}
+
+// Mapbox boundaries types definitions, not part of the mapbox SDK.
+interface BoundariesResponse {
+    features: Feature[];
+}
+
+interface Feature {
+    properties: Properties;
+}
+
+interface Properties {
+    id: string;
+    tilequery: Tilequery;
+}
+
+interface Tilequery {
+    layer: string;
+}
+
+// Map of admin resolutions to their URL params counterpart in the mapbox boundaries API.
+const adminToLayer = new Map<Resolution, string>([
+    [Resolution.Admin1, 'mapbox.enterprise-boundaries-a1-v2'],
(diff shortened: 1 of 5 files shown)


verification/curator-service/api/src/geocoding/adminsFetcher.ts
```diff
@@ -0,0 +1,95 @@
...
+
+import axios from 'axios';
+
+/**
+ * Mapbox administrative area fetcher using the boundaries API.
+ * https://www.mapbox.com/boundaries/.
+ *
+ * Some geocoding results do not contain the full administrative area details,
+ * such details can be subsequently fetched using the boundaries API by doing a
+ * tilequery fetch of the center of the geocode result with the administrative
```
axmb, line 11: > of the center of the geocode result

Can it be looked up by name, too? If our geocode result for "Adams, Ohio, US" is just the center of Ohio, that's unlikely to give us the right county.

verification/curator-service/api/src/geocoding/adminsFetcher.ts
```diff
@@ -0,0 +1,95 @@
...
+        }
+        // Fetch all missing admins in one query.
+        const url = `https://api.mapbox.com/v4/${getLayersParam(
+            missingAdmins,
+        )}/tilequery/${geocode.geometry.longitude},${
+            geocode.geometry.latitude
+        }.json?access_token=${this.accessToken}`;
+        try {
+            const resp = await axios.get(url);
+            const features = (resp.data as BoundariesResponse).features;
```
axmb, line 44: `axios.get` is parameterized to work with, e.g., `get<BoundariesResponse>`.

verification/curator-service/api/src/geocoding/adminsFetcher.ts
```diff
@@ -0,0 +1,95 @@
...
+                    case 'boundaries_admin_1':
+                    // TODO: Get name from prop.id and put in geocode.administrativeAreaLevel1;
+                    case 'boundaries_admin_2':
+                    // TODO: Get name from prop.id and put in geocode.administrativeAreaLevel2;
+                    case 'boundaries_admin_3':
+                    // TODO: Get name from prop.id and put in geocode.administrativeAreaLevel3;
+                }
+            });
+        } catch (e) {
+            // Fail gracefully, not being able to fetch all admins isn't a huge deal.
```
axmb, line 56: > isn't a huge deal.

I'm not totally sure. Do we want case entry to fail (e.g. for bulk upload) if "Adams, Ohio" doesn't find an admin2?

verification/curator-service/api/src/geocoding/adminsFetcher.ts
```diff
@@ -0,0 +1,95 @@
...
+            console.error(`Retrieving admins from url: ${url}:`, e);
+            return;
+        }
+    }
+}
+
+// Mapbox boundaries types definitions, not part of the mapbox SDK.
```
axmb, line 63: PR time :muscle:

verification/curator-service/api/src/geocoding/adminsFetcher.ts
```diff
@@ -0,0 +1,95 @@
...
+// Map of admin resolutions to their URL params counterpart in the mapbox boundaries API.
+const adminToLayer = new Map<Resolution, string>([
+    [Resolution.Admin1, 'mapbox.enterprise-boundaries-a1-v2'],
+    [Resolution.Admin2, 'mapbox.enterprise-boundaries-a2-v2'],
+    [Resolution.Admin3, 'mapbox.enterprise-boundaries-a3-v2'],
+]);
+
+// getLayersParams return the layers query param used by the mapbox boundaries API based on missing admin levels.
+function getLayersParam(admins: Set<Resolution>): string {
+    const layers = [];
```
axmb, line 90: Totally optional, but could `return admins.map(adminToLayer.get(admin)).sort().join(',');`

verification/curator-service/api/test/geocoding/adminsFetcher.test.ts
```diff
@@ -0,0 +1,84 @@
...
+            country: 'Brasilistan',
+            geoResolution: Resolution.Admin3,
+            geometry: {
+                latitude: 12.34,
+                longitude: 45.67,
+            },
+            name: 'the',
+            place: 'to be',
+        });
+        expect(mockedAxios.get).toHaveBeenCalledTimes(0);
```
axmb, line 30: hyper-nit: `.not.toHaveBeenCalled()`

verification/curator-service/api/test/geocoding/adminsFetcher.test.ts
```diff
@@ -0,0 +1,84 @@
...
+            country: 'Brasilistan',
+            geoResolution: Resolution.Country,
+            geometry: {
+                latitude: 12.34,
+                longitude: 45.67,
+            },
+            name: 'the',
+            place: 'to be',
+        });
+        expect(mockedAxios.get).toHaveBeenCalledWith(
```
axmb, line 80: Should we check here that the provided object has its fields filled, too?

verification/curator-service/api/src/geocoding/adminsFetcher.ts
attwad: no the tilequery api that boundaries is based on only works with lat lng as input.

verification/curator-service/api/src/geocoding/adminsFetcher.ts
attwad: For manual entry it's definitely optional (as one can still enter it in the UI if not found), if we want to require it we should also make it mandatory in the schema but honestly I doubt we can find ALL admin levels EVERYWHERE in the world that seems like a pretty strong statement... I'd feel better getting 99% of them and not making it mandatory, perhaps there can be a manual process to fix those missing afterwards.

For bulk upload we might want to fail early and have the users resubmit the csv, I am not sure, either way that sounds like a request param which we can add in a later PR to make this mandatory if we decide to, sg?

verification/curator-service/api/src/geocoding/adminsFetcher.ts
attwad: haha yeah, maybe... if there were official type defs, right now it's on a third party package and could break anytime so my willingness is relatively low... especially for an API that's not well documented to start with (because it's private and you have to contact mapbox sales to use it...)

verification/curator-service/api/src/geocoding/adminsFetcher.ts
attwad: that's what I tried, but Set() has no map() :(

verification/curator-service/api/test/geocoding/adminsFetcher.test.ts
attwad: done

verification/curator-service/api/src/geocoding/adminsFetcher.ts
axmb: Yep, SG!

verification/curator-service/api/src/geocoding/adminsFetcher.ts
axmb: Then does this address geocoding "Miami-Dade [County], Florida"? Or do we have a :chicken: <-> :egg: situation?

verification/curator-service/api/test/geocoding/adminsFetcher.test.ts
axmb: Ah, of course, thanks.

verification/curator-service/api/src/geocoding/adminsFetcher.ts
axmb: Ah. Could `Array.from([...]`, but less worth.

verification/curator-service/api/src/geocoding/adminsFetcher.ts
attwad: For the "no results case" (I suppose Miami-dade is such an example?) I added the ability to just provide a geocode ourselves manually.

verification/curator-service/api/src/geocoding/adminsFetcher.ts
attwad: I keep forgetting about that, thanks :)

verification/curator-service/api/src/geocoding/adminsFetcher.ts
axmb: OK, I think we'll have to consider this more (feel free to submit this PR).

There's lots of US data that's organized by county -- so the data contains admin2 (county), admin1 (state), and country. For bulk upload, IIUC, this basically means the only solution to getting admin2 properly into the data is to somehow manually geocode the raw CSV data and supply the lat/long directly as columns. Or, we could just live with bad lat/long and "force" the admin2 name, as I think you're describing for manual upload.

PR #612, opened 2020-07-23
Modify source UX

### LINKED ISSUE #531: Update UX of source entry
Adding a new source flow. Once a URL has been added in the bulk upload or new line list entry screen and select + a new screen appears. Current flow is confusing in 2 areas:

1. The URL field goes into the name field, the 2 should be separate and not run into each other
2. The cursor should automatically move to name when the dialogue pops up so the name can be entered. Currently the cursor remains in the URL field when dialogue opens, this is confusing and results in additions being made to the url by mistake.

Attached screenshot 
Screen Shot 2020-07-13 at 1 23 18 PM

allysonjp715:
Fixes #531 

- Makes source name a field that pops up if the user types in a new source URL
- Saves the source just before submitting the case instead of on dialog close.
- Works with single case and bulk

Existing source:
Screen Shot 2020-07-23 at 1 03 17 PM

New source:
Screen Shot 2020-07-23 at 1 03 05 PM


verification/curator-service/ui/src/components/new-case-form-fields/CaseFormValues.tsx
```diff
@@ -36,6 +36,10 @@ export default interface CaseFormValues {
     notes: string;
 }
 
+export interface CaseReferenceForm extends CaseReference {
```
axmb, line 39: Should we define/export this in the Source component, rather than creating this interface in both forms?

verification/curator-service/ui/src/components/common-form-fields/Source.tsx
```diff
@@ -194,41 +171,28 @@ export function SourcesAutocomplete(
...
 
-                    if (params.inputValue !== '') {
+                    if (
+                        params.inputValue !== '' &&
+                        !filtered.find(
+                            (caseRef) =>
+                                caseRef.sourceUrl === params.inputValue,
+                        )
+                    ) {
                         filtered.push({
```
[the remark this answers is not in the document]
allysonjp715: We implemented this in the other fields that are freeSolo after talking with Shroo. We thought it was better UX to show the option in the dropdown, so they know it is possible to type their own answer in.

verification/curator-service/ui/src/components/new-case-form-fields/CaseFormValues.tsx
allysonjp715: Good idea, done.

PR #623, opened 2020-07-24
Add Case, NewCase request/reponse defs to curator service

khmoran:


khmoran:
[S#7: Design & implement 'Case' API entity in curator service](https://trello.com/c/gnnb3fDQ/290-s7-design-implement-case-api-entity-in-curator-service)

verification/curator-service/api/openapi/openapi.yaml
```diff
@@ -500,11 +503,154 @@ components:
...
+                type: string
+            occupation:
+              type: string
+        location:
+          $ref: "#/components/schemas/Location"
+        events:
+          description: > 
+            An event with name "confirmed" must be specified. Other event names
+            include onsetSymptoms, firstClinicalConsultation, selfIsolation,
+            hospitalAdmission, icuAdmission, and outcome.
```
axmb, line 549: Not a change to make now, but if we're dictating the title of events, I wonder if the `name` should be an enum. Or for lack of that, if this is the best place to document the "standard" value set.

verification/curator-service/api/openapi/openapi.yaml
```diff
@@ -522,9 +689,27 @@ components:
...
+        newly-created cases.
+      allOf:
+        - $ref: "#/components/schemas/Case"
+      properties:
+        caseReference:
+          required:
+            - sourceId
+            - sourceUrl
+        location:
+          description: Note that either `geometry` or `query` is required.
```
axmb, line 703: Isn't it `query || (geometry && country && geoResolution)`?

verification/curator-service/api/openapi/openapi.yaml
khmoran: yeah enum probably makes sense -- add a todo

verification/curator-service/api/openapi/openapi.yaml
khmoran: yes, thanks!

Issue #607, opened 2020-07-22
Rename org and repo


attwad:
We need to be Global.Health not open-covid-data and we need a repo name that is not temporary.

https://docs.github.com/en/github/setting-up-and-managing-organizations-and-teams/renaming-an-organization

https://docs.github.com/en/github/administering-a-repository/renaming-a-repository

Happy to volunteer and break everyone once we have a repo name and are ready to go.

dakrownie:
@attwad Who needs to make the decision on this?

attwad:
Long term owners (so not us), perhaps Jared, Moritz and Spencer?

dakrownie:
@JaredHawkins Please make the call or direct this to someone who can.

JaredHawkins:
Moritz owns this repo, though perhaps we should have it transferred over to BCH as part of the handoff?

dakrownie:
@mougk over to you.

outbreakprepared:
did we make a final call here ?

attwad:
as per slack poll, it will be `globaldothealth/list`, will prepare the changes now.

dakrownie:
Thanks @attwad!

PR #624, opened 2020-07-27
rename repo and org

attwad:
DO NOT SUBMIT until the rename happens.

Created this PR with VSCode search and replace on all files in repo, hopefully it didn't miss anything.

For #607


attwad:
FYI test failure is expected as we're fetching stuff from the repo in integration tests.

PR #628, opened 2020-07-27
add title to edit page as per mocks

attwad:
Add
image

We don't have a link to the data entry guide yet so didn't add that.

### DIFF
--- a/verification/curator-service/ui/src/components/CaseForm.tsx
+++ b/verification/curator-service/ui/src/components/CaseForm.tsx
@@ -1,6 +1,6 @@
 import * as Yup from 'yup';
 
-import { Button, LinearProgress } from '@material-ui/core';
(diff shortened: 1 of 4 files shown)


verification/curator-service/ui/src/components/CaseForm.tsx
```diff
@@ -796,6 +796,13 @@ class CaseForm extends React.Component<Props, CaseFormState> {
                                 </div>
                             </nav>
                             <div className={classes.form}>
+                                <Typography variant="h4">
+                                    Enter the details for a new case
```
allysonjp715, line 800: We should change this text (the new case part) depending on if it's the new case or edit case flow.

attwad: sounds good, done.

PR #641, opened 2020-07-27
add left side links to navigate in case details

### LINKED ISSUE #429: Implement the case details navigation floating panel
image

attwad:
Feedback from UX session was that the grey titles didn't appear like links, so I made those tentatively just look like links.

Also using plain old html inner links instead of the third party library used in the case for to link to internal parts of the page, I am not sure why was that needed? Happy to move to that one and/or change the form one to be similar once we agree on the way to do this.

image

Fixes #429

### DIFF
--- a/verification/curator-service/ui/src/components/ViewCase.tsx
+++ b/verification/curator-service/ui/src/components/ViewCase.tsx
@@ -110,7 +110,7 @@ const useStyles = makeStyles((theme) => ({
         marginBottom: '1em',
         marginLeft: '10em',
         // width - left shift and nav menu size.
-        width: 'calc(100vw - 25em)',
+        width: 'calc(100vw - 23em)',
         scrollBehavior: 'smooth',
     },
     editBtn: {
--- a/verification/curator-service/ui/src/components/ViewCase.tsx
+++ b/verification/curator-service/ui/src/components/ViewCase.tsx
@@ -1,3 +1,4 @@
+/* eslint-disable jsx-a11y/anchor-is-valid */
 import { Case, GenomeSequence, Location, Travel } from './Case';
 import {
     Container,
@@ -8,6 +9,7 @@ import {
     Button,
 } from '@material-ui/core';
 
+import { Link as TypoLink } from '@material-ui/core';
 import AppModal from './AppModal';
 import MuiAlert from '@material-ui/lab/Alert';
 import React from 'react';
@@ -107,9 +109,17 @@ const useStyles = makeStyles((theme) => ({
     container: {
         marginTop: '1em',
         marginBottom: '1em',
+        marginLeft: '10em',
+        // width - left shift and nav menu size.
+        width: 'calc(100vw - 25em)',
     },
     editBtn: {
         marginLeft: '1em',
+    },
+    navMenu: {
+        position: 'fixed',
+        lineHeight: '2em',
+        width: '13em',
     },
 }));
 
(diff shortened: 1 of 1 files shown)


verification/curator-service/ui/src/components/ViewCase.tsx
```diff
@@ -134,344 +145,420 @@ function dateRange(range?: { start?: string; end?: string }): string {
...
+                    SYMPTOMS
+                </TypoLink>
+                <br />
+                <TypoLink variant="body2" href="#travel-history">
+                    TRAVEL HISTORY
+                </TypoLink>
+                <br />
+                <TypoLink variant="body2" href="#pathogens">
+                    PATHOGENS
+                </TypoLink>
```
allysonjp715, line 176: A couple sections are missing. Here are the ones from Edit case:
Screen Shot 2020-07-27 at 2 00 23 PM

verification/curator-service/ui/src/components/ViewCase.tsx
```diff
@@ -134,344 +145,420 @@ function dateRange(range?: { start?: string; end?: string }): string {
...
-                    >
-                        <Button
-                            variant="outlined"
-                            color="primary"
-                            className={classes.editBtn}
-                            endIcon={<EditIcon />}
+        <>
+            <nav className={classes.navMenu}>
+                <TypoLink variant="body2" href="#case-data">
+                    CASE DATA
```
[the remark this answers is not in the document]
attwad: used css instead, should be okay with translations as well.

verification/curator-service/ui/src/components/ViewCase.tsx
attwad: as discussed, made it consistent with the mocks.

attwad:
> The scroll package animates the scroll nicely. Does this method do that?

no, I tried adding scrollBehavior: "smooth" but it seems react router is messing with links and with react you HAVE to use some kind of library like react-scroll which is a shame :( so switched to it...

verification/curator-service/ui/src/components/ViewCase.tsx
```diff
@@ -107,10 +110,18 @@ const useStyles = makeStyles((theme) => ({
     container: {
         marginTop: '1em',
         marginBottom: '1em',
+        marginLeft: '10em',
+        // width - left shift and nav menu size.
+        width: 'calc(100vw - 25em)',
```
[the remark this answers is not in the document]
attwad: ah sorry I had 2em left, 25 != 10+13, 23 =10+13

verification/curator-service/ui/src/components/ViewCase.tsx
```diff
@@ -107,10 +108,20 @@ const useStyles = makeStyles((theme) => ({
     container: {
         marginTop: '1em',
         marginBottom: '1em',
+        marginLeft: '10em',
+        // width - left shift and nav menu size.
+        width: 'calc(100vw - 23em)',
+        scrollBehavior: 'smooth',
```
allysonjp715, line 114: Is this still needed now that we're using the library?

verification/curator-service/ui/src/components/ViewCase.tsx
```diff
@@ -133,345 +144,485 @@ function dateRange(range?: { start?: string; end?: string }): string {
...
+                <br />
+                <Button
+                    variant="text"
+                    onClick={(): void => scrollTo('pathogens')}
+                >
+                    pathogens
+                </Button>
+                <br />
+                <Button variant="text" onClick={(): void => scrollTo('notes')}>
+                    notes
```
allysonjp715, line 208: To match the sections that currently exist in the details page I think these titles should be (missing/differing titles bolded):
case data
demographics
location
event history,
symptoms
**preexisting conditions**
**transmission**
travel history
**pathogens & genome sequencing**
notes

I would also split up pathogens and genome sequencing since they aren't related. Pathogens is other pathogens the case has and genome sequencing is pertaining to the disease in question (covid). But happy for that to be in a different PR since this is just about adding the section titles.

verification/curator-service/ui/src/components/ViewCase.tsx
attwad: preexisting conditions is missing from mocks and is also quite long, I've added transmission to match the mocks .

verification/curator-service/ui/src/components/ViewCase.tsx
attwad: yes, default is not smooth.

PR #645, opened 2020-07-27
Fix UX of selecting Unknown fields

### LINKED ISSUE #444: UX for unknown fields
Selecting Unknown makes no change to the field, just remains blank.
UX to confirm expected functionality and UX when selected.
See attached design currently.

Screen Shot 2020-07-06 at 2 10 18 PM

allysonjp715:
Fixes #444

### DIFF
--- a/verification/curator-service/ui/src/components/new-case-form-fields/Symptoms.tsx
+++ b/verification/curator-service/ui/src/components/new-case-form-fields/Symptoms.tsx
@@ -11,7 +11,7 @@ import { useFormikContext } from 'formik';
 
 // TODO: get values from DB
 const symptomStatusValues = [
-    undefined,
+    'Unknown',
     'Asymptomatic',
     'Symptomatic',
     'Presymptomatic',
--- a/verification/curator-service/ui/src/components/new-case-form-fields/TravelHistory.tsx
+++ b/verification/curator-service/ui/src/components/new-case-form-fields/TravelHistory.tsx
@@ -33,10 +33,10 @@ const useStyles = makeStyles(() => ({
     },
 }));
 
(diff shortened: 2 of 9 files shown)


verification/curator-service/ui/src/components/new-case-form-fields/Symptoms.tsx
```diff
@@ -11,7 +11,7 @@ import { useFormikContext } from 'formik';
 
 // TODO: get values from DB
```
attwad, line 12: nit but should we really be getting those 3 from the DB? I don't think so.

verification/curator-service/ui/src/components/new-case-form-fields/TravelHistory.tsx
```diff
@@ -33,10 +33,10 @@ const useStyles = makeStyles(() => ({
     },
 }));
 
-const hasTravelledValues = [undefined, 'Yes', 'No'];
+const hasTravelledValues = ['Unknown', 'Yes', 'No'];
 
 // TODO: get values from DB.
```
attwad, line 38: ditto

verification/curator-service/ui/src/components/new-case-form-fields/Symptoms.tsx
allysonjp715: If we are going to verify the field is one of these 3 values in the backend, it would be nice to only have the values defined in one location eventually. Agree it's not high priority.

verification/curator-service/ui/src/components/new-case-form-fields/TravelHistory.tsx
allysonjp715: ditto :)

PR #652, opened 2020-07-27
Create a batchCreate endpoint

allysonjp715:
To be used in #450

### DIFF
--- a/verification/curator-service/api/openapi/openapi.yaml
+++ b/verification/curator-service/api/openapi/openapi.yaml
@@ -273,6 +273,25 @@ paths:
           $ref: "#/components/responses/422"
         "500":
           $ref: "#/components/responses/500"
+  /cases/batchCreate:
+    post:
+      tags: [Case]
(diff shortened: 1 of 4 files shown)


allysonjp715:
> Implementation looks great, but question on the approach.
> 
> If this is for the linked FR adding multiple rows for a given form, do we want to use this approach wherein the API accepts an array of the same case n times? Do we want to consider perhaps a param on the existing API (e.g. `caseCount`), since the actual case body is the same?
> 
> Semi-relatedly, I'll likely be adding something like the second for upsert (for the England case data, there are ~50k unique rows, which expand out to ~900k total rows if you factor in `caseCount` from their partially disaggregated data).

I was thinking that this API is more flexible and could be used in either situation, but if we don't need the flexibility and the latter is more efficient then we should do the latter. I'll sync with you today.

allysonjp715:
Dropping this in favor of AB's suggestion.

PR #653, opened 2020-07-28
Update parsing_library layer to work with new SAM CLI version (1.0.0)

axmb:
Moving `parsing_lib` into a dedicated package so that it's compatible with the new SAM CLI version. The [new version](https://aws.amazon.com/blogs/compute/the-aws-serverless-application-model-cli-is-now-generally-available/) includes actual support for building layers locally, so I shuffled the code/manifest to support that. Previously, the CLI must have implemented some special case logic to make layers appear to function.

Includes a workaround for a bug that I apparently hit the bleeding edge of today when I upgraded to SAM CLI 1.0 :drop_of_blood:  (please don't officially reference the issue here).


axmb:
[On-board Anya/Felix](https://trello.com/c/HKpBYN6w/302-on-board-anya-felix)

iamleeg:
All looks good, I remember some discussion before about them magically fixing up PYTHON_PATH on the lambda runner so that other files from the container were available for import, is that what's changed upstream to motivate this change?

axmb:
> All looks good, I remember some discussion before about them magically fixing up PYTHON_PATH on the lambda runner so that other files from the container were available for import, is that what's changed upstream to motivate this change?

This was motivated by the new/GA v1.0.0 SAM CLI officially supporting building Layers. Before this, it was permissible to use the parsing_lib file directly, but now that we're building the layer locally, it's simpler to structure it as a module for SAM. I think previously (and perhaps still in 1.0.0, though we aren't doing this anymore) files in `$layerDir/python` were added to PYTHONPATH; so for the SAM runs, the parser imported the file directly. But, in tests, we imported the `$layerDir` (`common/python`) as a module.

Now that we're actually building the Layer with SAM, it seems simpler to treat it as a module in both cases.

For the full story, there's some more detail about how Lambda/SAM builds artifacts and adds certain `opt` folders to PYTHONPATH. [This touches on it](https://aws.amazon.com/blogs/compute/working-with-aws-lambda-and-lambda-layers-in-aws-sam/), though the exact mechanisms have changed (I couldn't find a complete, up-to-date article).

Issue #572, opened 2020-07-16
Mapbox missing key location data in the UK


sratcliffe118:
Mapbox has spotty coverage of Admin level 3 data such as Districts / Councils and Boroughs (specifically for London)

As an example the London boroughs of Southwark, Westminster and Kensington and Chelsea, 3 of the biggest areas are missing completely from the MapBox autocomplete options. This is on top of the USA county level issues we have run into.

I propose we meet to discuss this item with Timothe when back next week, as we will need to make a call on how we want to proceed urgently.

sratcliffe118:
**Summary of issues**

**England**
- Missing London Borough information: Westminster, Southwark, Lambeth, Tower Hamlets, Hammersmith and Fulham, Richmond upon Thames
- Missing district level data: Example, Ribble Valley
- Mixed regions coverage: 
Included: West Midlands, London
Not included: South West, South East, East of England, East Midlands. Yorkshire and The Humber, North West, North East

**USA**
No county level data

attwad:
I am back, this is similar to #426 hopefully the tileset/boundaries APIs can give us that information, I should have more clarity on it after I got a bit more familiar with those.

attwad:
Ah that's actually different, there are NO results at all for southwark, london. That's problematic indeed and the tileset API wouldn't help at all... We can:
1. ask mapbox folks why there are no London Boroughs in their data
2. figure out a way for curators to manually provide geocoding information when there are no results
3. plug-in another geocoder as fall-back method

@JaredHawkins could you please send an email to your mapbox contacts asking why they are missing that data? (1)

in the meantime I can quickly draft a component to submit geolocation information (2) and we can think about (3) later.

JaredHawkins:
@attwad @sratcliffe118 Happy to email Mapbox, but can you give me a bit more info (and maybe some example queries)?  And can you please confirm if you have also looked at the Boundaries API for the same locations?  Thanks!

attwad:
Stefen summarized the issue in https://github.com/open-covid-data/healthmap-gdo-temp/issues/572#issuecomment-659549805, I also checked that searching for "Southwark" in their geolocation api playground didn't return the Borough as expected so this is unrelated to us using the permanent places API endpoint.

This is also unrelated to the boundaries API, we would use the boundaries API to get the right admin levels for a given point but here we don't even have a point.

JaredHawkins:
[...]

PR #654, opened 2020-07-28
Use only boundaries API admin areas 

attwad:
Turns out administrative areas are hard [citation needed], we were previously trying to guess them from the geocode feature context but it was mostly wrong past the admin 1, admin 2 and admin 3 definitions depend on countries (for example Brasil doesn't have an admin 3, UK has "country" as admin 1, county as admin 2, etc. It's a mess and we can't just try to get the feature type "region" for admin 1, "district" for admin2 etc as we were trying to do before.
From now on we'll rely on the mapbox boundaries API which has very decent coverage and give us correct results.

Just check out https://www.mapbox.com/boundaries/#coverage-map and visit various countries to see how many different admin level -> meaning there are. I realized that looking at that page, hence this PR.

Also:
- reorder definitions of methods in admin fetcher (some lint issues when using stuff that wasn't defined prior to being used, worked but I am not sure what the problem was about...)

- cache results of boundaries API to lessen the costs and get faster results as we do for geocode results

Example of wrong geocode before:
image

Example of correct geocode after:
image

For #572

### DIFF
--- a/verification/curator-service/api/test/geocoding/mapbox.test.ts
+++ b/verification/curator-service/api/test/geocoding/mapbox.test.ts
@@ -58,9 +58,6 @@ describe('geocode', () => {
         });
         expect(feats).toHaveLength(1);
         const wantFeature: GeocodeResult = {
-            /*administrativeAreaLevel1: 'Rhône',
-            administrativeAreaLevel2: '',
-            administrativeAreaLevel3: 'Lyon',*/
             country: 'France',
             geometry: { latitude: 45.75889, longitude: 4.84139 },
             place: '',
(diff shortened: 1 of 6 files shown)


verification/curator-service/api/test/geocoding/mapbox.test.ts
```diff
@@ -58,9 +58,9 @@ describe('geocode', () => {
         });
         expect(feats).toHaveLength(1);
         const wantFeature: GeocodeResult = {
-            administrativeAreaLevel1: 'Rhône',
+            /*administrativeAreaLevel1: 'Rhône',
```
allysonjp715, line 61: Uncomment or remove

attwad: ah sorry, done.

Issue #617, opened 2020-07-24
Potentially revisit decision to use modal dialogs or polish them enough to keep


attwad:
Warning: this is my opinion, I'm not a UXer, just speaking from an engineering perspective. I'll try to argument simply why I believe huge modal dialogs are not a good fit for this project.

**Describe the bug**

In no particular order:

## 1
Start page looks like this:
![image](https://user-images.githubusercontent.com/1255432/88371410-42ee7880-cd94-11ea-83c1-09ca9d5f5385.png)

When you open a modal dialog, what's below gets erased:
![image](https://user-images.githubusercontent.com/1255432/88371437-50a3fe00-cd94-11ea-9dd5-bfaa34da2c6e.png)

So this is *not* really popping up on top of anything really.

## 2 wasted unusable space on the left
Related to 1 above, the mocks seem to indicate there should be some space between the menu and the dialog, why? to make it clearer that this is a popup? does that justify the wasted space on the left?

## 3 modal dialogs killed mobile friendliness
Before, it wasn't perfect but at least mobile was usable. Now with the menu always opened and the modal dialog popped up further right, it's just impossible to enter anything.
![image](https://user-images.githubusercontent.com/1255432/88371594-a37db580-cd94-11ea-91b7-cde4b1ce4c9e.png)

## 4 close button isn't handled well
Right now the close button of the modal dialog doesn't always redirects to /cases, that's counter intuitive. I would expect it to just disappear and show what's beneath it, but again related to 2) there's "nothing" below it so it can't really do that.

## 5 I don't see a close match with the material UI guidelines for usage of modal dialogs in our project:

> A dialog is a type of modal window that appears in front of app content to provide critical information or ask for a decision. Dialogs disable all app functionality when they appear, and remain on screen until confirmed, dismissed, or a required action has been taken.
> Dialogs are purposefully interruptive, so they should be used sparingly.

IMHO we shouldn't be "interrupting" the user here, they are just trying to do *one* task: submit cases, let them have the whole screen available to do that instead of a fraction of it with noise around.

**To Reproduce**
Open the modal dialogs Click on create a new case or bulk entry.

Anyway, just a few thoughts, I thought I'd put them together in a bug so we can discuss how to move forward.
[...]

PR #657, opened 2020-07-28
Do not show the details page nav menu on small screens

attwad:
Trying to get back some of my screen estate 💻 

ezgif com-video-to-gif

Related to #617


attwad:
case form has the little confirmation icons on those sections which people love... so not sure, what we should do is collapse the app left menu on small screen yes. I might look into that next.

Issue #439, opened 2020-07-03
Symptoms: Show the most relevant symptoms for the disease as a quick selection option


sratcliffe118:
See attached UX
<img width="305" alt="Screen Shot 2020-07-03 at 11 24 26 AM" src="https://user-images.githubusercontent.com/63205311/86460276-fd540800-bd1f-11ea-90ee-bc99940a69b8.png">

allysonjp715:
Do we want these to be a reflection of the most commonly used symptoms in the DB? Or shall we hard code some in that we know are common to COVID?

acansler:
I think it would be appropriate to make them for Covid. Otherwise, we will likely see a lot of searching for those, and given that there are multiple forms of "cough", this may improve data consistency.

shroothless:
I would say they should be the most commonly used symptoms in the DB and not hardcoded. Also I think this is a nice to have since likely behaviour.

sratcliffe118:
Yes lets go with the most commonly used symptoms in the DB, no hardcoding

PR #658, opened 2020-07-28
Add an endpoint for listing the most frequently used symptoms

allysonjp715:
Beginning of #439

### DIFF
--- a/data-serving/data-service/src/controllers/case.ts
+++ b/data-serving/data-service/src/controllers/case.ts
@@ -226,6 +226,7 @@ export const listSymptoms = async (
         const symptoms = await Case.aggregate([
             { $unwind: '$symptoms.values' },
             { $sortByCount: '$symptoms.values' },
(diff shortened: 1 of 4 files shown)


data-serving/data-service/src/controllers/case.ts
```diff
@@ -207,3 +207,33 @@ export const del = async (req: Request, res: Response): Promise<void> => {
...
+): Promise<void> => {
+    const limit = Number(req.query.limit) || 5;
+    if (limit < 1) {
+        res.status(422).json('limit must be > 0');
+        return;
+    }
+    try {
+        const symptoms = await Case.aggregate([
+            { $unwind: '$symptoms.values' },
+            { $sortByCount: '$symptoms.values' },
```
attwad, line 228: Should we order by name as a second stage to have consistent results? (for ties)

allysonjp715: Done.

PR #659, opened 2020-07-28
Automatically hide/show the app menu on small/medium+ screens

attwad:
Had to rewrite the App component into a functional component because there is no class component access to media queries.

ezgif com-video-to-gif (1)

Related to #617


allysonjp715:
Is it still possible to open the drawer when it is a small screen?

attwad:
Yes of course

PR #661, opened 2020-07-28
Add endpoint in curator-service API for listing most frequently used …

allysonjp715:
…symptoms

#439

### DIFF
--- a/verification/curator-service/api/openapi/openapi.yaml
+++ b/verification/curator-service/api/openapi/openapi.yaml
@@ -273,6 +273,32 @@ paths:
           $ref: "#/components/responses/422"
         "500":
           $ref: "#/components/responses/500"
+  /cases/symptoms:
+    get:
+      summary: Lists most frequently used sypmtoms
+      operationId: listSymptoms
+      parameters:
+        - name: limit
+          in: query
+          description: The number of items to return
+          required: false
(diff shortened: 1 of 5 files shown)


verification/curator-service/api/openapi/openapi.yaml
```diff
@@ -273,6 +273,32 @@ paths:
           $ref: "#/components/responses/422"
         "500":
           $ref: "#/components/responses/500"
+  /cases/symptoms:
```
attwad, line 276: may I suggest to rename /cases/symptoms into /suggest(ion)/symptoms ?
Reasoning is that if we ever want to expose our symptoms as an API it should be different than the suggest one. Ditto for previous PR sorry I missed that.

allysonjp715: Perhaps we can cross that bridge when we get there? I think it's unlikely we'll expose symptoms as an API given we have the list in the github file. But if we did we could also modify this this endpoint at that point - make sort and limit optional query params.

attwad: the github file is just the suggestions, an API would imply more processing of the data or something related to a particular outbreak for example, but yeah we can cross that bridge later.

PR #664, opened 2020-07-28
Add frequently used symptoms as clickable chips

### LINKED ISSUE #439: Symptoms: Show the most relevant symptoms for the disease as a quick selection option

allysonjp715:
Fixes #439

symptomgif

(lag is only from gif creation, not on actual site)


attwad:
FYI I use https://ezgif.com/video-to-gif to gifs .mov files that I take with my laptop, works very nicely and doens't have this lag.

Issue #619, opened 2020-07-24
Generate acronyms for preexisting conditions


attwad:
So that the suggest box can suggest using those acronyms.

sratcliffe118:
@attwad do we need someone from BCH or a set of curators to take a pass through these to confirm they make sense?

attwad:
The file is here:
https://github.com/globaldothealth/list/blob/main/suggest/preexisting_conditions.txt

We have about 9K suggestions in this dialog, a good thing would be for someone knowledgeable to have a pass at it first and remove some of the cruft.

Then we have two choices, do we always want to see the acronyms in the form and the details pages? if so we have to generate them and store them in the same file. Otherwise we can generate the acronyms only for the suggest input. Either way I think we should try to just auto-generate them and we can manually apply fixes for those who ended up wrong based on feedback from curators.

My personal take: generate all acronyms automatically and show them everywhere next ot the name, example: "Brucella abortus brucellosis" -> "Brucella abortus brucellosis (BAB)".

sg?

sratcliffe118:
SG, @lagood1 would you be able to take a pass through?

"Brucella abortus brucellosis" -> "Brucella abortus brucellosis (BAB)".
I agree with this  @attwad to confirm, if a curator was to search BAB, this option would come up? If so then I think this gives us what we need

attwad:
> SG, @lagood1 would you be able to take a pass through?
> 
> "Brucella abortus brucellosis" -> "Brucella abortus brucellosis (BAB)".
> I agree with this @attwad to confirm, if a curator was to search BAB, this option would come up? If so then I think this gives us what we need

Yes it would suggest it.

sratcliffe118:
Great, that approach LGTM

attwad:
I've added a script and generated the list of basic acronyms for preexisting conditions, whoever does a pass can give some feedback, I can amend accordingly.

lagood1:
Yes, I will take a look and get the edited file back to you

PR #669, opened 2020-07-29
Add acronym generator and generate for preexisting conditions

attwad:
Script to generate simple acronyms based on first character of each word in a sentence.

Has some basic checks for alphanumerics, presence of existing acronyms and profanity words.

For #619

### DIFF
--- a/suggest/acronyms/gen_acros.py
+++ b/suggest/acronyms/gen_acros.py
@@ -1,0 +1,52 @@
+"""Script to generate simple acronyms based on first character of each word in a sentence.
+
+Has some basic checks for alphanumerics, presence of existing acronyms and profanity words.
+
+Run with python3 gen_acros.py -f <path to file containing one sentence per line>
+
+The provided file will be rewritten with the acronyms added.
+"""
+
+from better_profanity import profanity
+
+import pathlib
+import argparse
+
+parser = argparse.ArgumentParser(
+    description='Acronym generation script')
+parser.add_argument('-f', '--file', type=str,
+                    help='Path to the file containing the list of sentences to generate acronyms for')
+
+class AcronymGenerator:
+    def Generate(self, filePath: str):
+        print('Generating acronyms from file', filePath)
+        profanity.load_censor_words()
+        lines = set()
+        with open(filePath, "r") as f:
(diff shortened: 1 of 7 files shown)


suggest/acronyms/gen_acros.py
```diff
@@ -0,0 +1,52 @@
+"""Script to generate simple acronyms based on first character of each word in a sentence.
```
allysonjp715, line 1: nit on filename: I tend to prefer full names like generate_acronyms over abbreviations. While I understood from context, gen_acros could be confusing out of context.

suggest/acronyms/gen_acros.py
```diff
@@ -0,0 +1,52 @@
...
+
+Has some basic checks for alphanumerics, presence of existing acronyms and profanity words.
+
+Run with python3 gen_acros.py -f <path to file containing one sentence per line>
+
+The provided file will be rewritten with the acronyms added.
```
allysonjp715, line 7: Could you give an example so the format of "disease name (DA)" is clear?

attwad:
> Question on approach: This will have us save the acronyms in the DB too. Do we want that? If the acronyms are updated at any point, we could have two different saved entries for the same DB: one with the old acronym and one with the new.

I think that's acceptable as the simplest solution, what drove this feature request is user feedback saying that they didn't remember the preexisting conditions but just their acronyms, storing them in the file like this and the DB makes suggest trivial and also shows the acronyms wherever the preexisting conditions are shown, so it's consistent.

FWIW I also tried just generating the acronyms on the fly int he FormikAutocomplete with a special renderOptions={} directive but suggestions aren't picked up this way (you can't search for the acronym, it has to be part of the option itself).

The "right" way to do this might be to move from a txt file in github to a proper HTTP API that suggest can use where we would store the preexisting conditions on mongo db and their acronyms, we'd then need to store the acronyms IDs or potentially denormalize them in the cases collection, I don't think it is worth the added complexity tbh. Keeping those in the txt file also makes it trivial for curators to suggest one-off fixes to those acronyms once automatically generated. I also added a check for existing acronyms so that they are not overwritten, making it possible to do those one-offs fixes post initial generation.

PR #670, opened 2020-07-29
Handle conditional fields better

### LINKED ISSUE #618: Symptoms should clear if changing symptom status
To reproduce:
- Click symptomatic for symptom status
- Add a symptom
- Click unknown for symptom status

Issues:
- Green checkmark is still shown in table of contents
- The symptom is added to the datalist

allysonjp715:
- Do not store results when conditions are not met
- Do not show green checkmark when conditions are not met

Fixes #618

### DIFF
--- a/verification/curator-service/ui/src/components/CaseForm.tsx
+++ b/verification/curator-service/ui/src/components/CaseForm.tsx
@@ -378,13 +378,17 @@ class CaseForm extends React.Component<Props, CaseFormState> {
                 },
                 {
                     name: 'icuAdmission',
-                    dates: values.icuAdmissionDate,
+                    dates:
+                        values.admittedToIcu === 'Yes'
+                            ? values.icuAdmissionDate
+                            : undefined,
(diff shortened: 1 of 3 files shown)


verification/curator-service/ui/src/components/CaseForm.tsx
```diff
@@ -405,7 +409,11 @@ class CaseForm extends React.Component<Props, CaseFormState> {
                 }),
             symptoms: {
                 status: unknownToUndefined(values.symptomsStatus),
-                values: values.symptoms,
+                values:
+                    values.symptomsStatus === 'Symptomatic' ||
```
attwad, line 413: Am i understanding correctly that you could select Symptomatic, then select symptoms, then remove the Symptomatic selection but symptoms would still be sent? If this wasn't happening I'm not sure what we're getting by doing this.

allysonjp715: Yes that was happening. A big bug!

PR #680, opened 2020-07-29
Add batchUpsert method to data service case API

axmb:
Primary goal here is to make the bulk upsert MongoDB invocation more efficient.

In pursuit of avoiding a super large/dense PR, I'm cutting this at what's _sort of_ an intermediate state. Rather than expose this endpoint directly in a subsequent PR, I'll probably wrap it together with the existing batchValidate method before doing so (in that same PR). Happy to hear thoughts on this.

### DIFF
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -151,7 +151,7 @@ paths:
           $ref: "#/components/responses/500"
   /cases/symptoms:
     get:
-      summary: Lists most frequently used sypmtoms
+      summary: Lists most frequently used symptoms
       operationId: listSymptoms
       parameters:
         - name: limit
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -131,6 +131,24 @@ paths:
           $ref: "#/components/responses/207BatchValidateCaseResponse"
         "500":
           $ref: "#/components/responses/500"
+  /cases/batchUpsert:
+    post:
+      summary: Batch upserts cases
+      operationId: batchUpsertCases
+      requestBody:
+        description: Cases to upsert
+        required: true
+        content:
+          application/json:
+            schema:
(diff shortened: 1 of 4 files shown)


data-serving/data-service/api/openapi.yaml
```diff
@@ -541,6 +573,12 @@ components:
         application/json:
           schema:
             $ref: '#/components/schemas/BatchValidateCaseResponse'
+    "207BatchUpsertCaseResponse":
```
iamleeg, line 576: Wow, I don't think I've seen status 207 correctly used in an HTTP API before. Chapeau 🎩

data-serving/data-service/src/controllers/case.ts
```diff
@@ -136,6 +136,114 @@ export const batchValidate = async (
...
+ * all provided cases without validation errors will be written.
+ *
+ * TODO: Wrap batchValidate in this method.
+ */
+export const batchUpsert = async (
+    req: Request,
+    res: Response,
+): Promise<void> => {
+    try {
+        const toBeUpsertedCaseIds = await findCaseIdsWithCaseReferenceData(req);
```
khmoran, line 196: if this still isn't as fast as you want, in theory you could have this fn add the ids to the requests, so that below you don't have to re-filter? (assuming that's even an option in mongodb's bulkwrite fn)

data-serving/data-service/api/openapi.yaml
axmb: It's a bit elusive, isn't it? I'm hoping the final state here will taken better advantage yet.

data-serving/data-service/src/controllers/case.ts
axmb: Meaning to perform the updateOne()'s specifying the _id, rather than the caseRef data pair?

I think that's a solid idea. I'm a little worried about this step being generally slow, but plan to test and make these kinds of tweaks in the subsequent change.

Issue #679, opened 2020-07-29
Implement full search API


attwad:
Pendign exact mocks/behavior from @shroothless but we should implement the search bar full syntax to be able to select columns and to full text search in the same spot.

![image](https://user-images.githubusercontent.com/1255432/88843116-e4d0f380-d1e0-11ea-92d3-8a31d0b30ab8.png)

Bonus points for recent searches showing up, this is a nice to have though.

PR #682, opened 2020-07-30
add basic search query parser

attwad:
For #679

This makes use of https://www.npmjs.com/package/search-query-parser which is a zero-dependency parser that does what we want AFAICT. I'm happy to not have to write such a common parser/tokenizer, next step is to specify keywords and modify the mongo query accordingly.

### DIFF
--- a/data-serving/data-service/src/controllers/case.ts
+++ b/data-serving/data-service/src/controllers/case.ts
@@ -1,6 +1,7 @@
 import { Request, Response } from 'express';
 
 import { Case } from '../model/case';
+import parseSearchQuery from '../util/search';
 
 /**
  * Get a specific case.
@@ -33,13 +34,16 @@ export const list = async (req: Request, res: Response): Promise<void> => {
         return;
     }
     // Filter query param looks like &q=some%20search%20query
-    const searchQuery = String(req.query.q || '').trim();
-    const query = searchQuery
+    if (typeof req.query.q !== 'string' && typeof req.query.q !== 'undefined') {
+        res.status(422).json('q must be a unique string');
+        return;
+    }
+    const parsedSearch = parseSearchQuery(req.query.q || '');
+    const query = parsedSearch.fullTextSearch
         ? {
-              $text: { $search: searchQuery },
+              $text: { $search: parsedSearch.fullTextSearch },
           }
         : {};
-
     // Do a fetch of documents and another fetch in parallel for total documents
     // count used in pagination.
     try {
--- a/data-serving/data-service/src/util/search.ts
+++ b/data-serving/data-service/src/util/search.ts
@@ -1,0 +1,25 @@
+import { parse, SearchParserResult } from 'search-query-parser';
+
+export interface ParsedSearch {
+    fullTextSearch?: string;
+}
+
+export default function parseSearchQuery(q: string): ParsedSearch {
+    q = q.trim();
+    const parsedSearch = parse(q, {
+        offsets: false,
+        // TODO: Specify tokens here.
+    });
+    const res: ParsedSearch = {};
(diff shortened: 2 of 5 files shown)


data-serving/data-service/src/controllers/case.ts
```diff
@@ -33,13 +34,16 @@ export const list = async (req: Request, res: Response): Promise<void> => {
         return;
     }
     // Filter query param looks like &q=some%20search%20query
-    const searchQuery = String(req.query.q || '').trim();
-    const query = searchQuery
+    if (typeof req.query.q !== 'string' && typeof req.query.q !== 'undefined') {
```
allysonjp715, line 37: Can these checks be done in the OpenAPI?

data-serving/data-service/src/util/search.ts
```diff
@@ -0,0 +1,25 @@
+import { parse, SearchParserResult } from 'search-query-parser';
+
+export interface ParsedSearch {
+    fullTextSearch?: string;
+}
+
+export default function parseSearchQuery(q: string): ParsedSearch {
```
allysonjp715, line 7: Could you just return a string rather than a ParsedSearch object containing a string?

data-serving/data-service/test/util/search.test.ts
```diff
@@ -0,0 +1,18 @@
+import parseSearchQuery from '../../src/util/search';
+
+describe('search query', () => {
```
allysonjp715, line 3: All of these tests don't many any changes to the query provided. Could you add a test where it does?

data-serving/data-service/src/controllers/case.ts
attwad: We could, but we still have to do those here nonetheless, it's typescript types for req.query that is either string | string[] for any request params.

data-serving/data-service/src/util/search.ts
attwad: No, I'll need separate fields in that struct being returned for the extracted keywords once we start doing searches with "is:verified curator:foo@bar.com" etc.

data-serving/data-service/test/util/search.test.ts
attwad: sorry can you say that again? I think there's a typo and I don't understand what needs to change sorry

data-serving/data-service/src/controllers/case.ts
allysonjp715: Got it. We should add that indication to openAPI as well. We are using openAPI for verification and IIRC query is specified to be a string right now, so if a string[] was passed openAPI would fail validation.

data-serving/data-service/test/util/search.test.ts
allysonjp715: I was saying that all the tests do is take the input string and put it in an object e.g. `'some query'` -> `{ fullTextSearch: 'some query' }`
But I now see that is all search.ts does for now until the tokens are added, so no updates necessary.

data-serving/data-service/src/controllers/case.ts
attwad: OpenAPI has this:
```
- name: q
          in: query
          description: The search query
          required: false
          schema:
            type: string
```

So this is correct, I don't know what you want to add to it?
If clients pass multiple &q= params to the URL that's their fault and yes it will fail validation because it should fail validation I think we're good.
Let me know if I missed something.

allysonjp715: Sorry I misread your previous comment, I thought you were saying query *could* be a string[] and I was asking you to add that to the types. But now I see you were saying typescript determines that it could be a string[], but we don't want it to be.

PR #683, opened 2020-07-30
Allow typing in a source without selecting a dropdown option

### LINKED ISSUE #632: Manual data entry: Data source entry
Bug: Modify data source field to allow curators to copy paste the source, and if source already exists, not having to select it from the drop down.

@allysonjp715 can we support this functionality with the current form? Essentially an auto complete without a picker if the exact source already exists?

If this is more than a small tweak, @shroothless can we tweak the UX to make clearer to the curator that they need to select the source to continue

allysonjp715:
Fixes #632

### DIFF
--- a/verification/curator-service/ui/cypress/integration/components/NewCaseForm.spec.ts
+++ b/verification/curator-service/ui/cypress/integration/components/NewCaseForm.spec.ts
@@ -53,7 +53,7 @@ describe('New case form', function () {
         cy.visit('/cases/new');
         cy.contains('Create new COVID-19 line list case');
         cy.get('div[data-testid="caseReference"]').type('www.new-source.com');
-        cy.contains('li', 'www.new-source.com').click();
+        cy.get('input[name="caseReference.sourceName"]').click();
         cy.get('input[name="caseReference.sourceName"]').type('New source');
         cy.get('div[data-testid="location"]').type('France');
         cy.contains('li', 'France').click();
--- a/verification/curator-service/ui/src/components/common-form-fields/Source.tsx
+++ b/verification/curator-service/ui/src/components/common-form-fields/Source.tsx
@@ -172,7 +172,9 @@ export function SourcesAutocomplete(
             <Autocomplete
                 itemType="CaseReferenceForm"
                 getOptionLabel={(option: CaseReferenceForm): string =>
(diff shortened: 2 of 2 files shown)


verification/curator-service/ui/src/components/common-form-fields/Source.tsx
```diff
@@ -185,8 +185,24 @@ export function SourcesAutocomplete(
...
+                    if (typeof newValue === 'string') {
+                        const existingOption = options.find(
+                            (option) => option.sourceUrl === newValue,
+                        );
+                        newValue = existingOption ?? {
+                            inputValue: newValue,
+                            sourceUrl: newValue,
+                            sourceId: '',
+                            sourceName: '',
+                            additionalSources: ([] as unknown) as [
```
attwad, line 201: why is the `[] as unknown` needed?

verification/curator-service/ui/cypress/integration/components/NewCaseForm.spec.ts
```diff
@@ -53,7 +53,7 @@ describe('New case form', function () {
         cy.visit('/cases/new');
         cy.contains('Create new COVID-19 line list case');
         cy.get('div[data-testid="caseReference"]').type('www.new-source.com');
-        cy.contains('li', 'www.new-source.com').click();
+        cy.get('input[name="caseReference.sourceName"]').click();
```
attwad, line 56: do we want to keep a test where the source is clicked vs just entered?

verification/curator-service/ui/src/components/common-form-fields/Source.tsx
allysonjp715: This is the error message when it is removed: `Conversion of type '[]' to type '[{ sourceUrl: string; }]' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.`

verification/curator-service/ui/cypress/integration/components/NewCaseForm.spec.ts
allysonjp715: The other tests in this file still click the dropdown entry. There is also a test in BulkCaseForm that clicks the dropdown entry for a new source.

PR #687, opened 2020-07-30
Support keywords in search

attwad:
Future improvements:
- add tooltip to searchbar to explain the search syntax
- add more keywords based on UX feedback

### DIFF
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -46,7 +46,7 @@ paths:
               summary: Full text search with items that must or must not be present.
             keywords:
               value: "curator:foo@bar.com,baz@meh.com country:france"
-              summary: "values are OR'ed for the same keyword and all keywords are AND'ed. Only equality operator is supported. Supported operators are: curator, gender, nationality, occupation, country, outcome, caseid, source, admin1, admin2, admin3"
+              summary: "values are OR'ed for the same keyword and all keywords are AND'ed. Keyword values can be quoted for multi-words matches and concatenated with a comma to union them. Only equality operator is supported. Supported operators are: curator, gender, nationality, occupation, country, outcome, caseid, source, admin1, admin2, admin3"
       responses:
         "200":
           $ref: "#/components/responses/200CaseArray"
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -46,7 +46,7 @@ paths:
               summary: Full text search with items that must or must not be present.
             keywords:
               value: "curator:foo@bar.com,baz@meh.com country:france"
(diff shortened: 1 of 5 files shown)


data-serving/data-service/api/openapi.yaml
```diff
@@ -40,6 +40,13 @@ paths:
           required: false
           schema:
             type: string
+          examples:
+            "full text search":
+              value: "this -butnotthis"
+              summary: Full text search with items that must or must not be present.
+            keywords:
```
allysonjp715, line 47: Should keywords be in quotes like "full text search" is?

data-serving/data-service/src/controllers/case.ts
```diff
@@ -44,17 +44,29 @@ export const list = async (req: Request, res: Response): Promise<void> => {
               $text: { $search: parsedSearch.fullTextSearch },
           }
         : {};
-    // Do a fetch of documents and another fetch in parallel for total documents
-    // count used in pagination.
+    // Fill in keyword filters.
     try {
+        let q = Case.find(query);
```
allysonjp715, line 49: q isn't the most clear variable name now that it means more than query

data-serving/data-service/src/util/search.ts
```diff
@@ -1,16 +1,40 @@
...
+    filters: {
+        path: string;
+        values: string[];
+    }[];
 }
 
+// Map of keywords to their case data path.
+// IMPORTANT: If you change this mapping, reflect the new keys in the openapi.yaml file as well.
+const keywords = new Map<string, string>([
+    ['curator', 'revisionMetadata.creationMetadata.curator'],
```
allysonjp715, line 15: I wonder if people will expect their edits to come up as well when searching for curator

data-serving/data-service/test/controllers/case.test.ts
```diff
@@ -105,11 +105,6 @@ describe('GET', () => {
             // Simulate index creation used in unit tests, in production they are
             // setup by the setup-db script and such indexes are not present by
             // default in the in memory mongo spawned by unit tests.
-            await mongoose.connect(process.env.MONGO_URL || '', {
```
allysonjp715, line 108: Should the comment above this be removed too?

data-serving/data-service/api/openapi.yaml
attwad: if they are multi-words yes, otherwise this is not needed. added a clarification comment.

data-serving/data-service/src/controllers/case.ts
attwad: sorry, can't get rid of my golang habits that easily ;) fixed.

data-serving/data-service/src/util/search.ts
attwad: hmm, yeah. we could add an "editor" keyword that would match if we ever get that FR.

data-serving/data-service/test/controllers/case.test.ts
attwad: no this is still needed, the new connections was not needed.

PR #688, opened 2020-07-30
Stop hard-coding revisionMetadata on client

khmoran:
`revisionMetadata` is no longer required as part of requests, so they can stop sending data that will be dropped anyway

(sorry for bundling with deps update, might be worth looking by commit)


data-serving/data-service/src/model/case.ts
```diff
@@ -43,10 +43,7 @@ const caseSchema = new mongoose.Schema(
             _id: false,
         },
         location: locationSchema,
-        revisionMetadata: {
-            type: revisionMetadataSchema,
-            required: true,
-        },
+        revisionMetadata: revisionMetadataSchema,
```
attwad, line 46: Hmm, revision metadata is not required to be part of the client request but it is required when stored no? why removing it here?

khmoran: We use this model from request objects, which are incomplete (in the preprocessor). It's unfortunate... I think the API layer should be more separate from storage but right now they're coupled.

attwad: dooohhh didn't we start using the openapi spec instead to verify requests?

khmoran: The openapi spec is verifying the requests, but we're not generating TS types for the requests & responses in the spec, so to create a model from a request, we still use these mongoose classes. This is generally OK except for the preprocessing step, where there is no revisionMetadata (yet), but I still need a model object.

@axmb and I were talking about using a library that would generate these request/response bindings for us so we can truly separate the API & data layers, but we're not there yet

PR #690, opened 2020-07-30
Prototype of storing case revisions on update

khmoran:


data-serving/data-service/src/model/case-revision.ts
```diff
@@ -0,0 +1,24 @@
...
+import { ObjectId } from 'mongodb';
+import mongoose from 'mongoose';
+
+const caseRevisionSchema = new mongoose.Schema({
+    case: {
+        type: caseSchema,
+        required: true,
+    },
+}).index(
+    { 'case._id': 1, 'case.revisionMetadata.revisionNumber': -1 },
```
axmb, line 12: What are we using this index for? Is it simply to enforce uniqueness?

data-serving/data-service/test/controllers/preprocessor.test.ts
```diff
@@ -224,4 +282,63 @@ describe('upsert', () => {
...
+        const upsertCase = {
+            ...minimalCase,
+            caseReference: {
+                ...minimalCase.caseReference,
+                sourceEntryId: 'case_id',
+            },
+        };
+        const requestBody = {
+            ...upsertCase,
+            curator: { email: 'creator@gmail.com' },
```
axmb, line 295: This poor person's inbox. Think of all the automated testing emails.

data-serving/data-service/src/model/case-revision.ts
khmoran: This will be to speed up the query for fetching a particular revision or all revisions for a given document. Happy to hold off on it until we have that functionality if you rpefer

data-serving/data-service/test/controllers/preprocessor.test.ts
khmoran: omg i never thought of that... maybe i'll change in the next PR for EXTRA SAFETY

PR #696, opened 2020-07-31
Maintain the sourceName when changing the sourceURL

### LINKED ISSUE #689: Source name is cleared if source URL is edited
**Describe the bug**

**To Reproduce**
1. Go to new case form
2. Enter a new source URL (not existing in sources)
3. Enter a name for the new source
4. Edit the source URL
5. Note that the source name is cleared

**Expected behavior**
Source name should still be there

allysonjp715:
Fixes #689


allysonjp715:
> if this wasn't caught it means there's no test checking for it (I don't see any test changed in this PR), please add one.

Done

PR #706, opened 2020-08-03
Have button text ellipsis on small screens

allysonjp715:
Before:
Screen Shot 2020-08-03 at 2 16 29 PM

After:
Screen Shot 2020-08-03 at 2 16 08 PM

### DIFF
--- a/verification/curator-service/ui/src/components/App.tsx
+++ b/verification/curator-service/ui/src/components/App.tsx
@@ -121,6 +121,12 @@ const useStyles = makeStyles((theme: Theme) => ({
     root: {
         display: 'flex',
     },
+    buttonLabel: {
+        display: 'block',
+        overflow: 'hidden',
+        whiteSpace: 'nowrap',
+        textOverflow: 'ellipsis',
+    },
     title: {
         flexGrow: 1,
     },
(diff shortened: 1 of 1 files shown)


verification/curator-service/ui/src/components/App.tsx
```diff
@@ -121,6 +121,12 @@ const useStyles = makeStyles((theme: Theme) => ({
     root: {
         display: 'flex',
     },
+    buttonLabel: {
```
axmb, line 124: Hm -- should we apply this globally to .button in `index.css`?

allysonjp715: That's a good idea but it's a little awkward to do that. First it has to be applied to the class `.MuiButton-label` which while I'm sure it won't change anytime soon, that name is controlled by the material code and therefore can change. Secondly, the styling from the material code supersedes the logic from `index.css`, so we would have to double up the selector making the code in `index.css` the following:
```css
.MuiButton-label.MuiButton-label {
  display: block;
  [...]
}
```

On top of that, right now we don't actually have anymore buttons that can be pushed to have a small width. Due to all of that I'll keep as is for now.

Issue #662, opened 2020-07-28
Bulk CSV upload: Error handling process, error reason in CSV for download


sratcliffe118:
**Feature request**
Redefined process to handle errors in data format in the bulk CSV file uploaded to G.h 

**Currently process**
- User configures the data using the instructions and standard data format 
- Submits the CSV via the curator portal
- Problem with data format rejects the entire upload 
- Error message per row provided in the UI

**Proposed process and UI**
- Curator follows the same process to submit the CSV file
- If an error is found in the upload the full file is rejects (no new cases created or updates made)
- The CSV is returned to the curator with a new column added that details the error for each row identified
- This means that the upload process will need to cycle through each row and report back an error if there is one found
- The Curator has the functionality to download the CSV with error column, and then work to correct them before resubmitting.
- Once fixed curator can start the upload process again

axmb:
Shru/Aaron -- should I go ahead with what we have spec'd [here](https://www.figma.com/file/9HJlRkTBCMEQdkJ6WQJYkv/061120_Global.Health-Line?node-id=854%3A0)? With the CSV containing the original data in addition to a new final `error` column.

acansler:
Sounds good to me. I'll update the re-write of the documentation to reflect the change. 

Small nit: It may not be immediately clear to curators that they will be downloading a .csv file with the errors. Might be some small friction there (at least initially).

axmb:
Cool, sounds great.

One other scenario to account for is formatting errors in the CSV itself. For instance, if the uploaded CSV is missing a required column (or, more strictly, if it contains an unrecognized column). We could add the full error message to each row in the downloadable CSV (in addition to adding missing columns), or display a message in the UI.

sratcliffe118:
@acansler 
Ack on the comment, we will need to ensure that the documentation is clear here. I think we can also include in the error message e.g. Error uploading data, download CSV for upload error codes / reasons.

sratcliffe118:
@axmb 
ideal scenario here is a single error message showing column missing for all files. 
However, i think for MVP we can live with the same experience for all error messages, i.e. in the CSV per row.

PR #708, opened 2020-08-03
 Add additional fields to bulk upload format.

### LINKED ISSUE #646: Bulk CSV: Additional field columns to include in Bulk upload template
**Feature Request**
Add the **bolded** data fields to the Bulk upload template

1. sourceEntryId
2. country
3. admin1
4. admin2
5. admin3
6. gender
7. ageRangeStart
8. ageRangeEnd
9. **Race / ethnicity**
10. **Nationality**
11. **Occupation**
12. dateConfirmed
13. **Method of confirmation**
14. **Symptom onset date: (Note when there is data entered into this field we should also set the field Symptom status = Symptomatic )**
15. **Symptoms**
16. hospitalized 
17. dateHospitalized
18. **ICU admission: yes / no**
19. **ICU admission date**
20. **Pre existing conditions: yes/ no**
21. outcome 
22. dateOutcome
23. caseCount

axmb:
Closes #646.

### DIFF
--- a/verification/curator-service/ui/README.md
+++ b/verification/curator-service/ui/README.md
@@ -22,8 +22,24 @@ corresponding to this CSV format is available
 [here](https://docs.google.com/spreadsheets/d/1J-C7dq1rNNV8KdE1IZ-hUR6lsz7AdlvQhx6DWp36bjE).
 The following fields are currently incorporated:
 
+##### Case reference
 - `sourceEntryId`: Optional. The UUID provided by the data source to identify
 the specific case.
+
+##### Demographics
+- `gender`: Optional. One of `Female`, `Male`, `Non-binary/Third gender`, or
+`Other`.
+- `ageRangeStart`: Optional. Number (years) from 0 to 120.
+- `ageRangeEnd`: Optional. Number (years) from 0 to 120.
+- `ageRange`: Optional. Convenience field accepting an age range in the format
+`start-end`. If provided in this format, will be used in lieu of
+`ageRangeStart` and `ageRangeEnd`.
+- `ethnicity`: Optional.
+- `nationalities`: Optional. Delimit multiple nationalities with a semicolon
+(e.g., `British;Indian`).
+- `occupation`: Optional.
+
+##### Location
 - `country`: **Required**. Country name text (e.g., "United States").
 - `admin1`: Optional. Meaning differs by country (e.g., this is state in the
 US).
@@ -34,17 +50,34 @@ US).
 - `admin3`: Optional. Meaning differs by country (e.g., this is municipality in the
 US).
 [See full definition here](https://en.wikipedia.org/wiki/List_of_administrative_divisions_by_country).
-- `gender`: Optional. One of `Female`, `Male`, `Non-binary/Third gender`, or `Other`.
-- `ageRangeStart`: Optional. Number (years) from 0 to 120.
-- `ageRangeEnd`: Optional. Number (years) from 0 to 120.
+
+##### Events
 - `dateConfirmed`: **Required**. Date (MM/DD/YYYY) on which the case was
 confirmed.
+- `confirmationMethod`: Optional. Mechanism by which the case was confirmed
+(e.g., `PCR test`).
 - `hospitalized`: Optional. One of `true` or `false`.
 - `dateHospitalized`: Optional. Date (MM/DD/YYYY) on which the patient was
 hospitalized.
+- `icuAdmission`: Optional. One of `true` or `false`.
+- `dateIcuAdmission`: Optional. Date (MM/DD/YYYY) on which the patient was
+admitted to the ICU.
 - `outcome`: Optional. One of either `Death` or `Recovered`.
 - `dateOutcome`: Optional. Date (MM/DD/YYYY) on which the above outcome
 transpired.
+- `dateSymptomOnset`: Optional. Date (MM/DD/YYYY) on which symptoms first
+presented in the case.
+
+##### Symptoms and preexisting conditions
+- `symptomStatus`: Optional. One of `Symptomatic`, `Asymptomatic`, or
+`Presymptomatic`.
+- `symptoms`: Optional. Delimit multiple symptoms with a semicolon (e.g.,
+`cough;fever`).
+- `hasPreexistingConditions`: Optional. One of `true` or `false`.
+- `preexistingConditions`: Optional. Delimit multiple symptoms with a semicolon
+(e.g., `Lyme disease;COPD`).
+
+##### Metadata
 - `caseCount`: Optional. Number greater than 0. If provided, `caseCount` number
 of cases will be inserted into the database: each containing identical values
 for all other columns. Cannot be provided in combination with `sourceEntryId`.
--- a/verification/curator-service/ui/README.md
+++ b/verification/curator-service/ui/README.md
@@ -52,20 +52,20 @@ US).
 [See full definition here](https://en.wikipedia.org/wiki/List_of_administrative_divisions_by_country).
 
 ##### Events
-- `dateConfirmed`: **Required**. Date (MM/DD/YYYY) on which the case was
+- `dateConfirmed`: **Required**. Date (YYYY/MM/DD) on which the case was
 confirmed.
 - `confirmationMethod`: Optional. Mechanism by which the case was confirmed
 (e.g., `PCR test`).
 - `hospitalized`: Optional. One of `true` or `false`.
-- `dateHospitalized`: Optional. Date (MM/DD/YYYY) on which the patient was
+- `dateHospitalized`: Optional. Date (YYYY/MM/DD) on which the patient was
 hospitalized.
 - `icuAdmission`: Optional. One of `true` or `false`.
-- `dateIcuAdmission`: Optional. Date (MM/DD/YYYY) on which the patient was
+- `dateIcuAdmission`: Optional. Date (YYYY/MM/DD) on which the patient was
 admitted to the ICU.
 - `outcome`: Optional. One of either `Death` or `Recovered`.
-- `dateOutcome`: Optional. Date (MM/DD/YYYY) on which the above outcome
+- `dateOutcome`: Optional. Date (YYYY/MM/DD) on which the above outcome
 transpired.
-- `dateSymptomOnset`: Optional. Date (MM/DD/YYYY) on which symptoms first
+- `dateSymptomOnset`: Optional. Date (YYYY/MM/DD) on which symptoms first
 presented in the case.
 
 ##### Symptoms and preexisting conditions
--- a/verification/curator-service/ui/cypress/integration/components/BulkCaseForm.spec.ts
+++ b/verification/curator-service/ui/cypress/integration/components/BulkCaseForm.spec.ts
@@ -23,7 +23,7 @@ describe('Bulk upload form', function () {
         });
     });
 
-    it('Can upload all fields', function () {
+    it.only('Can upload all fields', function () {
         cy.addSource('Bulk source', 'www.bulksource.com');
 
         cy.visit('/cases');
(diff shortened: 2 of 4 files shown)


verification/curator-service/ui/README.md
```diff
@@ -22,8 +22,24 @@ corresponding to this CSV format is available
...
+##### Demographics
+- `gender`: Optional. One of `Female`, `Male`, `Non-binary/Third gender`, or
+`Other`.
+- `ageRangeStart`: Optional. Number (years) from 0 to 120.
+- `ageRangeEnd`: Optional. Number (years) from 0 to 120.
+- `ageRange`: Optional. Convenience field accepting an age range in the format
+`start-end`. If provided in this format, will be used in lieu of
+`ageRangeStart` and `ageRangeEnd`.
+- `ethnicity`: Optional.
+- `nationalities`: Optional. Delimit multiple nationalities with a semicolon
```
khmoran, line 38: should we link to the list of options in github? not necessarily in this PR

verification/curator-service/ui/cypress/integration/components/BulkCaseForm.spec.ts
```diff
@@ -59,20 +59,36 @@ describe('Bulk upload form', function () {
...
         // Outcome
         cy.contains('Recovered');
         cy.contains('2020-6-24');
+
+        // Symptoms
+        cy.contains('Symptomatic');
+        cy.contains('cough, fever');
+
+        // Preexisting conditions
+        cy.contains('Yes');
```
khmoran, line 90: README says true/false -- do we remap to yes/no?

axmb: We do -- this is a boolean in the data model, but the UI converts that to a `Yes` (or presumably `No`) here.

verification/curator-service/ui/README.md
axmb: Yeah -- for many of these fields, there should be that kind of linkage. @acansler is working on a more polished piece of documentation to link from the portal that should take this into account (and is the reason why I didn't go nuts with it in this PR).

verification/curator-service/ui/README.md
```diff
@@ -22,8 +22,24 @@ corresponding to this CSV format is available
...
+##### Case reference
 - `sourceEntryId`: Optional. The UUID provided by the data source to identify
 the specific case.
+
+##### Demographics
+- `gender`: Optional. One of `Female`, `Male`, `Non-binary/Third gender`, or
+`Other`.
+- `ageRangeStart`: Optional. Number (years) from 0 to 120.
+- `ageRangeEnd`: Optional. Number (years) from 0 to 120.
+- `ageRange`: Optional. Convenience field accepting an age range in the format
```
allysonjp715, line 34: What if they have an exact age and not a range?

verification/curator-service/ui/README.md
```diff
@@ -34,17 +50,34 @@ US).
 - `admin3`: Optional. Meaning differs by country (e.g., this is municipality in the
 US).
 [See full definition here](https://en.wikipedia.org/wiki/List_of_administrative_divisions_by_country).
-- `gender`: Optional. One of `Female`, `Male`, `Non-binary/Third gender`, or `Other`.
-- `ageRangeStart`: Optional. Number (years) from 0 to 120.
-- `ageRangeEnd`: Optional. Number (years) from 0 to 120.
+
+##### Events
 - `dateConfirmed`: **Required**. Date (MM/DD/YYYY) on which the case was
```
allysonjp715, line 55: Can we do YYYY/MM/DD for dates to match the rest of the portal?

verification/curator-service/ui/README.md
```diff
@@ -34,17 +50,34 @@ US).
...
 - `hospitalized`: Optional. One of `true` or `false`.
 - `dateHospitalized`: Optional. Date (MM/DD/YYYY) on which the patient was
 hospitalized.
+- `icuAdmission`: Optional. One of `true` or `false`.
+- `dateIcuAdmission`: Optional. Date (MM/DD/YYYY) on which the patient was
+admitted to the ICU.
 - `outcome`: Optional. One of either `Death` or `Recovered`.
 - `dateOutcome`: Optional. Date (MM/DD/YYYY) on which the above outcome
 transpired.
+- `dateSymptomOnset`: Optional. Date (MM/DD/YYYY) on which symptoms first
```
allysonjp715, line 68: There are 2 more event dates in the single case flow: firstClinicalConsultationDate and selfIsolationDate

verification/curator-service/ui/README.md
```diff
@@ -34,17 +50,34 @@ US).
...
+presented in the case.
+
+##### Symptoms and preexisting conditions
+- `symptomStatus`: Optional. One of `Symptomatic`, `Asymptomatic`, or
+`Presymptomatic`.
+- `symptoms`: Optional. Delimit multiple symptoms with a semicolon (e.g.,
```
allysonjp715, line 74: Could you note here that the symptoms are only allowed if the symptomStatus is "Symptomatic" or "Presymptomatic". Do we do that check on the API? We check for it only in the UI of the single case entry

verification/curator-service/ui/README.md
```diff
@@ -34,17 +50,34 @@ US).
...
+`cough;fever`).
+- `hasPreexistingConditions`: Optional. One of `true` or `false`.
+- `preexistingConditions`: Optional. Delimit multiple symptoms with a semicolon
```
allysonjp715, line 77: Similarly here

verification/curator-service/ui/README.md
```diff
@@ -22,8 +22,24 @@ corresponding to this CSV format is available
 [here](https://docs.google.com/spreadsheets/d/1J-C7dq1rNNV8KdE1IZ-hUR6lsz7AdlvQhx6DWp36bjE).
 The following fields are currently incorporated:
 
+##### Case reference
```
allysonjp715, line 25: What about transmission, travel history, genome, other pathogens, and notes sections? Will those be added later?

verification/curator-service/ui/README.md
```diff
@@ -34,17 +50,34 @@ US).
...
 US).
 [See full definition here](https://en.wikipedia.org/wiki/List_of_administrative_divisions_by_country).
-- `gender`: Optional. One of `Female`, `Male`, `Non-binary/Third gender`, or `Other`.
-- `ageRangeStart`: Optional. Number (years) from 0 to 120.
-- `ageRangeEnd`: Optional. Number (years) from 0 to 120.
+
+##### Events
 - `dateConfirmed`: **Required**. Date (MM/DD/YYYY) on which the case was
 confirmed.
+- `confirmationMethod`: Optional. Mechanism by which the case was confirmed
```
allysonjp715, line 57: Can you note that the confirmationMethod is only saved if dateConfirmed is populated? 

Similarly with dateHospitalized, dateIcuAdmission, dateOutcome. We should probably throw an error if those fields are populated but their dependent field isn't populated.

verification/curator-service/ui/src/components/BulkCaseForm.tsx
```diff
@@ -270,21 +315,60 @@ class BulkCaseForm extends React.Component<
...
             ageRangeStart = startEnd[0] ? Number(startEnd[0]) : undefined;
             ageRangeEnd = startEnd[1] ? Number(startEnd[1]) : undefined;
         }
         return { start: ageRangeStart, end: ageRangeEnd };
     }
 
+    createSymptoms(c: RawParsedCase): Symptoms | undefined {
+        let status = c.symptomStatus;
+        if (!status && (c.symptoms || c.dateSymptomOnset)) {
+            status = 'Symptomatic';
```
allysonjp715, line 329: This could be `Presymptomatic` too so we shouldn't fill this in for them

verification/curator-service/ui/README.md
axmb: They can populate `ageRangeStart`/`ageRangeEnd`. From a CSV manipulation standpoint, that's not too much work, since the curator would just duplicate and retitle the column.

verification/curator-service/ui/README.md
axmb: Definitely. This is technically a matter of documentation -- either format is correctly parsed by `new Date()`, which is the format used by the data service. I'll review this with Aaron when he drafts the official doc for this.

verification/curator-service/ui/README.md
axmb: Yep, for now I think we're just including the fields that we've identified to exist in sources we'd plausibly bulk upload. Stephen made a list in #646 which I'm working off of here.

verification/curator-service/ui/README.md
axmb: Stephen suggested in #646 that we'd automatically set `Symptomatic` based on the presence of symptom onset date. I think I'm also setting it if `symptoms` is populated.

It's a good question as to whether we'd enforce these types of relationships in the API. We aren't right now AFAICT, but it might be reasonable for `symptom.status` to be a required field within the `symptoms` stanza.

I don't feel strongly. I was happy to add some convenience autofills to the bulk upload, but we can instead error in some of these situations. Perhaps this is something to talk over (for all upload types) as we tackle the verification CUJ?

verification/curator-service/ui/README.md
axmb: As above, can add later if we choose.

verification/curator-service/ui/README.md
axmb: `dateConfirmed` is required -- but the larger point is spot on. One thing we should tweak in #662 is adding a layer of CSV/meta-validation. We don't have that right now; it's more or less reflecting API errors, which might silently drop data.

verification/curator-service/ui/README.md
axmb: (Same thread as above, I think.)

verification/curator-service/ui/src/components/BulkCaseForm.tsx
axmb: Fair question -- I'd sort of agreed to go along with the direction in #646 on the basis that `Symptomatic` should be far more prevalent, but let's get Stephen's thoughts.

verification/curator-service/ui/README.md
allysonjp715: If I were filling that out and I only filled in ageRangeStart I would assume I was filling in an open ended range e.g. 45+ rather than definitively 45. I think the safest thing to do is also having a dedicated age column, and if they want to enter an age range the form should require them to put in both boundaries on the age range (e.g. 45 and 120 for 45+). That's what we do in the single case form.

verification/curator-service/ui/README.md
allysonjp715: Gotcha. Given that data will be entered worldwide, it will be super common for people to mess it up and enter DD/MM/YYYY. If we require YYYY/MM/DD we can return an error if it's not in the right format, preventing human error.

verification/curator-service/ui/README.md
allysonjp715: Hm interesting... I would think we would want to allow uploading all linelist data using bulk upload. Sources could always add data or new sources could crop up. Not sure why we'd want to restrict it. (Happy for another PR but let's keep an issue open to track)

verification/curator-service/ui/README.md
allysonjp715: When talking with Shru through the single case form we decided against all autofill values. Since the data is of upmost importance, we'd rather add a little inconvenience in the UI to ensure the data is submitted as intended. I think we should remove all autofilled values here too.

verification/curator-service/ui/README.md
axmb: Sorry, to be clear, filling only start or end does result in an open range. An exact age is represented by these two fields containing the same value. I agree that providing an `age` field would make the exact value scenario clear, though to also support open ranges, it'd require 4 columns for age. That's because we want to provide a convenience for the common case of the source containing a single `start-end` entry; the difficulty of getting data into this format is a consideration for bulk that we don't have to make for the single entry form.

Let's poke Aaron on this subject once he drafts his instructions.

verification/curator-service/ui/README.md
axmb: Makes sense. This seems like a great check to add for CSV-format/[meta-validation](https://github.com/globaldothealth/list/issues/662#issuecomment-668585762).

verification/curator-service/ui/README.md
axmb: SG, I've left a comment to get Stephen's take.

verification/curator-service/ui/README.md
axmb: SG, I've opened #718.

verification/curator-service/ui/README.md
allysonjp715: Could you change the docs in the meantime?

verification/curator-service/ui/README.md
allysonjp715: I'd prefer to remove all the autofill added in this PR, since I think it's highly likely we won't want to add it. But if you promise to come back to it I suppose it's okay ;)

verification/curator-service/ui/README.md
axmb: Sure thing, done.

verification/curator-service/ui/README.md
axmb: No trouble at all. We can toss it for now, and update later as desired. I'll change these instances to align with existing checks ("silently don't include nonsensical data but we'll make it error later").

PR #709, opened 2020-08-03
Set revision metadata on data server for batch upserts

khmoran:
| 10k cases |   | 50k cases |  
-- | -- | -- | -- | --
  | at head | with change | at head | with change
create | 13.93 | 16.96 | 72 | 96
update | 15.76 | 22.96 | 336 | 588


data-serving/data-service/src/controllers/case.ts
```diff
@@ -181,15 +181,34 @@ const findCaseIdsWithCaseReferenceData = async (
...
-                  .or(providedCaseReferenceData)
-                  .lean()
-                  .exec()
-            : [];
-    return result.map((res) => String(res['_id']));
+
+    return providedCaseReferenceData.length > 0
+        ? Case.find()
+              .or(providedCaseReferenceData)
+              .select({ _id: 1, caseReference: 1, revisionMetadata: 1 })
```
axmb, line 188: Did the `lean()` not make a difference, here? I admit I didn't conclusively check that it altered query behavior if a `select` clause was present. I expect it did, and that with `lean()`, this won't return a `CaseDocument`.

data-serving/data-service/src/controllers/preprocessor.ts
```diff
@@ -61,16 +62,42 @@ export const setRevisionMetadata = async (
 ) => {
     const curatorEmail = request.body.curator.email;
 
-    // Find the case if it already exists so we can update its existing
+    // Find the case(s) if it/they already exists so we can update existing
     // metadata.
-    const c = await getCase(request);
+    if (request.body.cases) {
```
axmb, line 67: Might be simpler to have a separate preprocessor, but just a nit/thought.

data-serving/data-service/src/controllers/preprocessor.ts
```diff
@@ -61,16 +62,42 @@ export const setRevisionMetadata = async (
...
+        // Batch upsert.
+        const existingCases = await findCasesWithCaseReferenceData(request);
```
axmb, line 69: Small question (don't worry about it if it's a PITA to measure) -- did we confirm that the MongoDB call was what made this slow? Seems very likely, but I wonder whether these in-memory operations took much time in comparison for the large (50k) invocations.

data-serving/data-service/test/controllers/preprocessor.test.ts
```diff
@@ -342,3 +342,74 @@ describe('upsert', () => {
         );
     });
 });
+describe('batch upsert', () => {
+    it('sets create metadata', async () => {
```
axmb, line 346: hyper-nit: sets create and update metadata

data-serving/data-service/src/controllers/preprocessor.ts
khmoran: yeah I went back and forth on that, I think you're probably right -- I'll split them

data-serving/data-service/src/controllers/case.ts
khmoran: I've run it a few times and it doesn't seem to have helped (if anything, it made it a little worse -- though I am not doing this very scientifically!)

Since it doesn't seem to help much, I think I'd prefer to keep it in for the type safety. WDYT?

data-serving/data-service/src/controllers/preprocessor.ts
khmoran: nope -- I didn't do full profiling of where the bottleneck was, but from monitoring the server logs, it looks like a *lot* of the slowdown was MongoDB. It seemed like its page size was ~100 docs, and it was paging slowly. Not sure if there's some optimization we can do there?

data-serving/data-service/src/controllers/case.ts
axmb: SG. FMI, does it return a `CaseDocument` with only the `select` fields populated?

data-serving/data-service/src/controllers/preprocessor.ts
axmb: SG. I bet we'd see a speedup if we set `ordered` false in the `bulkWrite`. For some reason I'd internalized that it was false by default, but that's not the case. I can check that out separately.

khmoran: good idea!

data-serving/data-service/src/controllers/case.ts
khmoran: yep exactly!

PR #711, opened 2020-08-04
Remove some warnings from UI

allysonjp715:
Warning example: ` You have provided an out-of-range value 'undefined' for the select (name="traveledPrior30Days") component`


verification/curator-service/ui/src/components/CaseForm.tsx
```diff
@@ -260,7 +261,7 @@ function hasErrors(fields: string[], errors: any, touched: any): boolean {
 }
 
 function unknownToUndefined(value: string | undefined): string | undefined {
```
attwad, line 263: rename to unknownOrEmptyToUndefined ?

allysonjp715: Done

PR #717, opened 2020-08-04
Parser lambda for zurich confirmed cases

attwad:
For #483

Submitting this for review for now but shouldn't be used as is as no deduplication is done yet.
Will work on that yet.


ingestion/functions/parsing/ch_zurich/zurich.py
```diff
@@ -0,0 +1,101 @@
...
+_AGE_INDEX = 2
+_GENDER_INDEX = 3
+_CONFIRMED_INDEX = 4
+_SOURCE_INDEX = 6
+
+def convert_date(raw_date):
+    """
+    Convert raw date field into a value interpretable by the dataserver.
+
+    The date is listed in YYYY-mm-dd format, but the data server API will
```
axmb, line 38: This doesn't need to be converted -- YYYY-mm-dd is correctly interpreted by JS's Date (on the data server). You can verify this in console. The only ambiguous case, AFAICT, is dd-mm-YYYY.

ingestion/functions/parsing/ch_zurich/zurich.py
```diff
@@ -0,0 +1,101 @@
...
+    assume that ambiguous cases (e.g. "05/06/2020") are in mm/dd/YYYY format.
+    """
+    date = datetime.strptime(raw_date, "%Y-%m-%d")
+    return date.strftime("%m/%d/%YZ")
+
+def convert_gender(raw_gender: str):
```
axmb, line 44: I dig the types -- use elsewhere (or remove)?

ingestion/functions/parsing/ch_zurich/zurich.py
```diff
@@ -0,0 +1,101 @@
...
+                            "name": "confirmed",
+                            "dateRange":
+                            {
+                                "start": convert_date(row[_DATE_INDEX]),
+                                "end": convert_date(row[_DATE_INDEX]),
+                            },
+                        },
+                    ],
+                    "demographics": {
+                        "gender": convert_gender(row[_GENDER_INDEX]),
```
axmb, line 87: Consider organizing this and age into a convert_demographics. If we don't have gender nor age, we'll end up with `demographics: {}` rather than `demographics: None` (which gets dropped from the HTTP request).

ingestion/functions/parsing/ch_zurich/zurich.py
```diff
@@ -0,0 +1,101 @@
...
+           to dedupe.
+        2. We can't link confirmed cases and confirmed deaths because of (1)
+           so we're only importing confirmed cases and ignoring deaths.
+    """
+    with open(raw_data_file, "r") as f:
+        reader = csv.reader(f)
+        next(reader) # Skip the header.
+        cases = []
+        for row in reader:
+            num_cases = int(row[_CONFIRMED_INDEX])
```
axmb, line 66: nit: `case_index` or `num_case`?

ingestion/functions/parsing/ch_zurich/zurich_test.py
```diff
@@ -0,0 +1,57 @@
...
+            "ageRange": {
+                "start": 33,
+                "end": 33
+            },
+            "gender": "Female"
+        },
+    })
+
+
+class CHZurichTest(unittest.TestCase):
```
axmb, line 50: Does this play nice for runs of `pytest` at the top-level `ingestion/functions` dir? Is there a display/functional difference between this and vanilla test files as for india?

ingestion/functions/parsing/ch_zurich/zurich_test.py
```diff
@@ -0,0 +1,57 @@
...
+    def test_parse(self):
+        self.maxDiff = 5000
```
axmb, line 52: Unused (or is there :mage: at play)?

ingestion/functions/parsing/ch_zurich/zurich.py
attwad: good call, done.

ingestion/functions/parsing/ch_zurich/zurich.py
attwad: added in other places.

ingestion/functions/parsing/ch_zurich/zurich.py
attwad: ohhh, interesting I didn't know.
Okay then less code is better.
Removed.

ingestion/functions/parsing/ch_zurich/zurich_test.py
attwad: that's just so I can use self.assertCountEqual which shows much better output than a pure assert.
Those tests are still discovered by pytest, I just run it the same way as I do for the others:

```
python3.8 -m pytest .
============================================================= test session starts =============================================================
platform darwin -- Python 3.8.5, pytest-6.0.1, py-1.9.0, pluggy-0.13.1
rootdir: /Users/timothe/healthmap/healthmap-gdo-temp/ingestion/functions, configfile: pytest.ini
plugins: requests-mock-1.8.0
collected 19 items

parsing/ch_zurich/zurich_test.py .                                                                                                      [  5%]
parsing/common/python/parsing_lib/parsing_lib_test.py .......                                                                           [ 42%]
parsing/india/india_test.py .                                                                                                           [ 47%]
retrieval/retrieval_test.py ..........                                                                                                  [100%]

============================================================= 19 passed in 4.76s ==============================================================
```

ingestion/functions/parsing/ch_zurich/zurich_test.py
attwad: when there was a diff in the output, pytest was complaining that it was larger than the 1500 chars that's the default, I increased it so that if a diff comes up it is shown.

ingestion/functions/parsing/ch_zurich/zurich.py
attwad: that's the number of confirmed cases, so I renamed to num_confirmed_cases, hopefully that's clearer.

axmb: Ah, interesting. Hadn't seen data demarcated in this way, but I understand.

ingestion/functions/parsing/ch_zurich/zurich_test.py
axmb: Got it, thanks!

ingestion/functions/parsing/ch_zurich/zurich_test.py
axmb: Gotcha. Does `Testcase.maxDiff` work (in the class, above this method)? Just for the sake of clarity.

attwad: not sure, the error message said to set self.maxDiff so that's what I did, I think I'd stick to that (considering other test cases might not need that big diff num).

PR #722, opened 2020-08-04
Add some styles to app menu

allysonjp715:
Screen Shot 2020-08-04 at 4 57 23 PM

### DIFF
--- a/verification/curator-service/ui/src/components/App.tsx
+++ b/verification/curator-service/ui/src/components/App.tsx
@@ -51,6 +51,16 @@ const theme = createMuiTheme({
             main: '#000000',
         },
     },
+    overrides: {
+        MuiListItem: {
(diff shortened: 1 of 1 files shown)


axmb:
Semi-related: are the styles in Figma ~final? Some frames show new G.h-themed styling, but I'm not sure the extent of that nor how much it will differ from what's present.

allysonjp715:
> Semi-related: are the styles in Figma ~final? Some frames show new G.h-themed styling, but I'm not sure the extent of that nor how much it will differ from what's present.

No, the styles in Figma aren't final. Shru is working on getting us styling specs as we speak. I was venturing a bet that the styling here will remain the same, with probably only the color changing. Certainly it's good to have some indication of a selected menu item, even if the css of it changes.

Issue #723, opened 2020-08-05
Support CSV as source format


attwad:
I want to add a parser for Zurich in #483 but it seems only JSON is supported, need to add CSV as well.

PR #725, opened 2020-08-05
Support CSV as an automated ingestion source

attwad:
For #723  and #483

This basically exposes the "format" field in the sources table in the UI.

### DIFF
--- a/verification/curator-service/ui/src/components/SourceTable.tsx
+++ b/verification/curator-service/ui/src/components/SourceTable.tsx
@@ -45,7 +45,7 @@ interface Automation {
 interface Source {
     _id: string;
     name: string;
-    format: string;
+    format?: string;
     origin: Origin;
     automation?: Automation;
 }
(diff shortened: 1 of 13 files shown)


verification/curator-service/ui/src/components/SourceTable.tsx
```diff
@@ -294,6 +298,7 @@ class SourceTable extends React.Component<Props, SourceTableState> {
                                     />
                                 ),
                             },
+                            { title: 'Format', field: 'format' },
```
axmb, line 301: How about a dropdown with `CSV` and `JSON`?

attwad:
> Excellent, thanks. This was needed this sprint.
> 
> Why remove the API spec enum?

We also want to support the "empty" format source for manual entries in the case form I don't think this works with an enum.

verification/curator-service/ui/src/components/SourceTable.tsx
attwad: Good idea, let me send you a separate PR for that now.

PR #728, opened 2020-08-05
make json csv dropdown for format edit field

attwad:
For #723

### DIFF
--- a/verification/curator-service/ui/cypress/integration/components/SourceTableTest.spec.ts
+++ b/verification/curator-service/ui/cypress/integration/components/SourceTableTest.spec.ts
@@ -12,7 +12,8 @@ describe('Sources table', function () {
         cy.get('button[title="Add"]').click();
         cy.get('input[placeholder="Name"]').clear().type('Example source');
         cy.get('input[placeholder="URL"]').clear().type('www.example.com');
-        cy.get('input[placeholder="Format"]').clear().type('JSON');
+        cy.get('div[data-testid="format-select"]').click();
+        cy.contains('li', 'JSON').click();
         cy.get('button[title="Save"]').click();
 
         cy.contains('Example source');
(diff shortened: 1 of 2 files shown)


verification/curator-service/ui/cypress/integration/components/SourceTableTest.spec.ts
```diff
@@ -12,7 +12,8 @@ describe('Sources table', function () {
         cy.get('button[title="Add"]').click();
         cy.get('input[placeholder="Name"]').clear().type('Example source');
         cy.get('input[placeholder="URL"]').clear().type('www.example.com');
-        cy.get('input[placeholder="Format"]').clear().type('JSON');
+        cy.get('div[data-testid="format-select"]').click();
+        cy.contains('li', 'JSON').click();
```
axmb, line 16: This only really matters if there's another `li` with JSON in the DOM, but I think we can chain the `.contains[...]` off the previous `.click()`.

attwad: unfortunately no, the new divs are not children of the format-select one :( it's this annoying "portal" elements thing that I can't seem to disable for selects.

attwad:
> From the previous PR:
> 
> > > Why remove the API spec enum?
> 
> > We also want to support the "empty" format source for manual entries in the case form I don't think this works with an enum.
> 
> Can we save that as `undefined`, instead of `''`?

that's my understanding of how it works currently, when you specify no format it will be undefined.
Yup confirmed locally, created one with a format unspecified and then got:
`{"_id":"5f2ada27df34a7051c8b01fd","name":"test","origin":{"_id":"5f2ada27df34a7051c8b01fe","url":"test2"},"__v":0}` when querying it.

PR #733, opened 2020-08-06
Add basic monitoring with appmetrics-dash

axmb:
I'm not in love with the package, but it's pretty solid (more or less) out of the box.

Seems to be the best solution short of investing in Prometheus/Grafana.

appmetrics

### DIFF
--- a/data-serving/data-service/src/index.ts
+++ b/data-serving/data-service/src/index.ts
@@ -1,10 +1,10 @@
 // Set up appmetrics-dash before importing additional dependencies.
 // This ensures that the module captures metrics for dependent systems, like
 // MongoDB.
-import express from 'express';
-const app = express();
 import Dash from 'appmetrics-dash';
-Dash.attach();
+if (process.env.NODE_ENV !== 'test') {
+    Dash.attach();
+}
 
 import * as caseController from './controllers/case';
 import * as homeController from './controllers/home';
@@ -21,9 +21,12 @@ import { OpenApiValidator } from 'express-openapi-validator';
 import YAML from 'yamljs';
 import bodyParser from 'body-parser';
 import dotenv from 'dotenv';
+import express from 'express';
 import mongoose from 'mongoose';
 import swaggerUi from 'swagger-ui-express';
 import validateEnv from './util/validate-env';
+
+const app = express();
 
 dotenv.config();
 const env = validateEnv();
--- a/data-serving/data-service/src/index.ts
+++ b/data-serving/data-service/src/index.ts
@@ -1,3 +1,11 @@
+// Set up appmetrics-dash before importing additional dependencies.
+// This ensures that the module captures metrics for dependent systems, like
+// MongoDB.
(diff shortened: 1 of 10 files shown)


data-serving/data-service/src/index.ts
```diff
@@ -1,3 +1,11 @@
+// Set up appmetrics-dash before importing additional dependencies.
+// This ensures that the module captures metrics for dependent systems, like
+// MongoDB.
+import Dash from 'appmetrics-dash';
+if (process.env.NODE_ENV !== 'test') {
+    Dash.attach();
```
attwad, line 6: can we move the import in that if as well to avoid importing it if we don't need to?

verification/curator-service/api/src/index.ts
```diff
@@ -1,3 +1,11 @@
+// Set up appmetrics-dash before importing additional dependencies.
+// This ensures that the module captures metrics for dependent systems, like
+// MongoDB.
+import Dash from 'appmetrics-dash';
+if (process.env.NODE_ENV !== 'test') {
+    Dash.attach();
```
attwad, line 6: ditto here

data-serving/data-service/src/index.ts
axmb: The Typescript compiler is smart enough to do this, IIUC:

https://www.typescriptlang.org/docs/handbook/modules.html#optional-module-loading-and-other-advanced-loading-scenarios

verification/curator-service/api/src/index.ts
axmb: (See above.)

axmb:
> FYI I had to follow [nodejs/node-gyp#1927 (comment)](https://github.com/nodejs/node-gyp/issues/1927#issuecomment-661825834) to be able to install it on my macbook but then I couldn't build the docker images anymore for some reason to run the stack locally (E404 when building the data service image on the typescript-eslint/eslint-plugin@^2.34.0 version). We should probably update those deps it's 3.8.0 now... anyway that's for another PR.
> 
> Question: do we want to expose this externally as well? (seems to be the default), there's no mutate API right it's all read only? We could also use the options to expose() to serve on a different port and have to do `kubectl proxy` if we want to check the stats if we're worried about exposing this externally.

It's not totally read only. You can do things like trigger profiling, node reports, and heap dumps from the UI. It'd be better to have it secured, but it'd make access somewhat less convenient.

WDYT?

attwad:
Uhg, heap dumps and such freely available is asking to be DOS'ed. Can we perhaps expose the handler only to admins?

axmb:
> Uhg, heap dumps and such freely available is asking to be DOS'ed. Can we perhaps expose the handler only to admins?

No worries. Turns out there's a convenience option in the package that accepts middleware, so super simple to lock it down in this way.

I didn't modify the installation in the data service, since IIUC, we don't expose that pod in dev/prod.

PR #735, opened 2020-08-06
use batchUpsert in parsing api

### LINKED ISSUE #734: Use batch upsert in parsing lambdas common library
Otherwise I can't test locally with 4000 cases, seems more appropriate to use it anyway.

attwad:
Fixes #734


ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
```diff
@@ -43,21 +43,13 @@ def write_to_server(cases, headers):
     Upserts the provided cases via the G.h Case API.
 
     TODO: Link out to the Case API resource documentation, once available.
```
axmb, line 45: This TODO is also closable (though the link won't be for prod).

ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
```diff
@@ -43,21 +43,13 @@ def write_to_server(cases, headers):
...
-                         headers=headers).raise_for_status()
-            count_success += 1
-        except Exception as e:
-            print(e)
-            count_error += 1
-    return count_success, count_error
+    res = requests.post(put_api_url, json={"cases": cases},
+                       headers=headers)
+    res_json = res.json()
+    return len(res_json["createdCaseIds"]), len(res_json["updatedCaseIds"])
```
axmb, line 52: We probably want to surface some indication of the errors, here.

- Error count, as we're currently doing, though the meaning will change (since we could have up to `n` validation errors, but only 1 server error for the single API call).
- Index of cases that produced the errors.
- Raw errors.

WDYT?

ingestion/functions/parsing/common/python/parsing_lib/parsing_lib_test.py
```diff
@@ -1,4 +1,7 @@
-
+# Tests for the parsing library common logic.
+# If you come from a unittest background and wonder at how requests_mock
+# arrives by magic here, check out
```
axmb, line 3: It's :mage: all the way down.

ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
attwad: removed.

ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
attwad: I'm all for it but I don't see any way to surface granular errors at this point, the curator service just proxies to the data service and the data service batch upsert just looks at what's returned by bulkWrite:
http://mongodb.github.io/node-mongodb-native/3.1/api/Collection.html#~BulkWriteOpResult

that doesn't contain errors or anyting that would make it possible to return what you proposed above, am I missing something?

ingestion/functions/parsing/common/python/parsing_lib/parsing_lib_test.py
attwad: yeah this pytest magic is real magic, at least with unittest you setup your mocks before each tests so it's obvious where it comes from.

ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
attwad: I will merge this and we can talk about errors in a future PR.

PR #747, opened 2020-08-07
extract source id from event instead of s3 path

### LINKED ISSUE #745: Parsing lambda: stop extracting the source ID from the s3 bucket URL and take it from the event instead
We might move to something else than s3 where uuids cannot be used as path parts for example, it would also allow us to test on s3 files that aren't necessarily respecting the uuid/some/other/path.json format.

attwad:
Fixes #745

### DIFF
--- a/ingestion/functions/parsing/ch_zurich/input_event.json
+++ b/ingestion/functions/parsing/ch_zurich/input_event.json
@@ -1,5 +1,6 @@
 {
     "s3Bucket": "epid-sources-raw",
+    "sourceId": "5f2acbe8df34a7051c8b01f4",
     "s3Key": "5f2acbe8df34a7051c8b01f4/2020/08/05/1806/content.csv",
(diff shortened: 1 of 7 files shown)


ingestion/functions/parsing/ch_zurich/input_event.json
```diff
@@ -1,5 +1,6 @@
 {
     "s3Bucket": "epid-sources-raw",
+    "sourceId": "5f2acbe8df34a7051c8b01f4",
```
axmb, line 3: hyper-nit, but can we put this after the s3 params throughout (i.e. here and in the retrieval file method signature below)?

attwad: I changed here but in the python code it feels better close next to the sourceURL as it is now rather than having source url,s3 bucket,s3 key,source id

PR #749, opened 2020-08-07
Update deps (via ./dev/update_deps.sh).

axmb:


attwad:
We should probably try major version updates as well at some point, this is fine though

axmb:
I'll need to manually downgrade material-table, due to https://github.com/mbrn/material-table/issues/2269.

PR #750, opened 2020-08-07
Workflow to export prod db data daily and commit to repo

khmoran:
The secret I'm using is my own PAT (with public_repo scope), and we should transition this to one of the future maintainers after I confirm this works

I tested it locally with `act` but it probably won't work the first time. Never has, never will!

```
act -j export-case-data-prod -P ubuntu-latest=nektos/act-environments-ubuntu:18.04-full -s DB_CONNECTION_URL=<CONN_STRING> -s EXPORT_DATA_ACCESS_TOKEN=<GITHUB_TOKEN>
```

### DIFF
--- a/.github/workflows/case-data-export-prod.yml
+++ b/.github/workflows/case-data-export-prod.yml
@@ -1,0 +1,38 @@
+name: Export prod case data
+
+on:
+  schedule:
+    # Every day at 6AM UTC (data is typically ingested at 5AM UTC).
+    - cron: "0 6 * * *"
+
+jobs:
+  export-case-data-prod:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Install MongoDB CLIs
+        run: |
+          wget -qO - https://www.mongodb.org/static/pgp/server-4.2.asc | sudo apt-key add -
+          echo "deb [ arch=amd64,arm64 ] https://repo.mongodb.org/apt/ubuntu xenial/mongodb-org/4.2 multiverse" | sudo tee /etc/apt/sources.list.d/mongodb-org-4.2.list
+          sudo apt-key adv --keyserver keyserver.ubuntu.com --recv-keys 5DC22404A6F9F1CA 656408E390CFB1F5
+          sudo apt-get update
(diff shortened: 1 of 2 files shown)


.github/workflows/case-data-export-prod.yml
```diff
@@ -0,0 +1,38 @@
+name: Export prod case data
+
+on:
+  schedule:
+    # Every day at 6AM UTC (data is typically ingested at 5AM UTC).
```
axmb, line 5: This is done as an action in a different repo, right? If it's the same repo, we can trigger this off of the newly added `workflow_run`.

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run

khmoran: oh cool! we could use that but given that we're probably going to stop running that workflow starting next week (whenever we drop the data from mongodb & stop re-importing), it might not make sense. ultimately this is really "arbitrarily running once a day at 6am because why not"

axmb: SGTM

attwad:
Hey I just stumbled upon this by accident looking at my inbox late at night, you can blame me for that.
But when I was looking at a workflow that updates its own repo for the nCov2019 repo where the spreadsheets are currently processed I came across a few actions that were doing that instead of having to do some manual git trickery:
https://github.com/marketplace/actions/github-push, might be worth looking at it.

khmoran:
> Hey I just stumbled upon this by accident looking at my inbox late at night, you can blame me for that.
> But when I was looking at a workflow that updates its own repo for the nCov2019 repo where the spreadsheets are currently processed I came across a few actions that were doing that instead of having to do some manual git trickery:
> https://github.com/marketplace/actions/github-push, might be worth looking at it.

thanks Tim, modified to use this!

khmoran:
@lmanul I'm going to merge this so that we can start getting these running so I can debug inevitable issues, but happy to make changes based on feedback!

lmanul:
Is this running regularly? Can I find the most recent data somewhere? I see an archive here but it's 11 days old:

    https://github.com/globaldothealth/list/tree/c6ff15fda049a86aebbe6401aebac78e0edb6186/data

PR #752, opened 2020-08-07
Add data model for uploads.

axmb:


verification/curator-service/api/src/model/upload-summary.ts
```diff
@@ -0,0 +1,18 @@
+import mongoose from 'mongoose';
+
+export const uploadSummarySchema = new mongoose.Schema(
+    {
+        numCreated: Number,
+        numUpdated: Number,
+        error: String,
```
attwad, line 7: just to make sure, we won't have multiple errors? we stop at the first one?

verification/curator-service/api/src/model/upload.ts
```diff
@@ -0,0 +1,28 @@
...
+        type: String,
+        required: true,
+    },
+    summary: {
+        type: uploadSummarySchema,
+        required: true,
+    },
+    created: {
+        type: Date,
+        default: Date.now(),
```
attwad, line 16: should this be `Date.now` instead of `Date.now()` ? (aka. callback vs raw date value)

verification/curator-service/api/src/model/upload-summary.ts
axmb: We might have multiple case validation errors; I'm intending this as the top-level digested error for the upload. E.g.:

- `Retrieval failed: source not found at URL [...].`
- `Parsing failed: fatal runtime error.`

Haven't quite worked out what we'll do for all the detailed per-case errors (e.g., could store them in a separate field and display them on some sort of expansion in the UI). Can tackle that afterwards.

verification/curator-service/api/src/model/upload.ts
axmb: Yep, typo on my part, good catch.

PR #757, opened 2020-08-10
nicify profile page using chips and tooltips over roles

attwad:
image

### DIFF
--- a/verification/curator-service/ui/cypress/integration/components/ProfileTest.spec.ts
+++ b/verification/curator-service/ui/cypress/integration/components/ProfileTest.spec.ts
@@ -10,6 +10,7 @@ describe('Profile', function () {
 
         cy.contains('Alice Smith');
         cy.contains('alice@test.com');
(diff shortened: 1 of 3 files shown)


verification/curator-service/ui/cypress/integration/components/ProfileTest.spec.ts
```diff
@@ -10,6 +10,7 @@ describe('Profile', function () {
 
         cy.contains('Alice Smith');
         cy.contains('alice@test.com');
-        cy.contains('reader, curator');
+        cy.contains('reader');
```
axmb, line 13: If you want to test for it being in a chip, you could `get` rather than `contains`. No biggie, though.

attwad: nah I don't want the test to break if we start showing "admin" in a card or whatever next, contains is fine.

PR #760, opened 2020-08-10
Add upload ID to the case data model

axmb:
Not making it required, for now. We could consider requiring this, and having it be ~unused for manual (and perhaps bulk) entry, but doesn't seem worthwhile at the moment.

### DIFF
--- a/data-serving/data-service/src/model/case-reference.ts
+++ b/data-serving/data-service/src/model/case-reference.ts
@@ -11,6 +11,7 @@ export const caseReferenceSchema = new mongoose.Schema(
             type: String,
             required: true,
         },
+        uploadId: String,
         additionalSources: [
             {
                 sourceUrl: String,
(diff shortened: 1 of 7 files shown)


data-serving/data-service/src/model/case-reference.ts
```diff
@@ -31,6 +32,9 @@ export type CaseReferenceDocument = mongoose.Document & {
     /** The URL of the source of the case data at the time of ingestion. */
     sourceUrl: string;
 
+    /** The UUID of the upload by which this case was entered into the DB. */
```
khmoran, line 35: maybe make some reference to a batch of cases? upload to me could be single, batch, etc

axmb: Sounds good. I'll clarify with a description in the OpenAPI spec as well.

PR #761, opened 2020-08-10
Naive implementation of creating case revisions on batch upsert

khmoran:
| 10k cases |   | 50k cases |  
-- | -- | -- | -- | --
  | at head | with revisions | at head | with revisions
create | 20.69 | 19.29 | 120 | 168
update | 27.86 | 48.36 | 636 | OOM

we run out of heap at 50K here because we have to read full cases into memory to write them to the other collection. we can possibly mitigate this by a more sophisticated impl with batching the batches?


axmb:
Since the revisions aren't really part of the response, is it possible for us to execute this work _after_ `Response.send()`? I don't think there's anything like this in Mongoose (though that'd be great), but failing that, does just swapping this middleware invocation with the controller response handler in `index` do the trick?

khmoran:
> Since the revisions aren't really part of the response, is it possible for us to execute this work _after_ `Response.send()`? I don't think there's anything like this in Mongoose (though that'd be great), but failing that, does just swapping this middleware invocation with the controller response handler in `index` do the trick?
> 
> Edit: swapping won't work out the box, because we're writing the current data version, but I wonder if it's possible to either:
> 
> 1. Enqueue the work for after the request is handled in the existing middleware, or,
> 2. Snag the current data before the response handler in a read middleware, then invoke the actual MongoDB write after the response is sent in a write middleware. Simplest way to do this, if we have memory, is to store it in `req.locals` in the read middleware. This is assuming the bulk of the slowdown here is in the write portion.

I'm happy to play around with this a bit more but I think there are a couple considerations:
1. Is the slowdown the biggest concern here or the OOMing? (Of course we can try to fix both, but for the sake of prioritizing.)
2. Do we want to move ahead with this as an 'it works, but it's sub-optimal' solution and leave tuning it as a P1 or P2?

data-serving/data-service/src/controllers/preprocessor.ts
```diff
@@ -128,3 +135,27 @@ export const createCaseRevision = async (
...
+) => {
+    const casesToUpsert = (await findCasesWithCaseReferenceData(request)).map(
+        (c) => {
+            return {
+                case: c,
+            };
+        },
+    );
+
+    await CaseRevision.insertMany(casesToUpsert, {
```
axmb, line 152: Yeah, it _feels like_ there's a MongoDB operation that prevents us from placing these cases into memory by chaining these two operations, but perhaps not worth sweating too much over.

khmoran: I agree. :\ I looked for a `copy` or `clone` but couldn't find anything. But there very well could be something out there that I didn't find -- either way, there's definitely room for improvement. 

I filed #784 for this, but let me add that as something to investigate further

axmb: All good, thanks!

PR #763, opened 2020-08-11
Implement filtering of dates in parsing lib and add date filter columns to the UI

attwad:
For #744 

Wanted to do this in two PRs but I ended up merging the UI in here by accident sorry about that :( I hope you don't mind reviewing this whole together (might make more sense actually).

- The UI bit is adding 2 columns in the sources table, the cypress tests refuse to pass I have no idea why I tried everything really... but it seems that one can't select the last column whatever I do to it (I tried moving the JSON/CSV bit there and it fails as well so it seems position-related...). I verified that one could add/edit/remove date filter options locally though.
- The API bit is moving the dateFilter in the source model instead of automation as we also want to be able to run on specific dates without being automated (for testing for example and before setting up automation).

### DIFF
--- a/ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
+++ b/ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
@@ -94,6 +94,8 @@ def filter_cases_by_date(case_data, date_filter):
             return delta_days == 0
         elif op == "LT":
             return delta_days < 0
+        else:
+            raise ValueError(f'Unsupported date filter operand: {op}')
 
     return [case for case in case_data if case_is_within_range(case, cutoff_date, op)]
             
--- a/ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
+++ b/ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
@@ -49,6 +49,9 @@ def write_to_server(cases, headers):
     res = requests.post(put_api_url, json={"cases": cases},
                        headers=headers)
     res_json = res.json()
+    if res.status_code != 200:
+        raise RuntimeError(f'Error sending cases to server, status={res.status_code}, response={res_json}')
+    # TODO: Look for "errors" in res_json and handle them in some way.
     return len(res_json["createdCaseIds"]), len(res_json["updatedCaseIds"])
 
 
@@ -79,6 +82,11 @@ def get_today():
     return datetime.datetime.today()
 
 def filter_cases_by_date(case_data, date_filter):
+    """Filter cases according ot the date_filter provided.
+
+    Returns the cases that matched the date filter or all cases if
+    no filter was requested.
+    """
     if not date_filter:
         return case_data
     now = get_today()
@@ -140,7 +148,6 @@ def run_lambda(event, context, parsing_function):
         raw_data_file, source_id,
         source_url)
     api_creds = obtain_api_credentials()
-    case_data = filter_cases_by_date(case_data, date_filter)
     count_created, count_updated = write_to_server(
-        case_data, api_creds)
+        filter_cases_by_date(case_data, date_filter), api_creds)
     return {"count_created": count_created, "count_updated": count_updated}
--- a/ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
+++ b/ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
@@ -1,3 +1,4 @@
+import datetime
 import os
 import tempfile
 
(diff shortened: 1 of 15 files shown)


ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
```diff
@@ -47,6 +49,9 @@ def write_to_server(cases, headers):
     res = requests.post(put_api_url, json={"cases": cases},
                        headers=headers)
     res_json = res.json()
+    if res.status_code != 200:
+        raise RuntimeError(f'Error sending cases to server, status={res.status_code}, response={res_json}')
```
axmb, line 53: For a later PR to be sure, but we should consolidate the error handling to a helper method. I imagine we'll need to do some writes to the Upload API (soon(tm)) as part of error handling.

verification/curator-service/api/openapi/openapi.yaml
```diff
@@ -580,6 +580,18 @@ components:
               type: string
         format:
           type: string
+        dateFilter:
+          type: object
+          properties:
+            numDaysBeforeToday:
+              type: number
+              # Must be positive as we subtract it to "today".
```
axmb, line 588: What if we want "only today's data"? Or is that just too risky, such that we wouldn't even allow it?

verification/curator-service/api/src/model/source.ts
```diff
@@ -8,6 +8,24 @@ import { OriginDocument, originSchema } from './origin';
 import mongoose from 'mongoose';
 import { uploadSchema, UploadDocument } from './upload';
 
+const dateFilterSchema = new mongoose.Schema({
```
axmb, line 11: Can you put this in a separate file, like the remainder of sub-docs?

Having all of them in here is a mess, and it's simpler to have all of 'em out than have some included here.

Can do a separate PR if you prefer!

verification/curator-service/ui/cypress/integration/components/SourceTableTest.spec.ts
```diff
@@ -10,13 +10,23 @@ describe('Sources table', function () {
...
 
         cy.get('button[title="Add"]').click();
-        cy.get('input[placeholder="Name"]').clear().type('Example source');
-        cy.get('input[placeholder="URL"]').clear().type('www.example.com');
+        cy.get('input[placeholder="Name"]').type('Example source');
+        cy.get('input[placeholder="URL"]').type('www.example.com');
         cy.get('div[data-testid="format-select"]').click();
         cy.contains('li', 'JSON').click();
+        /* This doesn't work and I HAVE NO IDEA WHY PLEASE HELP.
+        cy.get('div[data-testid="op-select"]').scrollIntoView().click();
```
axmb, line 18: Use the `.click({ force: true })`, Luke?

verification/curator-service/ui/src/components/SourceTable.tsx
```diff
@@ -331,18 +356,16 @@ class SourceTable extends React.Component<Props, SourceTableState> {
                                         onChange={(event): void =>
                                             props.onChange(event.target.value)
                                         }
-                                        defaultValue={props.value}
+                                        defaultValue={props.value || ''}
```
axmb, line 359: What's the background for this change? Suppose this question also applies to having the empty string for `dateFilter.op`. Though in that case, we're adding an enum to the API, which is what's important. The empty string doesn't really cause any harm if we can prevent sending that data (replacing it with undefined instead) from the UI.

attwad: The purpose is getting rid of warnings showing up in the console, undefined is not a valid value (must be a valid string). Did the same for the JSON/CSV select as well now we have no more warnings about those.

verification/curator-service/ui/cypress/integration/components/SourceTableTest.spec.ts
attwad: Tried that (was even suggested by cypress in the error messages), but it didn't help.

verification/curator-service/api/openapi/openapi.yaml
attwad: We could change that minimum value to 0 to allow that with the EQ operator but honestly yeah it feels pretty risky.

verification/curator-service/api/src/model/source.ts
attwad: Sure, done.

ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
attwad: yup, definitely, there is the option of calling raise_for_status() when doing the post which will raise inc ase of an error status, here I chose to check for the status myself, we could have the upload api path handled in the exception handling code or here in a helper function, we'll see when that time comes I guess.

Issue #634, opened 2020-07-27
Manual entry: Occupation field, quick complete options


sratcliffe118:
Feature request: Provide the 10 most popular occupations in the database as quick complete chip options. (Same UX pattern as is used for Symptoms see below)

![Screen Shot 2020-07-27 at 11 41 02 AM](https://user-images.githubusercontent.com/63205311/88533368-1dfd3e00-cffe-11ea-97f4-36db7ca906bf.png)

PR #766, opened 2020-08-11
Add an endpoint for listing the most frequently used occupations

allysonjp715:
Essentially copied #658
Needed for #634

### DIFF
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -193,8 +193,6 @@ paths:
           $ref: "#/components/responses/400"
         "403":
           $ref: "#/components/responses/403"
-        "422":
-          $ref: "#/components/responses/422"
         "500":
           $ref: "#/components/responses/500"
   /cases/placesOfTransmission:
@@ -219,8 +217,6 @@ paths:
           $ref: "#/components/responses/400"
         "403":
           $ref: "#/components/responses/403"
-        "422":
-          $ref: "#/components/responses/422"
         "500":
           $ref: "#/components/responses/500"
   /cases/occupations:
@@ -245,8 +241,6 @@ paths:
           $ref: "#/components/responses/400"
         "403":
           $ref: "#/components/responses/403"
-        "422":
-          $ref: "#/components/responses/422"
         "500":
           $ref: "#/components/responses/500"
   /cases/{id}:
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -223,6 +223,32 @@ paths:
           $ref: "#/components/responses/422"
         "500":
           $ref: "#/components/responses/500"
+  /cases/occupations:
+    get:
+      summary: Lists most frequently used occupations
+      operationId: listOccupations
+      parameters:
+        - name: limit
+          in: query
+          description: The number of items to return
+          required: false
+          schema:
+            type: integer
+            format: int32
+            minimum: 1
+            maximum: 100
+            default: 5
+      responses:
+        "200":
+          $ref: "#/components/responses/200OccupationArray"
+        "400":
+          $ref: "#/components/responses/400"
+        "403":
+          $ref: "#/components/responses/403"
+        "422":
+          $ref: "#/components/responses/422"
+        "500":
+          $ref: "#/components/responses/500"
   /cases/{id}:
     parameters:
       - name: id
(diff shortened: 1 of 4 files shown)


data-serving/data-service/api/openapi.yaml
```diff
@@ -223,6 +223,32 @@ paths:
...
+            maximum: 100
+            default: 5
+      responses:
+        "200":
+          $ref: "#/components/responses/200OccupationArray"
+        "400":
+          $ref: "#/components/responses/400"
+        "403":
+          $ref: "#/components/responses/403"
+        "422":
```
axmb, line 248: Does this return `422`? I'd think just `400` given that the limit param is fully specified here.

data-serving/data-service/src/controllers/case.ts
```diff
@@ -424,3 +424,35 @@ export const listPlacesOfTransmission = async (
...
+/**
+ * List most frequently used occupations.
+ *
+ * Handles HTTP GET /api/cases/occupations.
+ */
+export const listOccupations = async (
+    req: Request,
+    res: Response,
+): Promise<void> => {
+    const limit = Number(req.query.limit) || 5;
```
axmb, line 437: Feels a little un-DRY to `|| 5` this, given that we're applying the default in the API layer.

data-serving/data-service/src/controllers/case.ts
```diff
@@ -424,3 +424,35 @@ export const listPlacesOfTransmission = async (
...
+    if (limit < 1) {
```
axmb, line 438: Likewise, isn't this impossible with the OpenAPI spec?

data-serving/data-service/src/controllers/case.ts
```diff
@@ -424,3 +424,35 @@ export const listPlacesOfTransmission = async (
...
+        res.status(422).json('limit must be > 0');
+        return;
+    }
+    try {
+        const occupations = await Case.aggregate([
+            { $sortByCount: '$demographics.occupation' },
+            { $sort: { count: -1, _id: 1 } },
```
axmb, line 445: nit, but using the string/name value in lieu of `_id` will make this behavior consistent between different environments of our application that have identical data.

data-serving/data-service/src/controllers/case.ts
```diff
@@ -424,3 +424,35 @@ export const listPlacesOfTransmission = async (
...
+        ]).limit(limit);
+        res.json({
+            occupations: occupations.map(
+                (occupationObject) => occupationObject._id,
```
axmb, line 449: Ah, perhaps I'm misunderstanding above. Is `_id` the name field (for the "free indexing," or something)?

data-serving/data-service/src/index.ts
```diff
@@ -26,6 +19,13 @@ import mongoose from 'mongoose';
 import swaggerUi from 'swagger-ui-express';
 import validateEnv from './util/validate-env';
 
+// Set up appmetrics-dash before importing additional dependencies.
```
axmb, line 22: Sorry about this. FYI, if you save without formatting (ctrl + k -> s) it won't run the imports sort.

Unfortunately the order needs to be maintained for now, but I'll take care of this today.

data-serving/data-service/api/openapi.yaml
allysonjp715: Good point, removed elsewhere as well.

data-serving/data-service/src/controllers/case.ts
allysonjp715: Removed

data-serving/data-service/src/controllers/case.ts
allysonjp715: Removed

data-serving/data-service/src/controllers/case.ts
allysonjp715: The $sortByCount operation returns objects of a form something like `{_id: 'occupationName', count: 5}` so that can't be changed here.

data-serving/data-service/src/controllers/case.ts
allysonjp715: explained above

data-serving/data-service/src/index.ts
allysonjp715: aha, that's the command I was looking for, thanks! Reverted.

PR #768, opened 2020-08-11
Fix typo in event param and handle dates consistently

attwad:
The dates filtering code expects the dates to be formatted in a certain way, the zurich parser wasn't producing them accordingly (they were understandable by JS on the backend but not by the python code).

Also logs received event that should help debugging.

Confirmed that the date filtering now works as expected.

### DIFF
--- a/ingestion/functions/parsing/ch_zurich/zurich.py
+++ b/ingestion/functions/parsing/ch_zurich/zurich.py
@@ -31,6 +31,16 @@ _GENDER_INDEX = 3
 _CONFIRMED_INDEX = 4
 _SOURCE_INDEX = 6
 
+def convert_date(raw_date):
+    """
+    Convert raw date field into a value interpretable by the dataserver.
+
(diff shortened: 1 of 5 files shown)


ingestion/functions/parsing/ch_zurich/zurich.py
```diff
@@ -31,6 +31,16 @@
 _CONFIRMED_INDEX = 4
 _SOURCE_INDEX = 6
 
+def convert_date(raw_date):
+    """
+    Convert raw date field into a value interpretable by the dataserver.
+
+    The date is listed in dd-mm-YYYY format, but the date filtering API
```
axmb, line 38: Little confused by this -- it says Y-m-d below, which I thought was correctly handled.

Perhaps this changed since I last tested?

attwad: Fixed that comment sorry, what changed since last time is the date filtering happening in python.

PR #771, opened 2020-08-12
 Add happy-path upload API usage to parsing_lib.

axmb:
Doesn't include error handling, which'll be a separate PR (I'll base the branch off this PR). Tested locally with `india` parser.

Some autofmt changes included (I have `autopep8` as my configured Python formatter in vscode).

### DIFF
--- a/ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
+++ b/ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
@@ -22,13 +22,15 @@ s3_client = boto3.client("s3")
 def extract_event_fields(event):
     if any(
             field not in event
-            for field in [SOURCE_URL_FIELD, SOURCE_ID_FIELD, S3_BUCKET_FIELD, S3_KEY_FIELD]):
+            for field
+            in [SOURCE_URL_FIELD, SOURCE_ID_FIELD, S3_BUCKET_FIELD, S3_KEY_FIELD]):
         error_message = (
             f"Required fields {SOURCE_URL_FIELD}; {S3_BUCKET_FIELD}; "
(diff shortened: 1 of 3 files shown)


axmb:
> Looks fine but a less brittle alternative would have been to let the curator service do those /upload calls within the /cases/batchUpsert handler, the batchUpsert request could accept an upload ID and could even potentially use transactions in the handler to make sure that we're not creating inconsistent upload/cases relations. I'm sure you thought about this before and I've even reviewed the upload API so sorry for not asking that before but what's the rationale for having the /upload API logic in the clients?

Some of the operations could be done by the server, but I think we'll need the separation for full error handling. For instance, if there's a 500 error in the parser (crashing on data transformation, for instance), we'd ideally catch that and update the upload. Or, if/when we add errors for retrieval, we'll probably "begin" the upload at retrieval, and capture errors like "source not found."

Perhaps we start here, but look to fold in operations to the curator server as we can? The (better) transactionality is certainly a benefit.

attwad:
> > Looks fine but a less brittle alternative would have been to let the curator service do those /upload calls within the /cases/batchUpsert handler, the batchUpsert request could accept an upload ID and could even potentially use transactions in the handler to make sure that we're not creating inconsistent upload/cases relations. I'm sure you thought about this before and I've even reviewed the upload API so sorry for not asking that before but what's the rationale for having the /upload API logic in the clients?
> 
> Some of the operations could be done by the server, but I think we'll need the separation for full error handling. For instance, if there's a 500 error in the parser (crashing on data transformation, for instance), we'd ideally catch that and update the upload. Or, if/when we add errors for retrieval, we'll probably "begin" the upload at retrieval, and capture errors like "source not found."
> 
> Perhaps we start here, but look to fold in operations to the curator server as we can? The (better) transactionality is certainly a benefit.

LGTM

PR #774, opened 2020-08-12
nicer formatting of date filtering column

attwad:
I think this is the first time we have a column handle more than one field so that's nice.

ezgif com-video-to-gif (3)

(video even includes a live bug fix, what a gif(t))

### DIFF
--- a/ingestion/functions/README.md
+++ b/ingestion/functions/README.md
@@ -138,16 +138,13 @@ inferred from the `samconfig.toml` file. Follow the confirmation dialogues.
 
 Some sources do not provide a unique ID for each case allowing us to update existing cases in subsequent parsing runs.
 
-To accomodate for that, here is the procedure to write a parser that only imports data that is three days old (a reasonable threshold chosen arbitrarily):
-
-NOTE: This is a proposed method and being implemented in #743.
+To accomodate for that, here is the procedure to write a parser that only imports data that is three days old (a reasonable threshold chosen arbitrarily, feel free to tune it according to your source's freshness):
 
 1. write the parser, it must produces all cases for its input source, the `parsing/common/parsing_lib.py` library will ensure no duplicates are entered if you follow the next steps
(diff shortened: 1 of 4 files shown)


ingestion/functions/README.md
```diff
@@ -138,16 +138,13 @@ inferred from the `samconfig.toml` file. Follow the confirmation dialogues.
...
+To accomodate for that, here is the procedure to write a parser that only imports data that is three days old (a reasonable threshold chosen arbitrarily, feel free to tune it according to your source's freshness):
 
 1. write the parser, it must produces all cases for its input source, the `parsing/common/parsing_lib.py` library will ensure no duplicates are entered if you follow the next steps
-2. set the `dateFilter.numDaysBeforeToday` in your source to 3
-3. set the `dateFilter.op` in your source to `"LT"`
-4. run the parser once to import all the data up to 3 days before today
-5. set the `dateFilter.op` in your source to `"EQ"`
-6. set the daily cron expression in your source and have the parser run every day
+2. edit your source in the curator portal UI: set the date filter to only fetch data up to 3 days ago
+3. run the parser once to import all the data up to 3 days before today
```
axmb, line 145: Something I forgot to ask last time: by "run the parser once" -- do we mean set it daily and let it go once, or something else? If we mean have the contributor manually run it, we'll need to make sure they're a curator in prod and have a mechanism to use personal creds for local parser runs.

attwad: I think we'll have a button to trigger one off runs in the source table UI no? I know it's not there yet but I was planning on relying on that once present. Sounds reasonable?

axmb: Gotcha, yeah, that'd be a great feature. Maybe drop a comment or expand this (3) explaining the current state?

attwad: yup, filed an issue and added comments here.

PR #782, opened 2020-08-12
Combine case indexes into single index spec to make setup-db script collection-agnostic

khmoran:
In classic Kelly style, I'm really doing two things here:

- I started out adding 2 more indexes to the `cases` collection to speed up the charts queries, which unwind the `events` and `location.country` arrays
- But then I realized we may want to use the setup-db script to also clear out/set up other collections like `caserevisions`, so I also combined all the indexes into a single index spec so that setup-db can be collection-agnostic

Output of `dev/setup_db.sh`:
```
connecting to: mongodb://127.0.0.1:27017/?compressors=disabled&gssapiServiceName=mongodb
Implicit session: session { "id" : UUID("798beb39-3c3f-44c1-a8ea-be3679005364") }
MongoDB server version: 4.2.6
Read text index from ./../../data-service/schemas/cases.indexes.json
Dropped all documents (100 total) 🗑️
Applied schema to existing collection "cases" 📑
Dropping all indexes 👇
Creating indexes 👆
Done 👍
{ "collection" : "covid19.cases", "documents" : 10000, "indexes" : 5 }
Thank you for joining us 💾

> data-service@1.0.0 import-data /usr/local/google/home/khmoran/Dev/globaldothealth-line/data-serving/data-service
> mongoimport --db covid19 --collection cases --jsonArray --file ../samples/cases.json

2020-08-12T10:19:22.114-0400    connected to: mongodb://localhost/
2020-08-12T10:19:22.128-0400    100 document(s) imported successfully. 0 document(s) failed to import.
```

Screenshot of indexes afterwards:
Screenshot from 2020-08-12 10-23-03

APIs:
- https://docs.mongodb.com/manual/reference/method/db.collection.dropIndexes/
- https://docs.mongodb.com/manual/reference/command/createIndexes/ (chose this instead of the `collection` one so that we can express all indexes in a single JSON)


data-serving/scripts/setup-db/src/index.ts
```diff
@@ -16,19 +15,15 @@ const setupDatabase = async ({
...
         const schema = JSON.parse(await cat(schemaPath));
         print(`Read schema from ${schemaPath}`);
 
-        const textIndex = JSON.parse(await cat(textIndexPath));
-        print(`Read text index from ${textIndexPath}`);
-
-        const caseRefIndex = JSON.parse(await cat(caseRefIndexPath));
-        print(`Read caseRef index from ${caseRefIndexPath}`);
+        const indexes = JSON.parse(await cat(indexesPath));
+        print(`Read text index from ${indexesPath}`);
```
axmb, line 26: nit: s/text index/indexes

data-serving/data-service/schemas/cases.indexes.json
```diff
@@ -0,0 +1,45 @@
...
+        }
+    },
+    {
+        "name": "casesCaseRefIdx",
+        "key": {
+            "caseReference.sourceId": -1,
+            "caseReference.sourceEntryId": -1
+        },
+        "unique": true,
+        "partialFilterExpression": {
```
axmb, line 27: Great side-benefit is that these opts are now in the same place as the key definition.

data-serving/data-service/schemas/cases.indexes.json
```diff
@@ -0,0 +1,45 @@
...
+            "caseReference.sourceEntryId": {
+                "$exists": true
+            }
+        }
+    },
+    {
+        "name": "eventsArrayIdx",
+        "key": {
+            "events": -1
```
axmb, line 36: Do we want the index on the whole event document, or a field (like name)? I'm not sure what the charts query is, apologies.

khmoran: oops you're totally right, thanks!

data-serving/data-service/schemas/cases.indexes.json
khmoran: :D

data-serving/scripts/setup-db/src/index.ts
khmoran: good catch, thanks!

PR #789, opened 2020-08-13
Move table actions to spec-ed locations

allysonjp715:
Moves edit and delete within a three-dot menu. The details view can be accessed by clicking anywhere on the row. 

I don't think the three-dot menu can be moved to the left of the checkboxes until https://github.com/mbrn/material-table/issues/2317 is fixed.

ezgif com-video-to-gif (2)

### DIFF
--- a/verification/curator-service/ui/src/components/LinelistTable.tsx
+++ b/verification/curator-service/ui/src/components/LinelistTable.tsx
@@ -1,5 +1,8 @@
 import {
     Button,
+    IconButton,
+    Menu,
+    MenuItem,
     Theme,
     Tooltip,
     makeStyles,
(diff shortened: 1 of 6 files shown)


verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -219,6 +222,81 @@ function SearchBar(props: {
...
+            >
+                <MoreVertIcon />
+            </IconButton>
+            <Menu
+                anchorEl={anchorEl}
+                keepMounted
+                open={Boolean(anchorEl)}
+                onClose={handleClose}
+            >
+                <MenuItem onClick={handleEdit}>
```
attwad, line 287: Instead of having to deal with history and pushing stuff to it, we should simply use <Link> elements, I think I've done that in the main menu or somewhere else... I've read it's better practice :) (it is definitely simpler)

allysonjp715: In this case I don't think it's worth it to use `<Link>`. It would only be useful on the Edit MenuItem since the delete MenuItem isn't doing a page change. It also can't be used for navigating to Details since we just get an onRowClicked callback to trigger that. For the edit menu item we still need to handle onClick on the MenuItem because we need to stop the event propagation from triggering the row click. Since we already need that it's just two more lines to `useHistory` and call `history.push(...)`

PR #798, opened 2020-08-14
Add and set verificationStatus in case schema/API

axmb:


axmb:
> Looks fine but do we expect more than a boolean status for the verified status? if not why not making this a boolean? Perhaps expliciting why it is a string/enum in the schema would help.

Yeah, I wasn't sure on this. We'd discussed having a "rejected" state for data, but most recently put forth that this data would just be removed from the DB. Without that, I think there are just two states here. I can switch this back over to a boolean if we find that's all we need for verification in beta.

attwad:
sgtm, thanks for the clarification

PR #802, opened 2020-08-17
Use cypress action to run on firefox and chrome in parallel

### LINKED ISSUE #776: Enable edge and firefox cross browser testing on our cypress tests
https://docs.cypress.io/guides/guides/cross-browser-testing.html#Continuous-Integration-Strategies

attwad:
Fixes #776 

Haven't looked into edge yet, it's not bundled with ubuntu by default (figures...) so we'd need to spin a windows VM, will look into that next.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -26,11 +26,11 @@ jobs:
       - uses: actions/checkout@v2
       - name: Build stack
         run: docker-compose -f dev/docker-compose.yml -f dev/docker-compose.ci.yml up --build -d -V
-      - name: Cypress integration tests
-        run: |
-          npm ci
-          npm run cypress-run
-        working-directory: verification/curator-service/ui
+      - uses: cypress-io/github-action@v2
+        with:
+          working-directory: verification/curator-service/ui
+          wait-on: 'http://localhost:3002'
+          wait-on-timeout: 300
         env:
           CI: true
       # Screenshots are only available on failures.
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -22,6 +22,11 @@ jobs:
   integration-tests:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        browser: ["chrome", "firefox"]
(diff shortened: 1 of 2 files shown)


verification/curator-service/ui/package.json
```diff
@@ -64,7 +64,9 @@
     "test-silent": "react-scripts test --watchAll=false",
     "eject": "react-scripts eject",
     "cypress-open": "cypress open",
-    "cypress-run": "cypress run"
+    "cypress-run": "cypress run",
```
allysonjp715, line 67: Does plain `cypress run` use chrome or Electron? If it uses Electron can you add a command for chrome?

.github/workflows/integration-tests.yml
```diff
@@ -22,15 +22,21 @@ jobs:
   integration-tests:
     runs-on: ubuntu-latest
 
+    strategy:
+      matrix:
+        browser: ["chrome", "firefox"]
```
allysonjp715, line 27: Will this double integration running time? I've had integration tests take ~15 minutes to run before, so doubling that would be impactful.  

I'm open to setting this here for now and seeing how it goes. Another option would be to run firefox on merging to main rather than on every pull request, since I'm hoping it's unlikely chrome will succeed and firefox will fail.

allysonjp715: Oh on this PR it looks like they may be running in parallel. That would be great.

verification/curator-service/ui/package.json
attwad: whatever your default setting is locally, if you have chrome it will use chrome otherwise it will use electron I believe which is bundled with it.

.github/workflows/integration-tests.yml
attwad: no the matrix setup makes the tests run in parallel so I don't expect an increase.
On a test run it took longer on firefox but I am not sure why, https://github.com/globaldothealth/list/actions/runs/212060717

let me see what the results are for the current run...

attwad: hmm they actually took exactly the same time on that last run so I think we're good
https://github.com/globaldothealth/list/actions/runs/212095963

PR #808, opened 2020-08-18
Display verification status in the linelist table

axmb:
verification

Header button doesn't toggle status yet -- that's probably better for a separate PR.


verification/curator-service/ui/src/components/VerificationStatusHeader.tsx
```diff
@@ -0,0 +1,14 @@
+import { IconButton, Tooltip } from '@material-ui/core';
+
+import React from 'react';
+import { ReactComponent as VerificationHeaderSvg } from './assets/verification_header.svg';
+
+export default function VerificationStatusHeader(): JSX.Element {
+    return (
+        <Tooltip title="Coming soon: toggle verification status">
```
allysonjp715, line 8: What do you mean by toggle? Usually clicking on a header sorts the column (which we currently have disabled on the table)

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -414,6 +421,25 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
                                   },
                               ]
                             : []),
+                        {
```
allysonjp715, line 424: If you want the column to be narrower you could add padding: 0 to cell style like the field above this one.

verification/curator-service/ui/src/components/VerificationStatusHeader.tsx
axmb: That's how I read [the description in the mocks](https://www.figma.com/file/9HJlRkTBCMEQdkJ6WQJYkv/061120_Global.Health-Line?node-id=981%3A901). Looking back, I could have taken that to mean the header is used to filter, as well. I'll change this text in the meantime.

verification/curator-service/ui/src/components/LinelistTable.tsx
axmb: Will do, thanks!

PR #815, opened 2020-08-18
Add source name and URL to response of listUploads

allysonjp715:


verification/curator-service/api/src/controllers/uploads.ts
```diff
@@ -62,7 +62,14 @@ export default class UploadsController {
         try {
             const uploads = await Source.aggregate([
                 { $unwind: '$uploads' },
-                { $replaceRoot: { newRoot: '$uploads' } },
+                {
+                    $project: {
+                        _id: 0,
```
axmb, line 67: nit, but slightly clearer for this to be `false` IMO

allysonjp715: I agree, done.

PR #818, opened 2020-08-18
Parse ListUploads response in alerts component per new API structure.

axmb:


allysonjp715 (approved):
Ah thank you! I had this in my next PR but should've split it out to get the cypress tests passing.

PR #819, opened 2020-08-18
Add a toggle to verify cases from the linelist

axmb:
There are outstanding questions on the spec for this. Adding a couple simple actions to manipulate verification status for now.

verify


allysonjp715 (approved):
Nice!

PR #821, opened 2020-08-19
update deps to fix 4k vulns or so...

attwad:


allysonjp715 (approved):
4k vulns just from react-scripts? jeez

PR #823, opened 2020-08-19
Use the sam CLI, now installed by default in action images, to sam deploy pushes to main

axmb:
Apologies for any commit churn. Not feasible to test this locally with `act`.

(Checks aren't run on this final commit, since it isn't against `main`, but you can view previous successful runs in the Checks tab.)


attwad (approved):
re commit churn: been there, done that, nw

PR #730, opened 2020-08-05
Japan

AnyaLindstromBattle:
Merging the files I have changed for the Japan parser to the main branch. Please note that because of server access issues I was unable to check that the writes to the API were as they should be. Instead I had the output of the parser print itself to the terminal and checked it that way and it looked correct. It might be good if someone with working access to the database/API could check the parser output in there, just to be on the safe side.

PR #827, opened 2020-08-20
Japan parser

### LINKED ISSUE #481: Import Japan case data
https://github.com/reustle/covid19japan-data/tree/master/docs/patient_data

attwad:
Taken from #730 as Anya couldn't get it to work locally, I patched and tested myself instead.

Also increases timeouts of functions because there is a lot of data in Japan and 60 secs wasn't enough.

Fixes #481


ingestion/functions/parsing/japan/japan.py
```diff
@@ -0,0 +1,87 @@
...
+                    "sourceEntryId": entry["patientId"],
+                    "sourceUrl": source_url
+                },
+                "revisionMetadata": {
+                    "revisionNumber": 0,
+                    "creationMetadata": {
+                        "curator": "auto",
+                        "date": date.today().strftime("%m/%d/%Y")
+                    }
+                },
```
axmb, line 66: A separate PR, but we should move this into the library. I can't recall -- is this part used, still, or does the service account identity get inferred/used based on it supplying the bearer token?

ingestion/functions/template.yaml
```diff
@@ -25,7 +25,7 @@ Resources:
         - AWSLambdaFullAccess
         - S3WritePolicy:
             BucketName: !Ref RawSourcesBucket
-      Timeout: 60 # Seconds
+      Timeout: 600 # Seconds
```
axmb, line 28: Probably better just to remove this. I had it at 60 originally for development purposes, but in prod we probably want it to take as much time as it needs (up to the 900 second limit).

ingestion/functions/parsing/japan/japan.py
attwad: oh I missed that, it's not needed at all, removed.

ingestion/functions/template.yaml
attwad: ok fair enough, let's remove it.

PR #834, opened 2020-08-21
Finish source creation modal and use it in lieu of source table add button

axmb:


verification/curator-service/ui/src/components/AutomatedSourceForm.tsx
```diff
@@ -129,6 +141,7 @@ export default function AutomatedSourceForm(props: Props): JSX.Element {
                                 </div>
                                 <div className={classes.formSection}>
                                     <FastField
+                                        helperText="Required"
```
allysonjp715, line 144: UX nit: (can be addressed later) if all fields are required I think it's a better practice to say that in descriptive text at the top or similar, rather than marking each individual field as required.

verification/curator-service/ui/src/components/SourceTable.tsx
```diff
@@ -538,8 +517,8 @@ class SourceTable extends React.Component<Props, SourceTableState> {
                             this.tableRef.current.onQueryChange();
                         }}
                         editable={{
-                            onRowAdd: (rowData: TableRow): Promise<unknown> =>
-                                this.addSource(rowData),
+                            // Don't allow adding, but do allow editing.
+                            onRowAdd: undefined,
```
allysonjp715, line 521: nit: we've just deleted this field on the other tables.

PR #843, opened 2020-08-21
Move remaining common code into common_lib.py

axmb:
Minimal real delta here. There's a net decrease in LoC, due to axing some duplication. Thereby advancing our quest for 0 LoC.

Tested local invoke for retrieval/parsing, which both appear to work just fine.


attwad (approved):
Thy quest is a noble one, I shall approve it.

PR #845, opened 2020-08-24
remove unused url filtering in source controller API

attwad:
This seems unused afaict, we used to have columns filtering in the table but disabled it, if we ever need this we should switch to the same UI/full text search/keyword search pattern as for the linelist imo.

### DIFF
--- a/verification/curator-service/api/openapi/openapi.yaml
+++ b/verification/curator-service/api/openapi/openapi.yaml
@@ -56,14 +56,6 @@ paths:
             minimum: 1
             maximum: 100
             default: 10
-        - name: url
-          in: query
-          description: Origin URL substring on which to filter the result set
-          required: false
-          allowEmptyValue: true
-          allowReserved: true
-          schema:
-            type: string
         - name: access_token
           in: query
           required: false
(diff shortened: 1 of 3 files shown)


attwad:
@axmb if you have any clue as to why the bulk upload tests are failing I'm all ears...  this seems unused but I might have missed something...

axmb:
> @axmb if you have any clue as to why the bulk upload tests are failing I'm all ears... this seems unused but I might have missed something...

It's used for autocompleting the source URL in the bulk and manual forms:

https://github.com/globaldothealth/list/blob/main/verification/curator-service/ui/src/components/common-form-fields/Source.tsx#L126

Perhaps the manual form Cypress test doesn't have a case where an existing source is selected, or the source selection logic is different?

attwad:
> > @axmb if you have any clue as to why the bulk upload tests are failing I'm all ears... this seems unused but I might have missed something...
> 
> It's used for autocompleting the source URL in the bulk and manual forms:
> 
> https://github.com/globaldothealth/list/blob/main/verification/curator-service/ui/src/components/common-form-fields/Source.tsx#L126
> 
> Perhaps the manual form Cypress test doesn't have a case where an existing source is selected, or the source selection logic is different?

ahhhh makes sense, my search regexp didn't catch that... okay I'll drop this PR then ") thanks

PR #847, opened 2020-08-24
Make it possible to update source with a parser arn only without requiring a schedule

### LINKED ISSUE #838: Make it possible to specify a parser ARN without a schedule
In the sources tab, that will make it easier to test parsers.

attwad:
Fixes #838

### DIFF
--- a/verification/curator-service/api/src/controllers/sources.ts
+++ b/verification/curator-service/api/src/controllers/sources.ts
@@ -113,7 +113,10 @@ export default class SourcesController {
     private async updateAutomationScheduleAwsResources(
         source: SourceDocument,
     ): Promise<void> {
-        if (source.isModified('automation.schedule.awsScheduleExpression')) {
+        // Careful here, source.isModified('automation.schedule.awsScheduleExpression')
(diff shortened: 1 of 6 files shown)


verification/curator-service/api/src/controllers/sources.ts
```diff
@@ -113,7 +113,10 @@ export default class SourcesController {
     private async updateAutomationScheduleAwsResources(
         source: SourceDocument,
     ): Promise<void> {
-        if (source.isModified('automation.schedule.awsScheduleExpression')) {
+        // Careful here, source.isModified('automation.schedule.awsScheduleExpression')
```
axmb, line 116: That's odd. I wonder if `isDirectModified` is better, here? All good if this new logic works as well.

https://mongoosejs.com/docs/api/document.html#document_Document-isDirectModified

attwad: I tried isDirectModified but it didn't trigger as expected :( pretty obscure tbh... the added tests and manual checks that I've done make me quite confident this is fine.

PR #849, opened 2020-08-24
Add beginnings of uploads table

allysonjp715:
Screen Shot 2020-08-24 at 1 14 53 PM

### DIFF
--- a/verification/curator-service/ui/src/components/UploadsTable.tsx
+++ b/verification/curator-service/ui/src/components/UploadsTable.tsx
@@ -1,0 +1,190 @@
+import MaterialTable, { QueryResult } from 'material-table';
+import React, { RefObject } from 'react';
+import {
+    Theme,
+    WithStyles,
+    createStyles,
+    withStyles,
+} from '@material-ui/core/styles';
+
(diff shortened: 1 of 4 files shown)


verification/curator-service/ui/src/components/UploadsTable.tsx
```diff
@@ -0,0 +1,190 @@
...
+                                listUrl += '&page=' + (query.page + 1);
+                                this.setState({ error: '' });
+                                const response = axios.get<ListUploadsResponse>(
+                                    listUrl,
+                                );
+                                response
+                                    .then((result) => {
+                                        const flattenedSources: TableRow[] = [];
+                                        const uploadData = result.data.uploads;
+                                        for (const u of uploadData) {
```
attwad, line 136: nit, you could do something like
```
const flattenedSources = result.data.uploads.map((u) => {
  return {...}
})
```

attwad (approved):

verification/curator-service/ui/src/components/UploadsTable.tsx
allysonjp715: Done

PR #851, opened 2020-08-24
make the search bar an autocomplete component to suggest keywords

attwad:


allysonjp715 (approved):
good idea!

attwad:
Hmm, seems like there is a bit more work to do to pass the state from URL, will look at that tomorrow.

PR #858, opened 2020-08-25
Use batch delete endpoint

allysonjp715:


verification/curator-service/api/src/controllers/cases.ts
```diff
@@ -103,7 +103,7 @@ export default class CasesController {
         try {
             const response = await axios.delete(
                 this.dataServerURL + '/api' + req.url,
-                req.body,
+                { data: req.body },
```
attwad, line 106: why do we need the data: indirection here while all the other methods proxying just pass the req.body as is ?

attwad, line 106: also if really needed, should data be mentioned in the openapi spec ?

attwad (approved):

verification/curator-service/api/src/controllers/cases.ts
allysonjp715: Axios quirk 🤷‍♀️   took me awhile to figure it out https://github.com/axios/axios/issues/736

allysonjp715: No, in the final request it is sent as the actual body. Just a weird thing with axios.

Issue #780, opened 2020-08-12
Create data dictionary and provide it on github alongside data exports


khmoran:
We should provide a data dictionary to accompany the CSV and JSON exports of the case data so users know (1) what it contains and (2) how to interpret it.

sratcliffe118:
@outbreakprepared 
I think it would be good practice to have someone who is going to be on the project long time compile this. Happy to consult and partner but needs to be well understood post fellowship.

Moritz - please let me know who can lead this and I will work with them.

dakrownie:
@outbreakprepared 
Bumping this up in your inbox.
Who do you suggest per Stephen's comment above?

PR #859, opened 2020-08-25
Switch data export script to python and provide data dictionary alongside input fields

attwad:
This is to make the data dictionary part of the export process to make sure we are always in sync (not provide fields we do not document).

The fields are now described in a yaml file, so I switched to python that just invokes the mongoexport CLI using subprocess, no big change overall.

For #780


allysonjp715 (approved):

.github/workflows/case-data-export-prod.yml
```diff
@@ -28,11 +28,17 @@ jobs:
           sudo dpkg --purge mongodb-org-database-tools-extra
           sudo apt-get update
           sudo apt-get install -f -y --allow-downgrades mongodb-org=4.2.6 mongodb-org-server=4.2.6 mongodb-org-shell=4.2.6 mongodb-org-mongos=4.2.6 mongodb-org-tools=4.2.6 libcurl3
-
+      - uses: actions/setup-python@v1
+        with:
+          python-version: 3.8
+      - name: Install python dependencies
+        run: |
+          python -m pip install --upgrade pip
```
axmb, line 36: Any reason we're using `python` here and `python3` below? Mostly a nit -- they're probably symlinked to the same thing in `setup-python`.

.github/workflows/case-data-export-prod.yml
```diff
@@ -28,11 +28,17 @@ jobs:
           sudo dpkg --purge mongodb-org-database-tools-extra
           sudo apt-get update
           sudo apt-get install -f -y --allow-downgrades mongodb-org=4.2.6 mongodb-org-server=4.2.6 mongodb-org-shell=4.2.6 mongodb-org-mongos=4.2.6 mongodb-org-tools=4.2.6 libcurl3
-
+      - uses: actions/setup-python@v1
```
axmb, line 31: I think v2 is the latest GA hotness.

.github/workflows/case-data-export-prod.yml
attwad: ah I didnt notice, switched to pythn3 for consistency.

.github/workflows/case-data-export-prod.yml
attwad: cool, done

PR #861, opened 2020-08-25
fix search bar ui

attwad:
Before:
image

After: 
image

I removed the search icon because when used as a start adornment the UI became what it was apparently that's a bug :(

Also moved the tooltip to a separate portion of the UI after the search bar so that we can have the X to clear input in it I think it's cleaner:
image


allysonjp715 (approved):
Nice, thanks! Small nit: could you add some left/right margin around the tooltip?

attwad:
image

PR #863, opened 2020-08-25
Add a button for selecting all rows across pages

allysonjp715:
Need to update actions and exact styling in future PRs

Screen Shot 2020-08-25 at 3 39 06 PM

### DIFF
--- a/verification/curator-service/ui/src/components/LinelistTable.tsx
+++ b/verification/curator-service/ui/src/components/LinelistTable.tsx
@@ -702,7 +702,9 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
                                       onClick: (_: any, rows: any): void => {
                                           if (
                                               this.state.totalNumRows ===
-                                              this.state.numSelectedRows
+                                                  this.state.numSelectedRows &&
+                                              this.state.numSelectedRows >
+                                                  this.state.pageSize
                                           ) {
                                               // TODO: Implement action for all rows
                                               alert(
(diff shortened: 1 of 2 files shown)


verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -300,6 +302,8 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
             error: '',
             pageSize: 50,
```
attwad, line 303: this default should be a constant reused in the test to avoid magic numbers

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -632,13 +658,60 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
...
+                                  },
                                   {
                                       icon: (): JSX.Element => (
                                           <VerifiedIcon data-testid="verify-action" />
                                       ),
                                       tooltip: 'Verify selected rows',
                                       // eslint-disable-next-line @typescript-eslint/no-explicit-any
                                       onClick: (_: any, rows: any): void => {
+                                          if (
+                                              this.state.totalNumRows ===
```
attwad, line 704: this if is repeated 3 times, factor it in a class method please

attwad (approved):

verification/curator-service/ui/src/components/LinelistTable.tsx

verification/curator-service/ui/src/components/LinelistTable.tsx
allysonjp715: I instead added clicks to change the number of rows per page in the test. That has a couple benefits: no longer relies on the default number here, adds test coverage for changing the number of rows per page, and decreases the number of cases needed to add in that test.

attwad: good call 👍

PR #865, opened 2020-08-26
Add confirmation dialog before deleting cases

allysonjp715:
Screen Shot 2020-08-26 at 11 23 12 AM

### DIFF
--- a/verification/curator-service/ui/src/components/LinelistTable.tsx
+++ b/verification/curator-service/ui/src/components/LinelistTable.tsx
@@ -1,5 +1,10 @@
 import {
     Button,
+    Dialog,
+    DialogActions,
+    DialogContent,
+    DialogContentText,
+    DialogTitle,
     IconButton,
     Menu,
     MenuItem,
@@ -43,8 +48,13 @@ interface LinelistTableState {
     error: string;
     pageSize: number;
     search: string;
+    // The rows which are selected on the current page.
+    selectedRowsCurrentPage: TableRow[];
+    // The total number of rows selected. This can be larger than
+    // selectedRowsCurrentPage.length if rows across all pages are selected.
     numSelectedRows: number;
     totalNumRows: number;
+    deleteDialogOpen: boolean;
 }
 
 // Material table doesn't handle structured fields well, we flatten all fields in this row.
@@ -232,6 +242,9 @@ function RowMenu(props: {
     refreshData: () => void;
 }): JSX.Element {
     const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null);
+    const [deleteDialogOpen, setDeleteDialogOpen] = React.useState<boolean>(
+        false,
+    );
     const classes = rowMenuStyles();
 
     const handleClick = (event: React.MouseEvent<HTMLButtonElement>): void => {
@@ -244,6 +257,11 @@ function RowMenu(props: {
         setAnchorEl(null);
     };
 
+    const openDeleteDialog = async (event?: any): Promise<void> => {
+        event?.stopPropagation();
+        setDeleteDialogOpen(true);
+    };
+
     const handleDelete = async (event?: any): Promise<void> => {
         event?.stopPropagation();
         try {
@@ -254,6 +272,7 @@ function RowMenu(props: {
         } catch (e) {
             props.setError(e.toString());
         } finally {
+            setDeleteDialogOpen(false);
             handleClose();
         }
     };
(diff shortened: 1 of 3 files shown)


verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -282,11 +301,40 @@ function RowMenu(props: {
...
-                <MenuItem onClick={handleDelete}>
+                <MenuItem onClick={openDeleteDialog}>
                     <DeleteIcon />
                     <span className={classes.menuItemTitle}>Delete</span>
                 </MenuItem>
             </Menu>
+            <Dialog
+                open={deleteDialogOpen}
+                onClose={(): void => setDeleteDialogOpen(false)}
+                onClick={(e): void => e.stopPropagation()}
```
attwad, line 312: is that for when you click on the dialog? what purpose does it serve?

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -282,11 +301,40 @@ function RowMenu(props: {
...
+                        Case {props.rowId} will be permanently deleted and can
+                        not be recovered.
+                    </DialogContentText>
+                </DialogContent>
+                <DialogActions>
+                    <Button
+                        onClick={(): void => {
+                            setDeleteDialogOpen(false);
+                        }}
+                        color="primary"
```
attwad, line 328: your screenshots didn't seem to have any color, is this "primary" working?

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -282,11 +301,40 @@ function RowMenu(props: {
...
+                        autoFocus
+                    >
+                        Cancel
+                    </Button>
+                    <Button onClick={handleDelete} color="primary">
```
attwad, line 333: ditto, also shouldn't be the same color as the other button no? perhaps primary and secondary are better.

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -323,16 +374,21 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
...
-                this.setState({ error: e.toString() });
-                reject(e);
-            });
+    async deleteCases(): Promise<void> {
+        if (this.hasSelectedRowsAcrossPages()) {
+            // TODO: Implement action for all rows
+            alert('Action not yet implemented when all rows selected');
+            return;
+        }
+        await axios.delete('/api/cases', {
```
attwad, line 383: I think you have to try/catch otherwise if an error occurs here the UI will crash

attwad, line 383: do we want a progress bar somewhere? this could take some time.

verification/curator-service/ui/src/components/LinelistTable.tsx
allysonjp715: Yes, perhaps hard to tell in the screenshot but the text on the buttons is green

verification/curator-service/ui/src/components/LinelistTable.tsx
allysonjp715: It stops the click being propagated to behind the dialog. Without this when you click on the dialog the onRowClick gets triggered and shows the ViewCase modal.

verification/curator-service/ui/src/components/LinelistTable.tsx
allysonjp715: Usually the YES button would be filled to draw attention to it. But when it’s a confirmation dialog you don’t want to encourage them to click YES. https://material.io/components/dialogs#confirmation-dialog

verification/curator-service/ui/src/components/LinelistTable.tsx
allysonjp715: Ah yes, thanks. Done.

allysonjp715: There is a slight lag, but the good news is it doesn't seem to be any longer if selecting more rows (I tried with 100 and it wasn't noticeably longer than 1 case). I tried quickly to put a loading spinner instead of the yes button, but it adds even more lag since it changes the components state causing the entire thing to re-render. We could potentially pull this dialog out into it's own component (and potentially share with the RowMenu dialog?) but that would be quite a big refactor. I don't think the lag is great enough to necessitate that yet, but we can see how it is on dev and make that change if we'd like to.

attwad: ok fine for now, we will definitely need this for bigger queries I don't expect a slowdown from 1 to 100 queries but when we start hitting the thousands/hundreds of thousands yes it will definitely be needed.

PR #866, opened 2020-08-26
Allow deleting cases through search query

allysonjp715:


data-serving/data-service/src/controllers/case.ts
```diff
@@ -357,20 +357,40 @@ export const upsert = async (req: Request, res: Response): Promise<void> => {
...
+                    res.status(500).json(err.message);
+                    return;
+                }
+                res.status(204).end();
+            },
+        );
+        return;
+    }
+    if (req.body.query.trim() === '') {
+        res.status(422).json('query must not be empty');
```
attwad, line 378: not sure if we can do this as part of the openapi spec?

data-serving/data-service/src/controllers/case.ts
```diff
@@ -357,20 +357,40 @@ export const upsert = async (req: Request, res: Response): Promise<void> => {
...
+        return;
+    }
+    const parsedSearch = parseSearchQuery(req.body.query);
+    const queryOpts = parsedSearch.fullTextSearch
```
attwad, line 382: huh, that won't get you keyword searches or mixed FTS/keyword based search queries.
i.e suppose I want to delete all the stuff I imported for india because it was wrong the query will be "india curator:timothe@google.com" but it will actually delete everything for india :-(

Take a look at L51 in this file on how to handle keywords as well.

data-serving/data-service/src/controllers/case.ts
allysonjp715: Ah, you can specify a regex, done.

data-serving/data-service/src/controllers/case.ts
allysonjp715: Ah, thanks. Done.

attwad: we should probably move that to a shared method, can you leave a todo or do it please? thanks

attwad (approved):

data-serving/data-service/src/controllers/case.ts
allysonjp715: Moved to shared method.

PR #868, opened 2020-08-26
Determine G.h environment in parsing functions via an env input parameter

axmb:
This covers:

- Creating the env var in our AWS deployment(s).
- Supplying it as JSON in our scheduled CloudWatch events.
- Using it to determine G.h API URL in retrieval.

Next step is using it in the parsing_lib, and removing references to the legacy os.environ param.

Handful of files impacted, but the commits should be nice logical chunks if you'd like to review that way.

To expand on the overall approach:

I'd considered doing full separation of serverless resources between environments. The "right" way to do that with AWS is to create separate CloudFormation stacks for each environment (using aliases to do it isn't recommended). It'd add a fair bit of cognitive overhead and complexity -- just didn't seem worth the payoff, given that there's no real resource separation advantage, since an "environment" for a Lambda is just a version of code to run.

So instead, G.h servers can include the environment in which a rule was configured in the input to ingestion, and our functions can determine which server to hit using that information. I'll talk this over with Chris/Spencer during handoffs; it won't be hard to simply templatize our SAM set up with some specific dev/prod params, and instead target the relevant environment's retrieval function from our servers, but to me it doesn't seem quite worth it for now.

### DIFF
--- a/ingestion/functions/retrieval/retrieval.py
+++ b/ingestion/functions/retrieval/retrieval.py
@@ -8,6 +8,7 @@ import requests
 
 from datetime import datetime, timezone
 
+ENV_FIELD = "env"
 OUTPUT_BUCKET = "epid-sources-raw"
 SOURCE_ID_FIELD = "sourceId"
 TIME_FILEPART_FORMAT = "/%Y/%m/%d/%H%M/"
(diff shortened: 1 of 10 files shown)


ingestion/functions/retrieval/retrieval.py
```diff
@@ -112,9 +123,10 @@ def upload_to_s3(file_name, s3_object_key, source_id, upload_id, api_headers):
 
 
 def invoke_parser(
-    parser_arn, source_id, upload_id, api_headers, s3_object_key,
+    env, parser_arn, source_id, upload_id, api_headers, s3_object_key,
         source_url, date_filter):
     payload = {
+        "env": env,
```
attwad, line 129: perhaps passing the API URL to the parser would make parsing lib simpler, although it's already in common_lib so no big deal... up to you.

ingestion/functions/retrieval/retrieval_test.py
```diff
@@ -64,21 +84,20 @@ def test_lambda_handler_e2e(valid_event, requests_mock, s3):
     s3.create_bucket(Bucket=retrieval.OUTPUT_BUCKET)
 
     # Set up mock request values used in multiple requests.
-    source_api_url = "http://foo.bar"
-    os.environ["SOURCE_API_URL"] = source_api_url
+    os.environ["SOURCE_API_URL"] = _SOURCE_API_URL  # Still needed for now.
```
attwad, line 87: for now? what do we need TODO to fix it? /wink

verification/curator-service/api/src/util/validate-env.ts
```diff
@@ -61,36 +75,28 @@ export default function validateEnv(): Readonly<{
...
+            default: true,
+        }),
+        MAPBOX_TOKEN: str({
+            desc: 'Mapbox token to use for geocoding',
+            devDefault: '',
+        }),
+        PORT: port({ default: 3001 }),
+        SERVICE_ENV: str({
+            choices: ['local', 'dev', 'prod'],
+            desc: 'Environment in which the service is running',
```
attwad, line 90: don't you need to edit the environment vars in kubernetes as well? plan to do it in a later PR?

attwad (approved):

verification/curator-service/api/src/util/validate-env.ts
axmb: Are the changes in `curator.yaml` sufficient?

ingestion/functions/retrieval/retrieval_test.py
axmb: Heh, thanks, done.

ingestion/functions/retrieval/retrieval.py
axmb: Yeah, I'd considered this. Should have added a comment in the PR description. This approach adds one "layer" as mentioned, but the plus side is that there's a single place this information is encoded. So if the API URL changes, or something, we can just edit and deploy `common_lib`. Not a huge deal, but nicer than having the information embedded in every event, and needing to somehow update or remake them.

Probably not a huge deal either way, but those were my thoughts.

attwad: fwiw that information would still be encoded in one place (retrieval.py) if we moved to a url param sent in the event to the parsing layers. But yeah no big deal either way. I just feel like layers in lambda add a layer of complexity (who would have thought?...) especially around weird import tricks that we have to do... anyway, fine :)

verification/curator-service/api/src/util/validate-env.ts
attwad: ohh, yeah sorry I missed that, nice use of metadata.labels.environment here

PR #872, opened 2020-08-26
Germany

calremmel:
Here's a first take on a parser for Germany. This version ignores redaction lines for the previous day (rows with -1 cases, deaths, or recoveries), as there aren't individual case IDs in the data to match them to. I believe these are resolved with the following publication, but would love to confirm this somehow!


ingestion/functions/parsing/germany/germany.py
```diff
@@ -0,0 +1,111 @@
...
+        cases = []
+        for row in reader:
+            num_confirmed_cases = int(row[_CASECOUNT])
+            if num_confirmed_cases < 1:
+                continue
+            try:
+                case = {
+                    "caseReference": {"sourceId": source_id, "sourceUrl": source_url},
+                    "location": {
+                        "country": "Germany",
```
attwad, line 79: I will clarify this in the docs for writing parsers today but parsers should only output a query, they do not need to pass anything else, in your case you know that's always admin2 so you can also pass `limitToResolution: 'Admin2'` in the location dictionary which will restrict geocodes to admin2.

ingestion/functions/parsing/germany/germany.py
```diff
@@ -0,0 +1,111 @@
...
+                        },
+                    ],
+                    "demographics": convert_demographics(
+                        row[_GENDER_INDEX], row[_AGE_INDEX]
+                    ),
+                }
+                # Death and recovery counts are always equal to case counts for the row.
+                if int(row[_DEATHCOUNT]) > 0:
+                    case["events"].append({"name": "outcome", "value": "Death"})
+                if int(row[_RECOVERYCOUNT]) > 0:
```
attwad, line 102: elif ? if I understood correctly...

ingestion/functions/parsing/germany/germany.py
```diff
@@ -0,0 +1,111 @@
...
+
+def convert_demographics(gender: str, age: str):
+    demo = {}
+    if gender:
+        demo["gender"] = convert_gender(gender)
+    if age:
+        age_range = age.split("-")
+        if len(age_range) > 1:
+            demo["ageRange"] = {
+                "start": float("".join([x for x in age_range[0] if x.isdigit()])),
```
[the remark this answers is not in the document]
attwad: oh, I see in the example below that you can have stuff like `A05-A14`, is there any meaning to `A` ? can we get anything else that would mean something different?

attwad (changes requested):
FYI you'll need to add the template.yaml definition as well for that new parser

ingestion/functions/parsing/germany/germany.py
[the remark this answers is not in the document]
calremmel: no you're right! everything is in the format AXX-AXX, except for 80+ which is A80+. pretty sure A just stands for Alter. I can really just index the string, will make that change

ingestion/functions/parsing/germany/germany.py
calremmel: thanks! do you know which of the parser docs are the most recent? I've been working off of the Swagger API docs (great for most things but no expected values) and the curator bulk upload docs https://github.com/globaldothealth/list/tree/main/verification/curator-service/ui (good expected values but incomplete set of fields)

attwad: if you run a stack locally you can go to localhost:3001/api-docs but I've just reloaded dev with the latest changes from this morning, this should be quite comprehensive, let us know if there's something missing from the API docs there.

https://dev-curator.ghdsi.org/api-docs/

calremmel: this is great, thanks a ton

ingestion/functions/template.yaml
```diff
@@ -109,7 +109,35 @@ Resources:
       Timeout: 300
       Layers:
         - !Ref ParsingLibLayer
-        - !Ref CommonLibLayer
+      Environment:
+        Variables:
+          SOURCE_API_URL:
+            Ref: SourceApiUrl
+          # Workaround for a bug with the existing SAM CLI 1.0 Docker images.
+          # Re: https://github.com/awslabs/aws-sam-cli/issues/2135
```
attwad, line 117: this is not needed anymore and has been removed from other lambdas

calremmel:
Hi all - I think I've addressed everything for this parser. Is there anything left? Also, any thoughts on why the test might be failing? Like Anya, I have mine working locally.

ingestion/functions/parsing/germany/germany.py
```diff
@@ -0,0 +1,105 @@
...
+
+def convert_demographics(gender: str, age: str):
+    demo = {}
+    if gender:
+        demo["gender"] = convert_gender(gender)
+    if age:
+        age_range = age.split("-")
+        if len(age_range) > 1:
+            demo["ageRange"] = {
+                "start": float(age_range[0][1:3]),
```
axmb, line 50: Perhaps toss this into a separate function with a brief explanation (i.e. why the data format necessites `1:3`)?

ingestion/functions/parsing/germany/germany.py
```diff
@@ -0,0 +1,105 @@
...
+                "end": float(age_range[1][1:3]),
+            }
+        else:
+            demo["ageRange"] = {
+                "start": float(age_range[0][1:3]),
+                "end": float(age_range[0][1:3]),
+            }
+    return demo
```
axmb, line 58: Nit, but it'd be slightly better to return `None`, rather than `{}`, in the event where neither gender/age are specified.

ingestion/functions/parsing/germany/germany.py
```diff
@@ -0,0 +1,105 @@
...
+                    ),
+                }
+                # Death and recovery counts are always equal to case counts for the row.
+                if int(row[_DEATHCOUNT]) > 0:
+                    case["events"].append({"name": "outcome", "value": "Death"})
+                elif int(row[_RECOVERYCOUNT]) > 0:
+                    case["events"].append({"name": "outcome", "value": "Recovered"})
+                cases.extend([case] * num_confirmed_cases)
+            except ValueError as ve:
+                print(ve)
```
axmb, line 100: Downside to catching these is that we won't know if we're missing substantial (but human-review-missable) amounts of data. If the quality of this data is such that we need to handle occasional errors like this, that's OK, but we should probably drop a TODO to figure out a monitoring solution for this.

axmb (approved):
Excellent!

ingestion/functions/parsing/germany/germany.py
calremmel: Sounds good! I just tested this on the latest dataset, and it actually doesn't throw any errors at all - just kept this in there following the example of Zurich since it seemed like a good idea just in case. 👍

ingestion/functions/parsing/germany/germany.py
```diff
@@ -0,0 +1,120 @@
+import os
+import sys
+from datetime import date, datetime
+import csv
+
+# Layer code, like parsing_lib, is added to the path by AWS.
+# To test locally (e.g. via pytest), we have to modify sys.path.
+# pylint: disable=import-error
+if "lambda" not in sys.argv[0]:
```
attwad, line 9: please fix this in the same way it was done in the other parsers
try/except

ingestion/functions/parsing/germany/germany.py
```diff
@@ -0,0 +1,120 @@
...
+        cases = []
+        for row in reader:
+            num_confirmed_cases = int(row[_CASECOUNT])
+            if num_confirmed_cases < 1:
+                continue
+            try:
+                case = {
+                    "caseReference": {"sourceId": source_id, "sourceUrl": source_url},
+                    "location": {
+                        "query": ", ".join(("Germany", row[_ADMIN1], row[_ADMIN2])),
```
attwad, line 89: admin1 and admin2 are always present? otherwise you might want to use a list comprehension that filters empty terms

ingestion/functions/parsing/germany/germany.py
attwad: zurich was my first parser, for new ones I actually end up raising ValueErrors more ofter, I think we can raise as well here.

ingestion/functions/template.yaml
```diff
@@ -127,6 +127,19 @@ Resources:
...
+    Type: AWS::Serverless::Function # More info about Function Resource: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
+    Properties:
+      CodeUri: parsing/germany/
+      Handler: germany.lambda_handler
+      Description: Parse case data for Germany
+      # Function's execution role
+      Policies:
+        - AWSLambdaBasicExecutionRole
+        - AWSLambdaReadOnlyAccess
+      MemorySize: 512
```
attwad, line 140: same commenta s the other PR, csv shouldn't require that much memory

ingestion/functions/template.yaml
```diff
@@ -149,8 +162,10 @@ Resources:
       Policies:
         - AWSLambdaBasicExecutionRole
         - AWSLambdaReadOnlyAccess
+      Timeout: 300
       Layers:
         - !Ref ParsingLibLayer
+        - !Ref CommonLibLayer
```
attwad, line 168: please remove

ingestion/functions/template.yaml
```diff
@@ -149,8 +162,10 @@ Resources:
       Policies:
         - AWSLambdaBasicExecutionRole
         - AWSLambdaReadOnlyAccess
+      Timeout: 300
```
attwad, line 165: please remove

ingestion/functions/parsing/germany/germany.py
calremmel: they are always present!

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.formatting.provider": "black"
```
attwad, line 2: add .vscode to .gitignore above please so that these files are not included

attwad (approved):

PR #873, opened 2020-08-26
Complete env-var-less determination of source API endpoint

axmb:
Some repeat commits here from #871 -- apologies, branching got a little goofed.

Edit: I can see some of the tests are slightly goofed. I'd recommend reviewing by commit (`efadf47` onwards), and I'll clean up from the branching fiasco after #871 is merged.


attwad (approved):
lgtm~

PR #877, opened 2020-08-27
Amapa Brazil parser

AnyaLindstromBattle:
Parser for Amapa State of Brazil - passed both unit test and e2e testing. A few things I could do with input on:

1. One of the commorbidities listed is "Portador de doenças cromossômicas ou estado de fragilidade imunológica" which is a combination of two entries in the Disease Ontology (DO) database, namely "primary immunodeficiency disease" OR "chromosomal disease". Is the way I have put this OK or is it better if this leads to the addition of both "primary immunodeficiency disease" and "chromosomal disease" to commorbidities_list, even though it should really be OR not AND?
2. Immunosuppression is not necessarily a disease as defined in the Disease Ontology (i.e. could be drug-induced, or the result of many diseases) so I have put this in notes instead of in pre-exisiting conditions. Thoughts?
3. The data has entries as high as 351 in the age column; I assume this must be days although in the absence of info I have had to exclude these. Thoughts?

Also note I haven't yet adjusted parser for the fact that we are unable to dedupe cases


AnyaLindstromBattle:
Also not quite sure why pytest is failing on here, as worked on my laptop.

axmb:
> 1. One of the commorbidities listed is "Portador de doenças cromossômicas ou estado de fragilidade imunológica" which is a combination of two entries in the Disease Ontology (DO) database, namely "primary immunodeficiency disease" OR "chromosomal disease". Is the way I have put this OK or is it better if this leads to the addition of both "primary immunodeficiency disease" and "chromosomal disease" to commorbidities_list, even though it should really be OR not AND?

Yeah, that's a bit strange. @Mougk -- thoughts on whether a single "disease A or disease B", or a slightly incorrect but more standard double ["disease A", "disease B"] is better, here?

> 2. Immunosuppression is not necessarily a disease as defined in the Disease Ontology (i.e. could be drug-induced, or the result of many diseases) so I have put this in notes instead of in pre-exisiting conditions. Thoughts?

Seems reasonable to me. Another item Moritz's input would be great on.

> 3. The data has entries as high as 351 in the age column; I assume this must be days although in the absence of info I have had to exclude these. Thoughts?

How many cases like this are there? It'd be a weird way to denote that (as days), since, for example, a 52-day old baby would be indistinguishable from a 52-years old adult. I ask how many, since their website lists 458 cases aged 0-5. You could compare totals with and without the assumption above. Plus, if any ages are listed as 0, it seems unlikely that cases like "351" are valid.

AnyaLindstromBattle:
Interesting that the website says 458 cases 0-5 when the downloadable data of 'confirmed cases' has 461 - maybe it's not fully up to date? There are 5 cases > 120 (351, 149, 144, 130, 125); I think given the circumstances it would be best to exclude them

axmb:
> Interesting that the website says 458 cases 0-5 when the downloadable data of 'confirmed cases' has 461 - maybe it's not fully up to date? There are 5 cases > 120 (351, 149, 144, 130, 125); I think given the circumstances it would be best to exclude them

Ah, yeah -- quite a small sample there. Dropping SGTM. If there are any >100 (especially say, 119 or 120) I'd be a little suspicious of those, too.

Mougk:
One of the commorbidities listed is "Portador de doenças cromossômicas ou estado de fragilidade imunológica" which is a combination of two entries in the Disease Ontology (DO) database, namely "primary immunodeficiency disease" OR "chromosomal disease". Is the way I have put this OK or is it better if this leads to the addition of both "primary immunodeficiency disease" and "chromosomal disease" to commorbidities_list, even though it should really be OR not AND?
Yeah, that's a bit strange. 

@Mougk -- thoughts on whether a single "disease A or disease B", or a slightly incorrect but more standard double ["disease A", "disease B"] is better, here?

**MK: If no direct match this should be moved to additional information.** 

Immunosuppression is not necessarily a disease as defined in the Disease Ontology (i.e. could be drug-induced, or the result of many diseases) so I have put this in notes instead of in pre-exisiting conditions. Thoughts?
Seems reasonable to me. Another item Moritz's input would be great on.

**MK: Same as above. Let's move to additional information if no direct match is found. **

ingestion/functions/parsing/brazil_amapa/amapa.py
```diff
@@ -0,0 +1,167 @@
...
+        "start": age,
+        "end": age
+    }
+
+def convert_confirmation_method(raw_test: str):
+    if raw_test == "RT-PCR":
+        return "PCR test"
+    elif "TESTE" in raw_test:
+        return "Serological test"
+    return "Unknown"
```
attwad, line 47: maybe `print(f'unknown confirmation method: {raw_test}')` before returning so that when we look at the logs we know what we're missing.

ingestion/functions/parsing/brazil_amapa/amapa.py
```diff
@@ -0,0 +1,167 @@
...
+        return "Mixed"
+    elif raw_ethnicity == "Amarela":
+        return "Asian"
+    elif raw_ethnicity == "Branca":
+        return "White"
+    elif raw_ethnicity == "Indigena":
+        return "Indigenous"
+    return None
+
+def convert_preexisting_conditions(raw_commorbidities: str):
```
attwad, line 69: you're only doing presence checks in this raw_commorbidities list so make it a set instead please. That's more efficient.

ingestion/functions/parsing/brazil_amapa/amapa.py
```diff
@@ -0,0 +1,167 @@
...
+    preexistingConditions = {}
+    if raw_commorbidities:
+        preexistingConditions["hasPreexistingConditions"] = True
+
+        commorbidities_list = []
+        if "Diabetes" in raw_commorbidities:
+            commorbidities_list.append("diabetes mellitus")
+        if "Portador de doenças cromossômicas ou estado de fragilidade imunológica" in raw_commorbidities:
+            commorbidities_list.append("primary immunodeficiency disease or chromosomal disease")
+            #Is combined DO OK?
```
attwad, line 79: what is DO ?

ingestion/functions/parsing/brazil_amapa/amapa.py
attwad, line 69: This whole method could also be replaced with a map from the raw_commorbitity to our corresponding English value. You'd then just loop through the map and build the commorbidities list with the values that are present.

ingestion/functions/parsing/brazil_amapa/amapa.py
```diff
@@ -0,0 +1,167 @@
...
+            commorbidities_list.append("chronic kidney disease") 
+        if "Doenças cardíacas crônicas" in raw_commorbidities:
+            commorbidities_list.append("heart disease")
+
+        preexistingConditions["values"] = commorbidities_list
+        return preexistingConditions
+    return None
+
+def convert_location(raw_entry: str):
+    if raw_entry:
```
attwad, line 96: you can just 
````
query = ", ".join(word for word in [raw_entry, "Amapá", "Brazil"] if word)
return {"query": query}
```

ingestion/functions/parsing/brazil_amapa/amapa.py
```diff
@@ -0,0 +1,167 @@
...
+        municipality = raw_entry
+    else:
+        municipality = ""
+    query_terms = (municipality,) + ("Amapá", "Brazil")
+    return {"query":  ", ".join(query_terms)}
+
+def convert_notes(raw_commorbidities:str, raw_notes_neighbourhood: str, raw_notes_indigenousEthnicity: str):
```
attwad, line 103: nit, space before str

ingestion/functions/parsing/brazil_amapa/amapa.py
```diff
@@ -0,0 +1,167 @@
...
+    raw_notes = []
+    if "Imunossupressão" in raw_commorbidities:
+        raw_notes.append("Patient with immunosupression")
+        """
```
attwad, line 107: Use a comment, not a string litteral please.

ingestion/functions/parsing/brazil_amapa/amapa.py
```diff
@@ -0,0 +1,167 @@
...
+           to dedupe.
+        2. There are two files in the source for Amapa, one for confirmed cases and one for confirmed
+           deaths; some of these cases may also be deaths but without patient IDs we are unable to confirm.
+    """
+    with open(raw_data_file, "r") as f:
+        reader = csv.reader(f)
+        next(reader) # Skip the header.
+        cases = []
+        for row in reader:
+            if float(row[_AGE_INDEX]) > 120: #We have entries as high as 351 - unclear if this is days.
```
attwad, line 133: that's really fishy, they don't have a data dictionary explaining their data format? we should clarify that

ingestion/functions/parsing/brazil_amapa/input_event.json
```diff
@@ -0,0 +1,6 @@
+{
+    "s3Bucket": "epid-sources-raw",
```
attwad, line 2: you need `env: 'local'` now as well

ingestion/functions/parsing/brazil_amapa/amapa.py
AnyaLindstromBattle: Disease Ontology - see conversation above with Moritz; sorted now.

ingestion/functions/parsing/brazil_amapa/amapa.py
AnyaLindstromBattle: Not that I could find, no.

ingestion/functions/parsing/brazil_amapa/amapa.py
```diff
@@ -0,0 +1,157 @@
+import os
+import sys
+from datetime import date, datetime
+import csv
+
+# Layer code, like parsing_lib, is added to the path by AWS.
+# To test locally (e.g. via pytest), we have to modify sys.path.
+# pylint: disable=import-error
+if ('lambda' not in sys.argv[0]):
```
attwad, line 9: please follow the try/except import method of other parsers, thanks

attwad (commented):
There are still open comments that I made earlier, please address them, thank you

attwad (approved):
Please merge the latest template and I'll merge into main, thanks!

AnyaLindstromBattle:
> Please merge the latest template and I'll merge into main, thanks!

done! thank you

ingestion/functions/template.yaml
```diff
@@ -102,6 +102,20 @@ Resources:
...
+      CodeUri: parsing/brazil_amapa/
+      Handler: amapa.lambda_handler
+      Description: Parse case data for the Amapa State of Brazil
+      # Function's execution role
+      Policies:
+        - AWSLambdaBasicExecutionRole
+        - AWSLambdaReadOnlyAccess
+      Layers:
+        - !Ref ParsingLibLayer
+        - !Ref CommonLibLayer
```
attwad, line 118: remove this one please

ingestion/functions/template.yaml
```diff
@@ -114,6 +128,7 @@ Resources:
         - AWSLambdaReadOnlyAccess
       Layers:
         - !Ref ParsingLibLayer
+        - !Ref CommonLibLayer
```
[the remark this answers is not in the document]
AnyaLindstromBattle: my bad, done

z023:
Data not ingesting.

PR #882, opened 2020-08-28
Don't populate source dateFilters unless we really want them

axmb:
This chokes up parsing, since dateFilter [isn't falsey](https://github.com/globaldothealth/list/blob/main/ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py#L116), but it also doesn't contain necessary fields.


attwad (approved):
good catch

PR #885, opened 2020-08-28
Use batch update endpoint in linelist

allysonjp715:
Endpoint was added in https://github.com/globaldothealth/list/pull/881


attwad (approved):
much cleaner :)

Issue #891, opened 2020-08-28
Make uploads display more digestible


axmb:
I've set up our first real ingestion source in prod, and there are a couple of sharp edges around the process. One is how uploads are displayed and shown. We see an upload for every run of automated ingestion -- regardless if no cases were actually modified.

In the Cases API batchUpsert method, we don't distinguish between "a case existed and the requested update was applied" and "a case was _actually_ modified." This wasn't possible to do while still furnishing the feature where a user would be automatically redirected to the linelist following a bulk upload -- we needed the IDs of impacted cases to do that. Now that we have the upload ID, however, we can search for cases that way. This critically allows us to just return aggregate #s from batchUpsert, which _then_ allows us to use MongoDB data indicated the total number _actually_ modified.

This issue tracks making uploads more digestible.

axmb:
Fixed. See:

#892 
#899 
#900 
#902
#913 
#976

PR #892, opened 2020-08-28
Make verificationStatus default to UNVERIFIED.

axmb:
Required for #891. This prevents a case from appearing modified if it's been previously verified by a curator, and no new data is found in the ADI run.

This unfortunately won't set a case to `UNVERIFIED` if new information is found in an update, which might be nice, but I'd reckon that the vast majority of upserts of existing cases will _not_ contain new information (so this behavior is desirable), and that searching for verified cases within an upload is a reasonable workaround.

### DIFF
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -429,7 +429,8 @@ components:
             verificationStatus:
               description: >
                 Whether the case document has been manually verified for
-                correctness.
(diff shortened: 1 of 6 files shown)


attwad (approved):
Looks good, just wonder if we should document that in the open API spec as well.

axmb:
> Looks good, just wonder if we should document that in the open API spec as well.

Good call, done.

PR #893, opened 2020-08-28
Clean up or suppress warnings/errors in preprocessor.ts

axmb:
Separating this out as not to confound some content changes I'm making.

The @ts-ignore error doesn't seem to be new, looking at the config; though I can't say for sure, as I don't think I've edited this file.


attwad (approved):
taking this on as it's a UK holiday today

Issue #754, opened 2020-08-10
Make it feasible to run ingestion functions locally without relying on s3 storage


attwad:
To make it easy for new contributors without access to the s3 bucket hosting the service account creds to run a function locally.
Ideally contributors can just use their own account when testing locally.

PR #896, opened 2020-08-31
make local runs of ingestion functions able to register a user and use it

attwad:
First step for #754, will tackle the parsing functions next.

This also includes a quick fix for the utf-8 encoding of sources, turns out one cannot just "decode" utf-8 into a utf-8 file if there are surrogates already present, instead if the source is already utf-8 I just dump it as is, no need to do the decoding/reencoding step that's simpler and faster and we just get the raw source as is this way. (Noticed the issue when trying on Estonian data source).

### DIFF
--- a/ingestion/functions/README.md
+++ b/ingestion/functions/README.md
@@ -49,7 +49,23 @@ Serverless Application Model
 Python and executed on a version 3.8 runtime. See set up instructions and
 common commands, below.
 
-### One-time setup
+### Setup for folks without AWS access
+
+If you're a first-time contributor to the project and don't have access to the S3 bucket containing the service account keys, you can run the ingestion and parsing functions fully locally, in the `retrieval/valid_scheduled_event.json`, add this auth param to it:
+
+```json
+"auth": {
+   "email": "local@ingestion.function"
+}
+```
+
+This will make the functions log-in as a new user specified by this email and use the cookies generated for this user instead of the service account creds stored on S3.
+
+Note that this only works in a local environment as the handler to register a user isn't exposed in production for obvious reasons.
+
+TODO: #754 Add param to store and retrieve source content locally, not on S3.
+
+### One-time setup for people with AWS access
 
 #### Prerequisites
 
--- a/ingestion/functions/common/common_lib.py
+++ b/ingestion/functions/common/common_lib.py
@@ -37,12 +37,13 @@ class UploadError(Enum):
     VALIDATION_ERROR = 8
 
 
-def create_upload_record(env, source_id, headers):
+def create_upload_record(env, source_id, headers, cookies):
     """Creates an upload resource via the G.h Source API."""
     post_api_url = f"{get_source_api_url(env)}/sources/{source_id}/uploads"
     print(f"Creating upload via {post_api_url}")
     res = requests.post(post_api_url,
                         json={"status": "IN_PROGRESS", "summary": {}},
+                        cookies=cookies,
                         headers=headers)
     if res and res.status_code == 201:
         res_json = res.json()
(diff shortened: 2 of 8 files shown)


ingestion/functions/README.md
```diff
@@ -49,7 +49,23 @@ Serverless Application Model
 Python and executed on a version 3.8 runtime. See set up instructions and
 common commands, below.
 
-### One-time setup
+### Setup for folks without AWS access
+
+If you're a first-time contributor to the project and don't have access to the S3 bucket containing the service account keys, you can run the ingestion and parsing functions fully locally, in the `retrieval/valid_scheduled_event.json`, add this auth param to it:
```
axmb, line 54: This is really convenient -- did you try running the `sam local invoke [...]` command without credentials for our project, too? I can't recall exactly what permissions are required for the local invocations, but if we can completely remove giving AWS credentials from the equation, that's super nice.

ingestion/functions/common/common_lib.py
```diff
@@ -85,11 +89,28 @@ def complete_with_error(
...
     raise exception
 
 
+def login(email: str):
+    """Logs-in a local curator server instance for testing.
+
+    Returns the cookie of the now logged-in user.
+    """
+    print('Logging-in user', email)
+    endpoint = "http://localhost:3001/auth/register"
```
axmb, line 103: I'm undecided as to whether this is right, or whether we should use the utility function, and have `login` check that `env` == local.

ingestion/functions/common/common_lib.py
```diff
@@ -65,18 +66,21 @@ def finalize_upload(
         "summary": {"numCreated": count_created, "numUpdated": count_updated}}
     res = requests.put(put_api_url,
                        json=update,
-                       headers=headers)
+                       headers=headers,
+                       cookies=cookies)
     # TODO: Look for "errors" in res_json and handle them in some way.
+    # TODO: This can end up in a error-loop where we try to upload and fail
+    #       endlessly
```
axmb, line 73: Thanks for getting this out of my :brain: and into the :computer:.

ingestion/functions/common/common_lib.py
attwad: FWIW I am checking in the main that env == 'local' before calling login()

ingestion/functions/README.md
attwad: ah good call, yep just did it now (after renaming my ~/.aws folder temporarily).
`sam build` works fine.
`sam local invoke` of the retrieval function works as expected as well (fetches ok, fails to upload to s3 currently until I work on the local storage for testing):

```
{"errorType":"S3UploadFailedError","errorMessage":"Failed to upload /tmp/content.csv to epid-sources-raw/5f4d0a0a650106002eb179b8/2020/08/31/1442/content.csv: An error occurred (InvalidAccessKeyId) when calling the CreateMultipartUpload operation: The AWS Access Key Id you provided does not exist in our records."
```

I'll work on that once I get the parsing functions to also use the local cookies.

ingestion/functions/common/common_lib.py
attwad: nw, that scared me a bit when I noticed... might tackle that now actually.

attwad:
Added a small optimization where we don't fetch the content on unsupported format.

ingestion/functions/common/common_lib.py
axmb: Yeah, I don't think there's a practical difference either way. In fact, what you have might be better, since the common lib appends `/api` to the base URL. Only downside is port changes, but that feels pretty minor.

PR #897, opened 2020-08-31
add moar commentz

attwad:
- document (all ?) methods in curator service
- fix tslint errors/warnings in frontend as I went along
- try to use [tsdoc](https://github.com/microsoft/tsdoc) as much as I could, I did not convert all existing one-liners though as it was too tedious, perhaps a vscode extension can do that.

### DIFF
--- a/verification/curator-service/api/src/controllers/auth.ts
+++ b/verification/curator-service/api/src/controllers/auth.ts
@@ -6,14 +6,18 @@ import {
 import { NextFunction, Request, Response } from 'express';
 import { User, UserDocument } from '../model/user';
 
-import { Strategy as BearerStrategy } from 'passport-http-bearer';
+import {
+    Strategy as BearerStrategy,
+    IVerifyOptions,
+} from 'passport-http-bearer';
 import { Router } from 'express';
 import axios from 'axios';
 import passport from 'passport';
 
-// mustBeAuthenticated is a middleware that checks that the user making the call is
-// authenticated.
-// Subsequent request handlers can be assured req.user will be defined.
+/**
+ * mustBeAuthenticated is a middleware that checks that the user making the call is authenticated.
+ * Subsequent request handlers can be assured req.user will be defined.
+ */
 export const mustBeAuthenticated = (
     req: Request,
     res: Response,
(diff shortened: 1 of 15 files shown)


verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -6,14 +6,18 @@ import {
...
 import { Router } from 'express';
 import axios from 'axios';
 import passport from 'passport';
 
-// mustBeAuthenticated is a middleware that checks that the user making the call is
-// authenticated.
-// Subsequent request handlers can be assured req.user will be defined.
+/**
+ * mustBeAuthenticated is a middleware that checks that the user making the call is authenticated.
+ * Subsequent request handlers can be assured req.user will be defined.
```
axmb, line 19: Should subsequent text beyond the one-line summary be `@remarks`?

verification/curator-service/api/src/controllers/cases.ts
```diff
@@ -152,6 +160,11 @@ export default class CasesController {
         }
     };
 
+    /**
+     * udpate will try to geocode the location found in the request if needed
```
axmb, line 164: upsert

But if you wanted, you could probably just start with "Geocodes the [...]" since the function name is clear. Including the name is fine by me, especially if it's conventional in tsdoc.

verification/curator-service/api/src/controllers/auth.ts
attwad: I don't understand when `@remarks` is supposed to be used... is this documented anywhere?
In doubt I decided not to use it at all ;-) to me everything is a remark somehow...

verification/curator-service/api/src/controllers/cases.ts
attwad: Ah sorry that's my Golang, all comments have to start with the name of the thing that they are commenting.
That's because godoc (a cli to generate and extract docs) becomes very useful when you do this (https://blog.golang.org/godoc).
As I did the same for all other methods I just changed to upsert for consistency, thanks for catching it.

axmb: Plus, this way if the code ever gets rewritten in Go, we won't even have to change the comments.

attwad: don't tempt me!! ;-) I got more than enough time left in this fellowhsip to do it!!!

PR #898, opened 2020-08-31
zero pad month and day when rendering dates

attwad:
Don't know about you but single digit day or months look incomplete to me.

Notice how I am *not* using any leftpad library to do that, pure magic.

https://en.wikipedia.org/wiki/Npm_(software)#Notable_breakages

### DIFF
--- a/verification/curator-service/ui/src/components/util/date.test.tsx
+++ b/verification/curator-service/ui/src/components/util/date.test.tsx
@@ -1,0 +1,9 @@
+import renderDate from './date';
+
+describe('Dates', () => {
+    it('are padded', () => {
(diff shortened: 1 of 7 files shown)


verification/curator-service/ui/src/components/util/date.test.tsx
```diff
@@ -0,0 +1,9 @@
+import renderDate from './date';
+
+describe('Dates', () => {
+    it('are padded', () => {
+        expect(renderDate(new Date(Date.UTC(2020, 6, 9)))).toEqual(
+            '2020-07-09',
```
axmb, line 6: I just can't believe `Date.UTC` uses a zero-based month.

attwad: well, at least it's consistently bad.

PR #899, opened 2020-08-31
Add an instance method to case model for determining semantic-JSON-equivalence

axmb:
Semantic meaning "has the content meaningfully changed." Using `lodash` for the semantic `isEqual` method.


data-serving/data-service/test/model/case.test.ts
```diff
@@ -38,3 +38,15 @@ describe('validate', () => {
         return new Case(fullModel).validate();
     });
 });
+
+describe('custom instance methods', () => {
+    it('equalsJSON returns true for identical case', () => {
+        const c = new Case(fullModel);
+        expect(c.equalsJSON(fullModel)).toBe(true);
+    });
+    it('equalsJSON returns false for differing case', () => {
```
attwad, line 47: please add a test where revision metadata or another irrelevant field is changed and expect equalsJSON to be true

attwad (approved):

data-serving/data-service/test/model/case.test.ts
axmb: SG, done.

PR #900, opened 2020-08-31
Don't rev revisionMetadata if a case is unchanged in batchUpsert.

axmb:


axmb:
Missed requesting a review here, apologies.

data-serving/data-service/src/controllers/preprocessor.ts
```diff
@@ -227,7 +253,7 @@ export const createBatchUpsertCaseRevisions = async (
     await CaseRevision.insertMany(casesToUpsert, {
         ordered: false,
         rawResult: true,
-        // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
```
attwad, line 256: funny how you have to disable linting over a comment...

attwad (approved):
my bad for not noticing

PR #902, opened 2020-09-01
Don't create revisions if a case is unchanged in batchUpsert.

axmb:
After writing this, I considered instead filtering the value of `request.body.cases` in the revisionMetadata middleware, which seems promising. I'll take a look at that in a different PR.

### DIFF
--- a/data-serving/data-service/src/controllers/preprocessor.ts
+++ b/data-serving/data-service/src/controllers/preprocessor.ts
@@ -116,13 +116,16 @@ export const setBatchUpsertRevisionMetadata = async (
                         c.caseReference.sourceEntryId,
                 );
                 if (existingCase !== undefined && existingCase.equalsJSON(c)) {
-                    unchangedIds.add(existingCase._id);
+                    unchangedIds.add(existingCase._id.toString());
                 }
             }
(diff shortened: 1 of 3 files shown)


data-serving/data-service/src/controllers/preprocessor.ts
```diff
@@ -116,22 +116,25 @@ export const setBatchUpsertRevisionMetadata = async (
...
+                    unchangedIds.add(existingCase._id.toString());
                 }
             }
             return unchangedIds;
         },
         new Set(),
     );
 
+    // Store the unchanged IDs for future middleware.
+    response.locals.unchangedCaseIdSet = unchangedCaseIdSet;
```
attwad, line 128: nit, we're not using views so putting locals in request.locals feels a bit more appropriate but no big deal either way

attwad (approved):

data-serving/data-service/src/controllers/preprocessor.ts
axmb: For some reason, in the Express dts (`express-serve-static-core`), `locals` is only present on Response and Application. /shrug

PR #903, opened 2020-09-01
do not loop endlessly on error-writing failure

attwad:
`finalize_upload` was calling `finalize_upload` if it failed to finalize the upload. nope.

Also removes TODO handling of `res.json().errors` field as well as the text is contained in the runtime exception we throw in each case or the API just doesn't provide fine grained errors like this (uploads API for example).

### DIFF
--- a/ingestion/functions/common/common_lib.py
+++ b/ingestion/functions/common/common_lib.py
@@ -47,9 +47,6 @@ def create_upload_record(env, source_id, headers, cookies):
                         headers=headers)
     if res and res.status_code == 201:
         res_json = res.json()
-        # Upload POST API doesn't provide 'errors' json in response body, all
-        # we have is the response body itself which is enough to describe the
-        # error.
         return res_json["_id"]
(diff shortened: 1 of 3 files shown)


ingestion/functions/common/common_lib.py
```diff
@@ -47,7 +47,9 @@ def create_upload_record(env, source_id, headers, cookies):
                         headers=headers)
     if res and res.status_code == 201:
         res_json = res.json()
-        # TODO: Look for "errors" in res_json and handle them in some way.
+        # Upload POST API doesn't provide 'errors' json in response body, all
+        # we have is the response body itself which is enough to describe the
+        # error.
```
axmb, line 52: Sorry, having a little trouble grokking this. Meaning `res.text` tells us the error, or?

attwad: yes, but re-reading this, those comments server no purpose really it's more confusing than without them. Removed.

PR #905, opened 2020-09-01
Suggest parsing function names in UI

### LINKED ISSUE #904: Add curator service endpoint to list available parsers
One step to setup automated ingestion requires curators to specify a parser ARN which can be quite tedious to find & copy.
We should just provide the available functions in the same way we are able to suggest other things when entering cases.

attwad:
Fixes #904 

Kind of big PR as I bundled the API and UI together hope you don't mind, at least it's all in one place :)

This also removes the "function must be referenced by their ARN" as the aws libraries are quite permissive they allow function names to be used instead which are much more readable and better to display in a suggest box, changed to that instead and removed the openapi and model regex checks.


verification/curator-service/api/src/model/parser.ts
```diff
@@ -3,8 +3,7 @@ import mongoose from 'mongoose';
 export const parserSchema = new mongoose.Schema({
     awsLambdaArn: {
```
axmb, line 4: Is it worth changing this field to awsLambdaName?

verification/curator-service/api/src/model/parser.ts
```diff
@@ -3,8 +3,7 @@ import mongoose from 'mongoose';
...
         type: String,
-        required: 'Enter a parser AWS Lambda ARN',
-        match: [/^arn\:aws\:lambda\:.+\:.+\:function\:.+/],
+        required: 'Enter a parser AWS Lambda ARN or function name',
```
axmb, line 6: It'd be good to reference the documentation that describes what a valid function name is, here and/or in the OpenAPI spec (e.g., meaning it can be ARN, semantic name, perhaps something else).

verification/curator-service/ui/src/components/ParsersAutocomplete.tsx
```diff
@@ -0,0 +1,93 @@
...
+        if (!open) {
+            setOptions([]);
+        }
+    }, [open]);
+
+    return (
+        <Autocomplete
+            style={{ width: 300 }}
+            open={open}
+            // freeSolo useful in tests or if a parsing function wasn't built yet.
```
axmb, line 56: Not sure it's worth adding if it's just for tests. Is testing a PITA without this? I don't think we'll want to allow preemptive setting of the parser, either.

verification/curator-service/ui/src/components/ParsersAutocomplete.tsx
```diff
@@ -0,0 +1,93 @@
...
+            }}
+            defaultValue={props.defaultValue}
+            onChange={(e, val) => props.onChange(val || '')}
+            onInputChange={(e, val) => props.onChange(val)}
+            options={options}
+            loading={loading}
+            renderInput={(params) => (
+                <TextField
+                    {...params}
+                    //label="Parsers"
```
axmb, line 73: nit

verification/curator-service/api/src/clients/aws-lambda-client.ts
```diff
@@ -103,4 +107,23 @@ export default class AwsLambdaClient {
...
+
+    /** Lists the available parser */
+    listParsers = async (): Promise<LambdaFunction[]> => {
+        try {
+            const res = await this.lambdaClient
+                .listFunctions({ MaxItems: 10000 })
+                .promise();
+            return (
+                res.Functions?.filter((f) =>
+                    f.FunctionName?.includes('ParsingFunction'),
```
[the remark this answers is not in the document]
attwad: I added a sentence in the readme yes, good call.

verification/curator-service/api/src/model/parser.ts
attwad: I think it's going to break all existing parsers, if we're fine with that then sure, but maybe in another PR.

verification/curator-service/api/src/model/parser.ts
attwad: I don't there there is a spec for function names, I added a real example in the openapi spec already but AFAICT this name is built by SAM based on the stuff we put in the template and the project overall (prefix is fixed, then function name from template, then random digitsletters)

verification/curator-service/ui/src/components/ParsersAutocomplete.tsx
attwad: testing is unfeasible without this as we don't currently mock the calls to the lambda service they just fail and the options are empty.
Maybe it's time we think about how to properly do that in integration tests as we did for geocoding (providing a fake might make sense)

verification/curator-service/ui/src/components/ParsersAutocomplete.tsx
attwad: oops, yeah I ended up using placeholder instead of label and forgot to remove this, thanks.

attwad:
> No gif? :-(

alright alright 

ezgif com-gif-maker

axmb:
> > No gif? :-(
> 
> alright alright
> 
> ezgif com-gif-maker

👏 👌 

Do we want to filter out the "epid-ingestion" prefix, and the ~UUID suffix?

attwad:
> > > No gif? :-(
> > 
> > 
> > alright alright
> > ezgif com-gif-maker
> 
> 👏 👌
> 
> Do we want to filter out the "epid-ingestion" prefix, and the ~UUID suffix (for display purposes/in the UI)?

I feel uneasy hardcoding those magic strings that come from SAM in here, I think those are short enough that they are fine as is.

PR #913, opened 2020-09-02
Return counts, with true modification numbers, from batchUpsert

axmb:
Slight change in behavior for post-bulk-upload, which I broached with Aaron/Shroo.

On a minor note, we're saving one query in the nominal case for ADI (batchUpsert with many existing cases), which seems to shave off about 10-15% of the total time.

### DIFF
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -473,7 +473,11 @@ components:
           description: The number of cases created by the request.
         numUpdated:
           type: number
-          description: The number of cases modified by the request.
+          description: >
+            The number of cases modified by the request. Note that this only
+            includes cases that are actually modified by the batch upsert. If a
+            case is included in the batch upsert request, but contains no
+            changes compared to the currently stored case, it won't be counted.
       required:
         - numCreated
         - numUpdated
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -468,17 +468,15 @@ components:
     BatchUpsertCaseResponse:
       description: Response to batch upsert case API requests
       properties:
-        createdCaseIds:
-          type: array
-          items:
-            type: string
-        updatedCaseIds:
-          type: array
-          items:
-            type: string
+        numCreated:
+          type: number
+          description: The number of cases created by the request.
+        numUpdated:
+          type: number
+          description: The number of cases modified by the request.
       required:
-        - createdCaseIds
-        - updatedCaseIds
+        - numCreated
+        - numUpdated
     Case:
       description: A single line-list case.
       properties:
--- a/data-serving/data-service/src/controllers/case.ts
+++ b/data-serving/data-service/src/controllers/case.ts
@@ -252,7 +252,6 @@ export const batchUpsert = async (
     res: Response,
 ): Promise<void> => {
     try {
-        const toBeUpsertedCaseIds = await findCaseIdsWithCaseReferenceData(req);
         const bulkWriteResult = await Case.bulkWrite(
             // Case data should be validated prior to this point.
             // eslint-disable-next-line @typescript-eslint/no-explicit-any
@@ -284,13 +283,10 @@ export const batchUpsert = async (
             { ordered: false },
         );
         res.status(207).json({
-            // Types are a little goofy here. We're grabbing the string ID from
-            // what MongoDB returns, which is data in the form of:
-            //   { index0 (string): _id0 (ObjectId), ..., indexN: _idN}
-            createdCaseIds: Object.entries(bulkWriteResult.insertedIds)
-                .concat(Object.entries(bulkWriteResult.upsertedIds))
-                .map((kv) => String(kv[1])),
-            updatedCaseIds: toBeUpsertedCaseIds,
+            numCreated:
+                (bulkWriteResult.insertedCount || 0) +
+                (bulkWriteResult.upsertedCount || 0),
+            numUpdated: res.locals.numModified,
         });
         return;
     } catch (err) {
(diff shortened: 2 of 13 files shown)


verification/curator-service/ui/src/components/UploadsTable.tsx
```diff
@@ -156,8 +156,21 @@ class UploadsTable extends React.Component<Props, UploadsTableState> {
                                 );
                                 response
                                     .then((result) => {
-                                        const flattenedSources = result.data.uploads.map(
-                                            (u) => {
+                                        const flattenedSources = result.data.uploads
```
allysonjp715, line 159: This change seems unrelated to this PR - the uploads table doesn't use the batchUpsert call. I also don't think this is the right way to achieve this outcome. It won't work correctly with paging e.g. if the server returns 50 uploads for the first page but we filter out 5 rows, it will show 45 rows in the current page but also have a next page button.

Perhaps remove this file from this PR? Then you or I can have a look at doing this on the server side.

verification/curator-service/ui/src/components/UploadsTable.tsx
```diff
@@ -170,12 +183,12 @@ class UploadsTable extends React.Component<Props, UploadsTableState> {
...
                                                 };
-                                            },
-                                        );
+                                            });
                                         resolve({
                                             data: flattenedSources,
                                             page: query.page,
-                                            totalCount: result.data.total ?? 0,
+                                            totalCount:
+                                                flattenedSources.length ?? 0,
```
allysonjp715, line 191: This needs to remain as `result.data.total ?? 0`. This is the totalCount of all uploads in case there is more than one page. `flattenedSources` only contains the first page of results.

data-serving/data-service/src/controllers/case.ts
```diff
@@ -284,13 +283,10 @@ export const batchUpsert = async (
...
         res.status(207).json({
-            // Types are a little goofy here. We're grabbing the string ID from
-            // what MongoDB returns, which is data in the form of:
-            //   { index0 (string): _id0 (ObjectId), ..., indexN: _idN}
-            createdCaseIds: Object.entries(bulkWriteResult.insertedIds)
-                .concat(Object.entries(bulkWriteResult.upsertedIds))
-                .map((kv) => String(kv[1])),
-            updatedCaseIds: toBeUpsertedCaseIds,
+            numCreated:
+                (bulkWriteResult.insertedCount || 0) +
```
allysonjp715, line 287: I thought the fields on BulkWriteResults were `nInserted` and `nUpserted`? https://docs.mongodb.com/manual/reference/method/BulkWriteResult/

data-serving/data-service/src/controllers/case.ts
```diff
@@ -284,13 +283,10 @@ export const batchUpsert = async (
...
+                (bulkWriteResult.upsertedCount || 0),
+            numUpdated: res.locals.numModified,
```
allysonjp715, line 289: Why not use `bulkWriteResult.nModified` ? 

"The number of existing documents updated or replaced. If the update/replacement operation results in no change to an existing document, such as setting the value of the field to its current value, nModified can be less than nMatched. Inserted documents do not affect the number of nModified; refer to the nInserted and nUpserted fields instead."
https://docs.mongodb.com/manual/reference/method/BulkWriteResult/

data-serving/data-service/api/openapi.yaml
```diff
@@ -468,17 +468,15 @@ components:
...
-        updatedCaseIds:
-          type: array
-          items:
-            type: string
+        numCreated:
+          type: number
+          description: The number of cases created by the request.
+        numUpdated:
+          type: number
+          description: The number of cases modified by the request.
```
allysonjp715, line 476: Perhaps we should add to this description that this number does not include cases that were matched but unchanged

verification/curator-service/ui/src/components/BulkCaseForm.tsx
```diff
@@ -523,30 +523,27 @@ class BulkCaseForm extends React.Component<
...
                 ? '1 case updated. '
-                : `${updatedIds.length} cases updated. `;
+                : `${upsertResponse.numUpdated} cases updated. `;
         this.props.history.push({
             pathname: '/cases',
             state: {
                 bulkMessage: `${filename} uploaded. ${createdMessage} ${updatedMessage}`,
-                newCaseIds: createdIds,
-                editedCaseIds: updatedIds,
+                searchQuery: `uploadid:${uploadId}`,
```
allysonjp715, line 546: Will this search result include cases that were matched but not updated?

verification/curator-service/ui/src/components/UploadsTable.tsx
axmb: SG, I'll remove the file.

data-serving/data-service/src/controllers/case.ts
axmb: It's a `BulkWriteOpResult`, which is slightly different:

http://mongodb.github.io/node-mongodb-native/3.1/api/Collection.html#~BulkWriteOpResult
https://mongoosejs.com/docs/api/model.html#model_Model.bulkWrite

data-serving/data-service/src/controllers/case.ts
axmb: I'd anticipated using this, but then noticed that the document's technically modified -- we're still updating the `upload`. I don't think we want a revision for this, though, particularly after we change upload to be an array.

data-serving/data-service/api/openapi.yaml
axmb: SG, done.

verification/curator-service/ui/src/components/BulkCaseForm.tsx
axmb: It will -- the behavior is ~unchanged as compared to before this PR. We have the option, after this, to remove adding the upload ID to unchanged cases: which will help with the verification process.

allysonjp715 (approved):

PR #922, opened 2020-09-03
document method of test

attwad:


allysonjp715 (approved):

allysonjp715 (approved):
Is this in the data-service openapi already?

attwad:
> Is this in the data-service openapi already?

no, the data-service api isn't as well documented as the curator service one as it's redundant and not exposed in dev/prod, I am tempted to just leave it as is.

PR #924, opened 2020-09-03
Paraiba Brazil parser

AnyaLindstromBattle:
Parsing files for Paraiba State of Brazil. A few things it would be good to have input on:
1. I have translated "Doença Hematológica" to "hematopoietic system disease", even though "hematologic disease" is the direct translation but is not present in DO database. I realise this is not 100 % accurate though, so do you think I should put this in notes? @Mougk would be useful to have your thoughts on this.
2. In the case of pre-exisiting conditions, the entry is sometimes 'Others'. As it stands this returns an unpopulated preexistingConditions, but do you think it would be more accurate to do preexistingConditions["hasPreexistingConditions"] = True and then leave values empty?


Mougk:
_I have translated "Doença Hematológica" to "hematopoietic system disease", even though "hematologic disease" is the direct translation but is not present in DO database. I realise this is not 100 % accurate though, so do you think I should put this in notes? @Mougk would be useful to have your thoughts on this._

If no direct translation that is present in the DO database this should go to Notes (or others if that exists?)

_In the case of pre-exisiting conditions, the entry is sometimes 'Others'. As it stands this returns an unpopulated preexistingConditions, but do you think it would be more accurate to do preexistingConditions["hasPreexistingConditions"] = True and then leave values empty?_

I'd recommend to put "pre-existing conditions" == YES and leave the condition empty. Notes - unspecified pre-existing condition.

ingestion/functions/parsing/brazil_paraiba/paraiba.py
```diff
@@ -0,0 +1,193 @@
+import os
+import sys
+from datetime import date, datetime
+import csv
+
+# Layer code, like parsing_lib, is added to the path by AWS.
+# To test locally (e.g. via pytest), we have to modify sys.path.
+# pylint: disable=import-error
+if ('lambda' not in sys.argv[0]):
```
attwad, line 9: please replace that with

```
try:
    import parsing_lib
except ImportError:
    sys.path.append(
        os.path.join(
            os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
            'common/python'))
    import parsing_lib
```

ingestion/functions/parsing/brazil_paraiba/paraiba.py
```diff
@@ -0,0 +1,193 @@
...
+        return "Female"
+    return None
+
+def convert_age(age: str):
+    if age.isdecimal():
+        return {
+            "start": float(age),
+            "end": float(age)
+        }
+    else:
```
attwad, line 47: Comment on what format is the age if not all decimal please, give an example.

ingestion/functions/parsing/brazil_paraiba/paraiba.py
```diff
@@ -0,0 +1,193 @@
...
+        return "Unknown"
+
+def convert_preexisting_conditions(raw_commorbidities: str):
+    preexistingConditions = {}
+    if raw_commorbidities != "Sem comorbidades" or "Outros" or "Tabagismo" or "Imunossupressão":
+    	#How do we want to deal with 'Others' - do we want to have hasPreexistingConditions = True
+    	#and not populate the values?
+        preexistingConditions["hasPreexistingConditions"] = True
+        commorbidities = {
+                        "Diabetes Mellitus": "diabetes mellitus",
```
attwad, line 90: I recommend to run autopep8 on your code (it's a simple plugin in vscode if you use it) to format the code nicely.

ingestion/functions/parsing/brazil_paraiba/paraiba.py
```diff
@@ -0,0 +1,193 @@
...
+
+def convert_notes(raw_commorbidities: str):
+    raw_notes = []
+    if "Imunossupressão" in raw_commorbidities:
+        raw_notes.append("Patient with immunosupression")
+    if "Tabagismo" in raw_commorbidities:
+        raw_notes.append("Smoker")
+    notes = (', ').join(raw_notes)
+    
+    if notes != '':
```
attwad, line 130: you can just return an empty string for notes, it doesn't matter in the end (same as None).

ingestion/functions/parsing/brazil_paraiba/paraiba.py
```diff
@@ -0,0 +1,193 @@
...
+                        }, 
+                        "value" : "Death"                       
+                    }
+                ],
+                "preexistingConditions": convert_preexisting_conditions(row[_PREEXISTING_CONDITIONS_INDEX]),
+                "notes": convert_notes(row[_PREEXISTING_CONDITIONS_INDEX])
+            }
+            cases.append(case)
+        # TODO(AB, Anya): Handle ensuring None fields aren't sent in requests.
+        return [c for c in cases if c["preexistingConditions"] is not None]
```
attwad, line 190: `is not None` is not needed.

ingestion/functions/parsing/brazil_paraiba/paraiba.py
AnyaLindstromBattle: Do you mean in the script or here?

attwad: in the script, for future readers of this parser (people who will want to understand how data was parsed won't look at this PR but at the code), thank you.

AnyaLindstromBattle: nice one, done.

ingestion/functions/parsing/brazil_paraiba/paraiba.py
```diff
@@ -0,0 +1,191 @@
...
+        return "Serological test"  
+    else:
+        print(f'unknown confirmation method: {raw_test}')
+        return "Unknown"
+
+def convert_preexisting_conditions(raw_commorbidities: str):
+    preexistingConditions = {}
+    if raw_commorbidities not in ["Sem comorbidades","Doença Hematológica","Tabagismo","Imunossupressão"]:
+        preexistingConditions["hasPreexistingConditions"] = True
+        commorbidities = {
```
attwad, line 91: you can define this outside of the function as commented in the other PR

ingestion/functions/parsing/brazil_paraiba/paraiba.py
```diff
@@ -0,0 +1,191 @@
...
+           to dedupe.
+        2. This data only includes deaths from Covid-19 so the outcome of all cases will be death.
+        3. There is no date of confirmation.
+    """
+    with open(raw_data_file, "r") as f:
+        reader = csv.reader(f)
+        next(reader) # Skip the header.
+        cases = []
+        for row in reader:
+            if row[_DATE_SYMPTOMS_INDEX] < "2019-11-01": #One date is recorded as year 2000
```
attwad, line 148: you need to parse dates to do this, this will do a string comparison which will use length and not the actual values in the string.

attwad (approved):
Thanks!

Issue #907, opened 2020-09-01
Upload search functionality returns no results


sratcliffe118:
Curator portal: Production environment 
User journey: Checking an upload and want to search for all the cases created and run some spot checks

Issue: Clicked on the number of new cases created, this link takes you to the linelist tab and populates the search with the upload ID.  No line list results are shown for search 

uploadid:5f485eff867f293c4ec809d8

PR #927, opened 2020-09-03
Store all upload ids for a case

allysonjp715:
#907

Also adds logic to drop cases that would be unchanged from the batch upsert logic.


data-serving/data-service/schemas/cases.indexes.json
```diff
@@ -31,12 +31,12 @@
...
-        "name": "uploadIdIdx",
+        "name": "uploadIdsIdx",
         "key": {
-            "caseReference.uploadId": -1
+            "caseReference.uploadIds": -1
         },
         "partialFilterExpression": {
-            "caseReference.uploadId": {
+            "caseReference.uploadIds": {
                 "$exists": true
```
axmb, line 40: Since this is an array, do we now need to check that the length of the array is >0, rather than it exists?

data-serving/data-service/src/controllers/preprocessor.ts
```diff
@@ -102,58 +101,82 @@ export const setBatchUpsertRevisionMetadata = async (
...
-            if (c.caseReference?.sourceId && c.caseReference?.sourceEntryId) {
-                const existingCase = existingCasesByCaseRefCombo.get(
-                    c.caseReference.sourceId +
-                        ':' +
-                        c.caseReference.sourceEntryId,
-                );
-                if (existingCase !== undefined && existingCase.equalsJSON(c)) {
-                    unchangedIds.add(existingCase._id.toString());
-                }
+    const unchangedCaseIdSet = new Set();
```
axmb, line 104: Is this used, anymore?

data-serving/data-service/src/controllers/preprocessor.ts
```diff
@@ -102,58 +101,82 @@ export const setBatchUpsertRevisionMetadata = async (
...
+    request.body.cases = request.body.cases.filter((c: any) => {
```
axmb, line 105: Could we filter this in-place (remove array items manually)? I almost never really worry about memory usage, but this is potentially double-digit MB, and we'd previously seen OOMs developing locally for 50k-100k cases. IIU[Javascript]C, `.filter` will always allocate a second array, so we'll have a peak of 2xNumCases in memory here.

data-serving/data-service/src/controllers/preprocessor.ts
```diff
@@ -102,58 +101,82 @@ export const setBatchUpsertRevisionMetadata = async (
...
+            .map((c) => [
+                c.caseReference.sourceId + ':' + c.caseReference.sourceEntryId,
+                c,
+            ]),
+    );
+
     // Store modification details for response handler.
-    response.locals.numModified =
-        existingCasesByCaseRefCombo.size - unchangedCaseIdSet.size;
+    response.locals.numModified = existingCasesByCaseRefCombo.size;
```
axmb, line 143: Now that we're removing cases that only differed by upload -- can we use MongoDB's numModified (in the CaseController)? IIRC, `uploadId` was the only piece of information that prevented us from doing so. Rather than pass this via middleware.

I can take care of this in a separate PR if you'd like.

data-serving/data-service/src/index.ts
```diff
@@ -93,7 +94,8 @@ new OpenApiValidator({
         apiRouter.post('/cases/batchValidate', caseController.batchValidate);
         apiRouter.post(
             '/cases/batchUpsert',
-            setBatchUpsertRevisionMetadata,
+            batchUpsertDropUnchangedCases,
```
axmb, line 97: Putting this in a separate middleware was a great call -- super clean.

axmb (commented):
Nicely done. Could you benchmark the speed of batchUpsert by running a few time trials?

I typically run three batchUpserts (via bulk upload), beginning with a clear database, with a 10k-line CSV. I'll send over a CSV. The first will be a pure insert, and the second two will be longer/identical (no op, now) upserts. I do this on both `main` and my feature branch to compare timings.

Optionally, you could also test for OOM issues, if you wanted to increase the 10k-line CSV to ~40k/50k/100k.

ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
```diff
@@ -74,7 +74,7 @@ def prepare_cases(cases, upload_id):
     TODO: Migrate source_id/source_url to this method.
     """
     for case in cases:
-        case["caseReference"]["uploadId"] = upload_id
+        case["caseReference"]["uploadIds"] = [upload_id]
```
attwad, line 77: the argument to this method should be `upload_ids` and be the array of existing uploads no?

data-serving/data-service/src/controllers/preprocessor.ts
allysonjp715: It was used in future middleware to drop unchanged cases before calling `Case.insertMany`. Removed both.

ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
allysonjp715: How I've configured it is that the middleware will take any upload_ids from the request and add it to the existing upload_ids (removing duplicates). That way a user could specify all existing upload ids in the request, or just the one new upload id that needs to be added to the existing list. This means the requester does not have to look up existing upload_ids in order to add their new uploadId to the list. 

This is an example of just supplying the new upload_id.

data-serving/data-service/src/controllers/preprocessor.ts
allysonjp715: Done.

data-serving/data-service/src/controllers/preprocessor.ts
allysonjp715: Done.

data-serving/data-service/schemas/cases.indexes.json
allysonjp715: Looks like this can be done by checking in the first element in the array exists, done.

allysonjp715:
> Nicely done. Could you benchmark the speed of batchUpsert by running a few time trials?
> 
> I typically run three batchUpserts (via bulk upload), beginning with a clear database, with a 10k-line CSV. I'll send over a CSV. The first will be a pure insert, and the second two will be longer/identical (no op, now) upserts. I do this on both `main` and my feature branch to compare timings.
> 
> Optionally, you could also test for OOM issues, if you wanted to increase the 10k-line CSV to ~40k/50k/100k.

I'm getting a lot of `socket hang up` errors when doing batchUploads, so I wasn't able to complete the whole trial. I ran it on my feature branch and got:
1st trial: 47s
2nd trial: 57s
3rd trial: 53s

On main branch I got:
1st trial 2.1 mins

All those numbers are quite unexpected to me... I wouldn't have thought on the feature branch that the no-op batch upserts of the 2nd and 3rd trial would be longer (though roughly equal) to the 1st trial. I'm also surprised the feature branch was half the time of the main branch on the 1st trial since those are all inserts. I'm thinking this might not be a fair comparison... perhaps my computer had less resources available for different runs. But I'm struggling to run them again with all these `socket hang up` errors. We can chat on slack once your in .

ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
attwad: oh so even if you just add a new one it has to be a list, I guess that's what tripped me. thanks for the clarification.

attwad (approved):

axmb (approved):

PR #931, opened 2020-09-03
hide source and upload tables from readers

### LINKED ISSUE #930: Reader Access view showing source and upload tabs
- If a profile is reader access they should only be able to access the tabs: Home / Line list / Profile

- Currently they can also see sources and uploads tabs which is not correct.

- Can we remove access to these two tabs for reader profiles

attwad:
Fixes #930

### DIFF
--- a/verification/curator-service/ui/cypress/integration/components/AppTest.spec.ts
+++ b/verification/curator-service/ui/cypress/integration/components/AppTest.spec.ts
@@ -43,6 +43,7 @@ describe('App', function () {
         cy.contains('Linelist').should('not.exist');
         cy.contains('Sources').should('not.exist');
         cy.contains('Profile').should('not.exist');
+        cy.contains('Uploads').should('not.exist');
         cy.contains('Manage users').should('not.exist');
     });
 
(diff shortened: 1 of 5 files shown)


verification/curator-service/ui/cypress/integration/components/AppTest.spec.ts
```diff
@@ -89,8 +89,8 @@ describe('App', function () {
         cy.contains('Create new').should('not.exist');
         cy.contains('Home');
         cy.contains('Linelist');
-        cy.contains('Sources');
         cy.contains('Profile');
+        cy.contains('Sources').should('not.exist');
```
allysonjp715, line 93: Could you add a check for the uploads tab in all these "Homepage with logged in *" tests?

allysonjp715 (approved):

verification/curator-service/ui/cypress/integration/components/AppTest.spec.ts
attwad: sure, done.

PR #946, opened 2020-09-04
Don't send None values as null in server requests.

axmb:
I'd previously created a pattern of declaring a field ~unset in parsers by setting its value to `None`. This works for vanilla `requests`, but if you use the autoencoding (by supplying POST bodies with the `json=` param), `None` values actually get converted to null.

Adding a quick utility function (that doesn't take meaningful time -- about 20ms for 2k cases) to strip `None` values prior to json encoding/sending to the server.


attwad (approved):
Good catch

ingestion/functions/parsing/common/python/parsing_lib/parsing_lib.py
```diff
@@ -69,11 +69,20 @@ def prepare_cases(cases, upload_id):
...
+    for i in range(len(cases)):
+        cases[i]["caseReference"]["uploadId"] = upload_id
+        cases[i] = remove_nested_nones(cases[i])
     return cases
 
 
+def remove_nested_nones(d):
+    if not isinstance(d, (dict, list)):
+        return d
+    if isinstance(d, list):
```
attwad, line 81: after reviewing a few other parsers from Anya and Felix I think it'd be nice to also remove empty strings, can you add a check in here as well for that please?

axmb: Sure, can do.

PR #947, opened 2020-09-04
Brazil - Rio Grande do Sul Parser

calremmel:
Here's a first take on this parser. A couple of thoughts:

1. I'm not totally sure which preexisting conditions should go in comorbidities and which in notes. Is it just that anything that doesn't have an exact match here goes in notes? https://disease-ontology.org/
2. I think I saw somewhere that everything is going to be converted to UTF-8 upstream. Is that right? If not, I'll need to open the file differently.

Thanks all!


attwad:
> Here's a first take on this parser. A couple of thoughts:
> 
> 1. I'm not totally sure which preexisting conditions should go in comorbidities and which in notes. Is it just that anything that doesn't have an exact match here goes in notes? https://disease-ontology.org/
> 2. I think I saw somewhere that everything is going to be converted to UTF-8 upstream. Is that right? If not, I'll need to open the file differently.
> 
> Thanks all!

Yes every sources ingested is converted to utf-8 by the retrieval function so you can assume everything's utf-8.

.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "python.pythonPath": "/home/calre/.cache/pypoetry/virtualenvs/functions-IfbyLCQu-py3.8/bin/python"
```
attwad, line 2: revert please?

ingestion/functions/parsing/riograndedosul/riograndedosul.py
```diff
@@ -0,0 +1,246 @@
...
+    if other == "SIM":
+        symptoms.append("Other")
+    return symptoms
+
+
+def convert_preexisting_conditions(raw_commorbidities: str):
+    preexistingConditions = {}
+    if raw_commorbidities:
+        preexistingConditions["hasPreexistingConditions"] = True
+        commorbidities = {
```
attwad, line 113: please declare that map outside of the method at the top of the file so that it can be reused accross all cases and not instantiated every time.

ingestion/functions/parsing/riograndedosul/riograndedosul.py
```diff
@@ -0,0 +1,246 @@
...
+    if shortbreath == "SIM":
+        symptoms.append("Shortness of Breath")
+    if other == "SIM":
+        symptoms.append("Other")
+    return symptoms
+
+
+def convert_preexisting_conditions(raw_commorbidities: str):
+    preexistingConditions = {}
+    if raw_commorbidities:
```
attwad, line 111: up to you but I prefer to keep indented left as much as possible, that means checking for early returns first:

if not raw_commorbidities:
  return None
...

ingestion/functions/parsing/riograndedosul/riograndedosul.py
```diff
@@ -0,0 +1,246 @@
...
+
+def convert_demographics(gender: str, age: str, ethnicity: str):
+    demo = {}
+    if gender:
+        demo["gender"] = convert_gender(gender)
+    if age:
+        if age == "80 e mais":
+            demo["ageRange"] = {
+                "start": 80,
+                "end": 80
```
attwad, line 164: 80 e mais means 80 or more right? in that case end should be set to some kind of max value like 120 for example.

ingestion/functions/parsing/riograndedosul/riograndedosul.py
```diff
@@ -0,0 +1,246 @@
...
+        raw_notes.append("Patient with immunosupression")
+    if "Imunodeficiênci" in raw_commorbidities:
+        raw_notes.append("Patient with immunodeficiency")
+    if raw_notes_neighbourhood:
+        raw_notes.append("Neighbourhood: " + raw_notes_neighbourhood.title())
+    if raw_notes_indigenousEthnicity != "NAO ENCONTRADO":
+        raw_notes.append("Indigenous ethnicity: " +
+                         raw_notes_indigenousEthnicity)
+    notes = (', ').join(raw_notes)
+    if notes != '':
```
attwad, line 197: just `return notes` is fine (empty string will be ignored)

ingestion/functions/parsing/riograndedosul/riograndedosul.py
```diff
@@ -0,0 +1,246 @@
...
+                if row[_COMORBIDITIES]:
+                    case["preexistingConditions"] = convert_preexisting_conditions(
+                        row[_COMORBIDITIES])
+                notes = convert_notes(
+                    row[_COMORBIDITIES], row[_NEIGHBORHOOD], row[_INDIGENOUS])
+                if notes:
+                    case["notes"] = notes
+                cases.append(case)
+            except ValueError as ve:
+                print(ve)
```
attwad, line 241: (might be useful to print the row as well for debugging)

ingestion/functions/template.yaml
```diff
@@ -149,6 +149,17 @@ Resources:
...
+    Type: AWS::Serverless::Function # More info about Function Resource: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#awsserverlessfunction
+    Properties:
+      CodeUri: parsing/riograndedosul/
+      Handler: riograndedosul.lambda_handler
+      Description: Parse case data for the Rio Grande do Sul State of Brazil
+      # Function's execution role
+      Policies:
+        - AWSLambdaBasicExecutionRole
+        - AWSLambdaReadOnlyAccess
+      MemorySize: 512
```
attwad, line 162: If you're parsing CSV line by line I'm surprized this is needed, this was needed for Japan when it's one json blob that we can't stream-parse but it your case it shouldn't be needed and will just increase the costs of running the function on aws.
On a local run, what was the max memory usage printed at the end? (sam local invoke prints that info)

attwad (approved):

ingestion/functions/parsing/riograndedosul/riograndedosul.py
```diff
@@ -0,0 +1,242 @@
...
+                "start": 0,
+                "end": 0
+            }
+        else:
+            age_range = age.split(' a ')
+            demo["ageRange"] = {
+                "start": float(age_range[0]),
+                "end": float(age_range[1])
+            }
+        if ethnicity != "NAO INFORMADO":
```
axmb, line 177: Seems this should be de-indented.

ingestion/functions/parsing/riograndedosul/riograndedosul.py
```diff
@@ -0,0 +1,242 @@
...
+            demo["ethnicity"] = convert_ethnicity(ethnicity.title())
+    return demo
```
axmb, line 179: As nit'd previously, it'd be better to return `None` if we have none of gender/age/ethnicity.

ingestion/functions/parsing/riograndedosul/riograndedosul.py
```diff
@@ -0,0 +1,242 @@
...
+        raw_notes.append("Patient with immunosupression")
+    if "Imunodeficiênci" in raw_commorbidities:
+        raw_notes.append("Patient with immunodeficiency")
+    if raw_notes_neighbourhood:
+        raw_notes.append("Neighbourhood: " + raw_notes_neighbourhood.title())
+    if raw_notes_indigenousEthnicity != "NAO ENCONTRADO":
+        raw_notes.append("Indigenous ethnicity: " +
+                         raw_notes_indigenousEthnicity)
+    notes = (', ').join(raw_notes)
+    return notes
```
axmb, line 197: Similar re: `''` vs `None`

ingestion/functions/parsing/riograndedosul/riograndedosul.py
```diff
@@ -0,0 +1,242 @@
...
+        reader = csv.reader(f, delimiter=';')
+        next(reader)  # Skip the header.
+        for row in reader:
+            try:
+                case = {
+                    "caseReference": {
+                        "sourceId": source_id,
+                        "sourceUrl": source_url
+                    },
+                    "location": ", ".join([row[_COUNTY].title(), "Rio Grande do Sul", "Brazil"]),
```
axmb, line 214: This should be in `location.query`, as opposed to top-level `location`. Did you have an opportunity to test this locally?

axmb (changes requested):

ingestion/functions/parsing/riograndedosul/riograndedosul.py
```diff
@@ -0,0 +1,245 @@
...
+        reader = csv.reader(f, delimiter=';')
+        next(reader)  # Skip the header.
+        for row in reader:
+            try:
+                case = {
+                    "caseReference": {
+                        "sourceId": source_id,
+                        "sourceUrl": source_url
+                    },
+                    "location": {"query": ", ".join([row[_COUNTY].title(), "Rio Grande do Sul", "Brazil"])},
```
attwad, line 218: nit, I don't think mapbox cares about `title()`

ingestion/functions/parsing/riograndedosul/riograndedosul_test.py
```diff
@@ -0,0 +1,48 @@
...
+        'location': {
+            'query': 'Água Santa, Rio Grande do Sul, Brazil'
+            },
+        'events': [
+            {
+                'name': 'confirmed',
+                'dateRange': {'start': '08/03/2020Z', 'end': '08/03/2020Z'}
+            },
+            {
+                'name': 'onsetSymptoms',
```
attwad, line 29: perhaps tune the test data to cover more cases (hospotalization/outcome)

calremmel: I am so glad you asked me to do this, uncovered some mismatched indices that I hadn't caught otherwise

ingestion/functions/parsing/riograndedosul/riograndedosul.py
```diff
@@ -0,0 +1,246 @@
...
+_COMORBIDITIES = 17
+_INDIGENOUS = 21
+_NEIGHBORHOOD = 24
+
+_COMORBIDITIES_MAP = {
+    'Doenças renais crônicas em estágio avançado (graus 3, 4 ou 5)': 'chronic kidney disease',
+    'Asm': 'asthma',
+    'Doença Cardiovascular Crônica': 'cardiovascular disease',
+    'Doenças cardíacas crônicas': 'heart disease',
+    'Doença Hematológica Crônic': "hematopoietic system disease",
```
AnyaLindstromBattle, line 40: Following a conversation with Moritz (see Paraiba PR - now closed), Doença Hematológica Crônica directly translates to hematologic disease, which is not in the DO database. Hematopoietic system disease is actually subtly different. As a result I would put this in notes, if it is present.

ingestion/functions/parsing/riograndedosul/riograndedosul.py
```diff
@@ -0,0 +1,246 @@
...
+            demo["ageRange"] = {
+                "start": float(age_range[0]),
+                "end": float(age_range[1])
+            }
+    if ethnicity != "NAO INFORMADO":
+        demo["ethnicity"] = convert_ethnicity(ethnicity.title())
+    return demo
+
+
+def convert_notes(raw_commorbidities: str, raw_notes_neighbourhood: str, raw_notes_indigenousEthnicity: str):
```
AnyaLindstromBattle, line 185: Again to be consistent with the Paraiba parser, if Others is present as a comorbidity, add 'Unspecified pre-existing condition' to notes.

AnyaLindstromBattle (commented):
Nice work!

ingestion/functions/parsing/riograndedosul/riograndedosul_test.py
```diff
@@ -0,0 +1,50 @@
...
+                 'dateRange': {'start': '06/02/2020Z', 'end': '06/02/2020Z'}},
+                {'name': 'hospitalAdmission',
+                 'dateRange': {'start': '06/18/2020Z', 'end': '06/18/2020Z'},
+                 'value': 'Yes'},
+                {'name': 'outcome',
+                 'dateRange': {'start': '07/08/2020Z', 'end': '07/08/2020Z'},
+                 'value': 'Death'}],
+        'demographics': {'gender': 'Male',
+                         'ageRange': {'start': 40.0, 'end': 49.0},
+                         'ethnicity': None},
```
AnyaLindstromBattle, line 29: To me it looks as though, in sample_data, the ethnicity ('RACA_COR') is 'BRANCA' and that the information about indigenous ethnicity ('ETNIA_INDIGENA') is 'NAO ENCONTRADO'. I think you may need to check your indexing.

ingestion/functions/parsing/riograndedosul/riograndedosul.py
```diff
@@ -0,0 +1,246 @@
...
+                        "sourceId": source_id,
+                        "sourceUrl": source_url
+                    },
+                    "location": {"query": ", ".join([row[_COUNTY], "Rio Grande do Sul", "Brazil"])},
+                    "events": convert_events(
+                        row[_DATE_CONFIRMED],
+                        row[_DATE_SYMPTOMS],
+                        row[_HOSPITALIZED],
+                        row[_DATE_DEATH]),
+                    "demographics": convert_demographics(
```
AnyaLindstromBattle, line 225: You may want to include the information included in the 'PROFISSIONAL_SAUDE' entry, namely whether the patient is a Healthcare worker, in 'occupation'.

ingestion/functions/parsing/riograndedosul/riograndedosul_test.py
calremmel: Great catch! Thanks for this! In future parsers, I think I'm going to switch over to csv.DictReader to avoid indexing issues. It's a pretty easy error for me to make, and that just removes it from the equation.

AnyaLindstromBattle: Agreed, just made that change for Paraiba, will switch over for new parsers too

Mougk (approved):
All seems good. thanks @calremmel

PR #951, opened 2020-09-07
Limit num of deletable cases to 10K for curators with no limit for admins

### LINKED ISSUE #937: Preventing accidental large deletes of line list entries: Curator limit to line list deletions
**Feature request**
- Ability to prevent accidental large line list case deletion by limiting the number of line list cases that can be deleted by a curator in one attempt

*Ack this means that a curator can use the work around to delete small portions below the max limit as a workaround to not require the curator admin.

**Solution**
- Only allow curators to delete a maximum number of cases to prevent accidental deletion of the full line list.
- Any deletes above that number would require the curator admin to make the deletion.
- If a delete of total numbers was above this amount the delete function would fail and error message presented to the user informing them that they need to contact the admin curator them to perform the delete
- This would require the curator being able to pass on details of the search query that needs to be performed to return back the case set and then perform the delete.

Proposal to set the curator max one time deletion amount at 10,000 cases.

attwad:
Fixes #937

This also makes the linelist visible to admins so that they can issue the delete call from there.
I did not give them more permissions other than delete (for example edit/verify is not granted).

### DIFF
--- a/verification/curator-service/ui/src/components/LinelistTable.tsx
+++ b/verification/curator-service/ui/src/components/LinelistTable.tsx
@@ -579,7 +579,8 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
                 <MaterialTable
                     tableRef={this.tableRef}
                     columns={[
-                        ...(this.props.user.roles.includes('curator')
+                        ...(this.props.user.roles.includes('curator') ||
+                        this.props.user.roles.includes('admin')
                             ? [
                                   // TODO: move to the left of selection checkboxes when possible
                                   // https://github.com/mbrn/material-table/issues/2317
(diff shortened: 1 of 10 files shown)


verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -579,7 +579,8 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
                 <MaterialTable
                     tableRef={this.tableRef}
                     columns={[
-                        ...(this.props.user.roles.includes('curator')
+                        ...(this.props.user.roles.includes('curator') ||
```
allysonjp715, line 582: This would also give admins the option to edit a row (which would cause an error if they clicked on it). It's probably best to not give admins visibility to this three-dot menu. They can still delete a single row by clicking on the selection box and clicking delete in the toolbar.

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -813,7 +816,8 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
                         }
                     }}
                     actions={
-                        this.props.user.roles.includes('curator')
+                        this.props.user.roles.includes('curator') ||
```
allysonjp715, line 819: This would also give admins the ability to verify/unverify cases (which would cause an error).

allysonjp715 (approved):

verification/curator-service/ui/src/components/LinelistTable.tsx
attwad: ok reverted

verification/curator-service/ui/src/components/LinelistTable.tsx
attwad: I used the hidden attribute to hide those now.

PR #957, opened 2020-09-08
catch axios and express vlaidator errors properly

### LINKED ISSUE #954: Catch error.response.data in UI to display nicer error messages
As it was done in https://github.com/globaldothealth/list/pull/950

attwad:
As I did in #950 previously but even simpler this time using ||.

Example of error before:
image

(styling is weird btw for large errors)

Example of same error now:

image

Fixes #954

### DIFF
--- a/verification/curator-service/ui/src/components/App.tsx
+++ b/verification/curator-service/ui/src/components/App.tsx
@@ -308,7 +308,6 @@ export default function App(): JSX.Element {
             })
             .catch((e) => {
                 setUser({ _id: '', name: '', email: '', roles: [] });
-                console.error(e);
             });
     };
 
--- a/data-serving/data-service/src/controllers/case.ts
+++ b/data-serving/data-service/src/controllers/case.ts
@@ -112,7 +112,7 @@ export const list = async (req: Request, res: Response): Promise<void> => {
         res.json({ cases: docs, total: total });
     } catch (e) {
         if (e instanceof ParsingError) {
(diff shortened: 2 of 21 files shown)


verification/curator-service/ui/src/components/App.tsx
```diff
@@ -308,7 +308,6 @@ export default function App(): JSX.Element {
             })
             .catch((e) => {
                 setUser({ _id: '', name: '', email: '', roles: [] });
-                console.error(e);
```
allysonjp715, line 311: Why remove?

allysonjp715 (approved):
Thank you!!

verification/curator-service/ui/src/components/App.tsx
attwad: This is going to print a console error for all logged out users, not super useful imo

image

attwad:
Okay it turns out express validator errors have to be caught via a middleware for proper formatting (it's in their docs, don't know why we didn't follow those). So I did that.
And I made ALL our errors match the same principle that is the json output of errors should contain a message field otherwise the stringified version of the error output is used (catch-all in case the server doesn't follow that best practice).

message field is part of all errors in javascript so it actually makes the code simpler in a few places.

Now we get nice errors on 404s for example:
Before:
image

Now:
Screenshot 2020-09-08 at 11 21 28

allysonjp715 (approved):

Issue #909, opened 2020-09-01
No ability to download selected line list data


sratcliffe118:
No ability to download selected line list data from the line list tab

@shroothless do you have the UX design for the download functionality?

shroothless:
![image](https://user-images.githubusercontent.com/67452108/91873714-5ddfc280-ec47-11ea-98b3-71b4c4df4bbd.png)
 Is this what you are referring to?

sratcliffe118:
Yes that is correct. My current UI doesn't shows this in PROD or DEV
![Screen Shot 2020-09-01 at 5 06 44 PM](https://user-images.githubusercontent.com/63205311/91877211-a01ff880-ec75-11ea-9690-79945a7e92ec.png)

shroothless:
UX was delivered recently so likely in progress.

sratcliffe118:
@allysonjp715 is the download functionality still in development?

PR #960, opened 2020-09-08
Add a button for downloading all case data

allysonjp715:
#909

Screen Shot 2020-09-08 at 3 05 08 PM


data-serving/data-service/src/controllers/case.ts
```diff
@@ -18,6 +21,35 @@ export const get = async (req: Request, res: Response): Promise<void> => {
...
+ * Handles HTTP GET /api/cases/download.
+ */
+export const download = async (req: Request, res: Response): Promise<void> => {
+    const cases = await Case.find({}).lean();
+
+    res.setHeader('Content-Type', 'text/csv');
+    res.setHeader('Content-Disposition', 'attachment; filename="cases.csv"');
+    res.setHeader('Cache-Control', 'no-cache');
+    res.setHeader('Pragma', 'no-cache');
+    fetch(
```
attwad, line 36: let's use axios for http requests as we do in the curator service, unless there's a real reason to use this new library.

data-serving/data-service/api/openapi.yaml
```diff
@@ -197,6 +197,16 @@ paths:
...
         "500":
           $ref: "#/components/responses/500"
+  /cases/download:
+    get:
+      summary: Streams a CSV attachment of all cases
+      tags: [Case]
+      operationId: downloadCases
+      responses:
+        "200":
+          $ref: "#/components/responses/200"
```
attwad, line 207: can we describe a bit more what the response body is like here?

attwad (commented):
My understanding was that this button would download all data that matches the current query (bug says "download **selected** linelist data"), if people want to download the whole data set we have the daily dumps for that. Can you amend this PR to account for that or where you planning to do that in a next one?

allysonjp715:
> My understanding was that this button would download all data that matches the current query (bug says "download **selected** linelist data"), if people want to download the whole data set we have the daily dumps for that. Can you amend this PR to account for that or where you planning to do that in a next one?

Yes planning to add the current query to the next PR.

verification/curator-service/api/src/index.ts
```diff
@@ -225,6 +225,11 @@ new OpenApiValidator({
             mustHaveAnyRole(['reader', 'curator', 'admin']),
             casesController.list,
         );
+        apiRouter.get(
+            '/cases/download',
+            mustHaveAnyRole(['reader', 'curator', 'admin']),
```
attwad, line 230: not sure if it makes sense for admins to be able to download

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -467,6 +477,12 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
         });
     }
 
+    downloadCases(): void {
+        axios
+            .get('/api/cases/download')
+            .then((response) => fileDownload(response.data, 'cases.csv'));
```
attwad, line 483: this library can't parse the "cases.csv" from the headers in the response? (it has that default name already)

attwad (commented):
I'm wondering how much we want to put in the data service vs in the curator service, I would have envisioned the curator service to just call the data service's GET /cases API to get the cases then do the CSV conversion itself, it keeps the data service unchanged this way so feels simpler, wdyt?

data-serving/data-service/src/controllers/case.ts
allysonjp715: Done

verification/curator-service/api/src/index.ts
allysonjp715: I chose it because it matched `/cases` above. Figured if they could query for all the cases then they could download all the cases. But I don't feel strongly here - I could change one or both?

attwad: ah right I added that because of the no-limit delete functionality I added recently. Fine as is then makes sense to be able to download if they can read indeed.

verification/curator-service/ui/src/components/LinelistTable.tsx
allysonjp715: No, the fileDownload library is very simple - only takes the data and the filename: https://github.com/kennethjiang/js-file-download

I could parse it from the response headers, but the header it's in is formatted like `content-disposition: attachment; filename="cases.csv"`. We'd have to parse through that string, which doesn't really add benefit IMO.

allysonjp715:
> I'm wondering how much we want to put in the data service vs in the curator service, I would have envisioned the curator service to just call the data service's GET /cases API to get the cases then do the CSV conversion itself, it keeps the data service unchanged this way so feels simpler, wdyt?

This would be a good idea if we didn't have streaming. But with streaming we can pass far more data through. Will add more thoughts in the comment about streaming.

verification/curator-service/api/openapi/openapi.yaml
```diff
@@ -492,6 +492,16 @@ paths:
           $ref: "#/components/responses/422"
         "500":
           $ref: "#/components/responses/500"
+  /cases/download:
+    get:
+      summary: Streams a CSV attachment of all cases
```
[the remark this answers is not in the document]
allysonjp715: I'm not 100% sure if this is the right way to check for streaming, but the network response is gradually incrementing the size of the response that comes back. I can't attach videos to this comment but I sent one over slack. I did a download of 50k cases with no OOM issue, which leads me to believe the streaming is working. 

The stringify library in the curator-service claims to begin the streaming process, which is then piped through (all requests and responses in express are streams).

data-serving/data-service/api/openapi.yaml
allysonjp715: Done.

verification/curator-service/api/openapi/openapi.yaml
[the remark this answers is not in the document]
attwad: lgtm, thanks for checking.

verification/curator-service/ui/src/components/LinelistTable.tsx
attwad: ok no big deal

attwad:
> > I'm wondering how much we want to put in the data service vs in the curator service, I would have envisioned the curator service to just call the data service's GET /cases API to get the cases then do the CSV conversion itself, it keeps the data service unchanged this way so feels simpler, wdyt?
> 
> This would be a good idea if we didn't have streaming. But with streaming we can pass far more data through. Will add more thoughts in the comment about streaming.

So apparently mongo has streaming built-in for queries, we're just not using it
https://mongoosejs.com/docs/queries.html#streaming

Using this now falls out of scope of this PR obviously so fine as is.

data-serving/data-service/src/controllers/case.ts
```diff
@@ -18,6 +21,35 @@ export const get = async (req: Request, res: Response): Promise<void> => {
...
+ */
+export const download = async (req: Request, res: Response): Promise<void> => {
+    const cases = await Case.find({}).lean();
+
+    res.setHeader('Content-Type', 'text/csv');
+    res.setHeader('Content-Disposition', 'attachment; filename="cases.csv"');
+    res.setHeader('Cache-Control', 'no-cache');
+    res.setHeader('Pragma', 'no-cache');
+    axios
+        .get<string>(
```
attwad, line 37: you might want to catch() here as well or wrap in try/catch the axios call otherwise the request will just hang if we can't connect to github (happens more often that what I'd like to admit ;) )

attwad (approved):

data-serving/data-service/src/controllers/case.ts
allysonjp715: Ah yes, thanks, done.

PR #962, opened 2020-09-08
Don't set HTTP timeouts.

### LINKED ISSUE #961: Fix 60-second request timeout / socket hang up in dev/prod
**Describe the bug**
Requests in dev/prod time out after 60 seconds.

**To Reproduce**
Send an API request, say to batch upsert, with a large amount of data (>10k rows).

**Expected behavior**
Requests, at least batch upsert, should allow more time to complete.

**Environment (please complete the following information):**
Only occurs in dev/prod -- can't repro locally. I've seen the issue both for bulk upload and ADI, so the issue isn't specific to the UI/browser.

I've previously tweaked our nginx config to avoid 504s, but in these cases, the API client gets `500: socket hang up`.

axmb:
This is a little heavy-handed, but I can dial it back to the right level if this works.

Trying to debug #961 (which I can't repro locally), so casting a wide net at first.

### DIFF
--- a/data-serving/data-service/src/server.ts
+++ b/data-serving/data-service/src/server.ts
@@ -6,4 +6,7 @@ const server = app.listen(app.get('port'), async () => {
     console.log('  Press CTRL-C to stop\n');
 });
 
+// Set global 1-hour timeout.
+server.setTimeout(60 * 60 * 1000);
+
 export default server;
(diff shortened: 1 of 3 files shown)


attwad (approved):
Might be easier to revert later on if you leave a few todos with the #962 trace, up to you.

axmb:
> Might be easier to revert later on if you leave a few todos with the #962 trace, up to you.

Yep, good call.

PR #964, opened 2020-09-09
Send parsed cases in batches to avoid OOMs and ease lambda memory usage

attwad:
For #961 

Overall feels much efficient than storing the whole list in memory in multiple places.

### DIFF
--- a/ingestion/functions/parsing/common/python/parsing_lib/input_event.json
+++ b/ingestion/functions/parsing/common/python/parsing_lib/input_event.json
@@ -7,7 +7,7 @@
     "uploadIds": ["123456789012345678901234"],
     "dateFilter": {
         "numDaysBeforeToday": 3,
-        "op": "EQ"
+        "op": "LT"
     },
     "auth": {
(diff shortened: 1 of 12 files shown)


ingestion/functions/parsing/common/python/parsing_lib/input_event.json
```diff
@@ -7,7 +7,7 @@
     "uploadIds": ["123456789012345678901234"],
     "dateFilter": {
         "numDaysBeforeToday": 3,
-        "op": "EQ"
+        "op": "LT"
```
attwad, line 10: somehow this was needed to pass unit tests because that file is read by the e2e test, I have no idea how this passed in prod as it failed continuously locally.

axmb: I suspect there's some timezone goofiness that impacts this; I don't remember in which TZ our actions run.

axmb (approved):
Nice change!

PR #965, opened 2020-09-09
If there is a search query only download search results

allysonjp715:
#909

### DIFF
--- a/data-serving/data-service/src/controllers/case.ts
+++ b/data-serving/data-service/src/controllers/case.ts
@@ -32,10 +32,10 @@ export const download = async (req: Request, res: Response): Promise<void> => {
     let cases: any;
     try {
         if (req.query.q) {
-            cases = await (casesMatchingSearchQuery({
+            cases = await casesMatchingSearchQuery({
                 searchQuery: req.query.q as string,
                 count: false,
-            }) as DocumentQuery<CaseDocument[], CaseDocument, unknown>).lean();
+            });
         } else {
             cases = await Case.find({}).lean();
         }
@@ -106,7 +106,7 @@ export const casesMatchingSearchQuery = (opts: {
             countQuery.where(f.path).in(f.values);
         }
     });
-    return opts.count ? countQuery : casesQuery;
+    return opts.count ? countQuery : casesQuery.lean();
 };
 
 /**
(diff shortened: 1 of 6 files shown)


data-serving/data-service/src/controllers/case.ts
```diff
@@ -29,28 +29,48 @@ export const get = async (req: Request, res: Response): Promise<void> => {
...
  */
 export const download = async (req: Request, res: Response): Promise<void> => {
-    const cases = await Case.find({}).lean();
+    let cases: any;
+    try {
+        if (req.query.q) {
+            cases = await (casesMatchingSearchQuery({
+                searchQuery: req.query.q as string,
+                count: false,
+            }) as DocumentQuery<CaseDocument[], CaseDocument, unknown>).lean();
```
attwad, line 38: wondering if we want to move the lean() inside caseMatchingSearchQuery so that callers don't have to care about it?

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -482,9 +483,20 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
...
+        let downloadUrl = '/api/cases/download';
+        const trimmedQ = this.state.search.trim();
+        if (trimmedQ) {
+            downloadUrl += '?q=' + encodeURIComponent(trimmedQ);
+        }
+        this.setState({ error: '' });
         axios
-            .get('/api/cases/download')
-            .then((response) => fileDownload(response.data, 'cases.csv'));
+            .get(downloadUrl)
```
attwad, line 493: I would expect the get() method from axios to do the URI encoding no? do we really have to do it ourselves?

attwad (approved):
next PR is for selected rows only I presume? :)

verification/curator-service/ui/src/components/LinelistTable.tsx
allysonjp715: Yes we have to do it ourselves :( throws an error if we don't. We had to do it in the request for table data as well.

allysonjp715:
> next PR is for selected rows only I presume? :)

Indeed :)

data-serving/data-service/src/controllers/case.ts
allysonjp715: Done.

PR #966, opened 2020-09-09
increase memory of japan parsing func

attwad:
Note: the json input file is 30Mb (as of today, still growing of course), there is no way to stream this in unfortuantely as the object that it contains is an array of cases :(

They do not have a csv as an alternative format.

When ran locally the function was just timing out on json.load(f) with no logs.
After increasing the memory to 512Mb it worked.

https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/sam-resource-function.html#sam-function-memorysize

#961

### DIFF
--- a/ingestion/functions/template.yaml
+++ b/ingestion/functions/template.yaml
@@ -92,6 +92,7 @@ Resources:
         - AWSLambdaBasicExecutionRole
         - AWSLambdaReadOnlyAccess
       Timeout: 300
+      MemorySize: 512
       Layers:
         - !Ref ParsingLibLayer
         - !Ref CommonLibLayer


axmb (approved):
> After increasing the memory to 512Mb it worked.

I've considered playing around with this -- does `sam local` respect the `MemorySize` for a function? IOW, were you timing out locally or in AWS?

attwad:
> > After increasing the memory to 512Mb it worked.
> 
> I've considered playing around with this -- does `sam local` respect the `MemorySize` for a function? IOW, were you timing out locally or in AWS?

yes local respects that limit, the other PR has some info on how I debugged that (checked env var in the local docker container).

PR #968, opened 2020-09-09
Move common parsing template values into Globals.

axmb:
A timeout of 900 is the "default" (in that Lambdas are terminated after 900 seconds). It's probably worthwhile to declare it, so that our behavior doesn't implicitly change if AWS defaults are tweaked in the future.


attwad (approved):
nice

PR #971, opened 2020-09-10
Add a basic client for sending emails via gmail SMTP

axmb:
Basic idea is that we'll register addressees per-source, and send emails on relevant updates in our sources controller.

Using GMail's public SMTP server, because YAGNI. We can still add custom domains, IIUC, and if someone's feeling particularly passionate later on we can run/use our own SMTP server instead.

Re: #978


verification/curator-service/api/src/clients/email-client.ts
```diff
@@ -0,0 +1,89 @@
...
+
+        return new EmailClient(service, transport);
+    }
+
+    /**
+     * Sends an email via the configured transport.
+     *
+     * @param addressees - Addressees to whom the email is sent.
+     * @param subject - Subject of the email.
+     * @param text - Body/content of the email.
```
attwad, line 68: can this be html? or just raw text?

attwad (approved):

attwad:
Reference a github issue perhaps?

verification/curator-service/api/src/clients/email-client.ts
axmb: Text has to be plaintext. There's an `html` option, too, which I'm hoping to quickly poke UX for. :point_right:

axmb:
> Reference a github issue perhaps?

Can do!

Issue #502, opened 2020-07-09
Import Estonia case data


iamleeg:
https://opendata.digilugu.ee/opendata_covid19_test_results.csv

attwad:
Description of dataset here:
https://www.terviseamet.ee/et/koroonaviirus/avaandmed

attwad:
Needs to be setup in the UI now.

z023:
The Estonia source lists all people tested so includes both those who were positive and negative so would need to be able to identify those who were positive and negative??
Also the Estonia CSV file when opened is changing some of the age ranges into dates, e.g. age 5-9 is coming through as 5-Sept. This might just be excel when opening and there isn't actually a problem with the file.

attwad:
Yes only positive cases are imported. And ages should be parsed correctly as well.

Mougk:
Ok, will add feature request to add negatives but that can wait until October.

attwad:
Why? The list has always been about only positive cases, not a list of all test data. Did something change about that recently?

attwad:
FTR we have no way of storing negative tests results in the schema, only positive cases.

allysonjp715:
We have an issue about non-positive cases already: #457

Mougk:
Thanks @allysonjp715 @attwad !

z023:
Estonia is set up but isn't being ingested.

attwad:
The parser is a CSV parser but it was setup with a JSON source, I've changed the URL to be https://opendata.digilugu.ee/opendata_covid19_test_results.csv, should be ingested in next run, feel free to trigger it yourself if you want before that.

PR #974, opened 2020-09-10
Add estonia parser

attwad:
For #502 

```
Creating upload via http://localhost:3001/api/sources/5f59e8c47ef66004bc9c91ce/uploads
Retrieving raw data from s3://epid-sources-raw/5f59e8c47ef66004bc9c91ce/2020/09/10/0855/content.csv
Raw file retrieved at /tmp/tmpo1k99tqa
Sending 1000 cases to http://localhost:3001/api/cases/batchUpsert
Sending 1000 cases to http://localhost:3001/api/cases/batchUpsert
Sending 600 cases to http://localhost:3001/api/cases/batchUpsert
sent 2600 cases
Updating upload via http://localhost:3001/api/sources/5f59e8c47ef66004bc9c91ce/uploads/5f59eade7ef66004bc9c91d9
END RequestId: d5482f91-a2ea-1c4b-f551-f093a28c3dca
REPORT RequestId: d5482f91-a2ea-1c4b-f551-f093a28c3dca	Init Duration: 713.60 ms	Duration: 11683.83 ms	Billed Duration: 11700 ms	Memory Size: 512 MB	Max Memory Used: 60 MB

{"count_created":2600,"count_updated":0}
```

seems correct: 
image


ingestion/functions/parsing/estonia/estonia.py
```diff
@@ -0,0 +1,108 @@
...
+try:
+    import parsing_lib
+except ImportError:
+    sys.path.append(
+        os.path.join(
+            os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
+            'common/python'))
+    import parsing_lib
+
+# Fixed location, all cases are for Estonia and no coarser location is available in the source.
```
axmb, line 18: nit: s/coarser/finer?

ingestion/functions/parsing/estonia/estonia.py
```diff
@@ -0,0 +1,108 @@
...
+                "start": float(start),
+                "end": float(end),
+            }
+        else:
+            demo["ageRange"] = {
+                "start": 85.0,
+                # Arbitrary max age for 85+
+                "end": 120.0,
+            }
+    return demo
```
axmb, line 71: Slightly better to return `None` here rather than `{}`.

ingestion/functions/parsing/estonia/estonia.py
```diff
@@ -0,0 +1,108 @@
...
+        if '-' in age:
+            start, _, end = age.partition("-")
+            demo["ageRange"] = {
+                "start": float(start),
+                "end": float(end),
+            }
+        else:
+            demo["ageRange"] = {
+                "start": 85.0,
+                # Arbitrary max age for 85+
```
axmb, line 68: Do we need to apply a max, here? Doesn't [85,] connote this?

ingestion/functions/parsing/estonia/estonia.py
```diff
@@ -0,0 +1,108 @@
...
+    else:
+        raise ValueError(f'Unhandled gender: {raw_gender}')
+
+
+def convert_demographics(gender: str, age: str):
+    demo = {}
+    if gender:
+        demo["gender"] = convert_gender(gender)
+    if age:
+        # Brackets are like "35-39" or "over 85".
```
axmb, line 58: Might be better to be explicit about these two options, rather than using the `else` statement here. For instance, if they decide to change format or add something like "under 1".

ingestion/functions/parsing/estonia/estonia.py
```diff
@@ -0,0 +1,108 @@
...
+                        "dateRange":
+                        {
+                            "start": convert_date(row['ResultTime']),
+                            "end": convert_date(row['ResultTime']),
+                        },
+                    },
+                ],
+                "demographics": convert_demographics(
+                    row['Gender'], row['AgeGroup']),
+                "notes": f"Case residence: {row['Country']} {row['County']}".strip(),
```
axmb, line 102: Would it be better to encode `County` in location, or is this explicitly the county of residence, and we can't assume it's also the county in which the case was confirmed (if there's a distinction there)?

axmb (approved):
Nice! Want to test this in dev, given that we've observed some differences between our local server stack and that in dev?

ingestion/functions/parsing/estonia/estonia.py
attwad: I always confuse those... thanks

ingestion/functions/parsing/estonia/estonia.py
attwad: right, now raising a value error if we can't parse it.

ingestion/functions/parsing/estonia/estonia.py
attwad: that's just residence, there is no data about where the infection took place unfortunately, that's why I put it in the notes.

PR #984, opened 2020-09-11
Testability changes to EmailClient.

axmb:
I liked the static factory pattern better, but it was a nightmare to test with Jest, given our relatively high fidelity app/controller tests. Because we don't await app async set up in our superagent controller tests, I needed to mock the `email-client` module (reaching the test SMTP server was too slow, and our unit tests ran before the source controller finished installation). And mocking an ES6 class (via the module mocker, because we can't touch the actual client prototype in the unit test) with a static factory method proved difficult.

### DIFF
--- a/verification/curator-service/api/src/clients/email-client.ts
+++ b/verification/curator-service/api/src/clients/email-client.ts
@@ -16,6 +16,7 @@ export enum EmailService {
 export default class EmailClient {
     private service?: EmailService;
     private transport?: Mail;
+    private initialized = false;
     constructor(private user?: string, private password?: string) {}
 
     /**
@@ -25,6 +26,10 @@ export default class EmailClient {
      * connection. If either is absent, a testing service (Ethereal) is used.
      */
     async initialize(): Promise<EmailClient> {
+        if (this.initialized) {
+            return this;
+        }
+
         if (!this.user || !this.password) {
             this.service = EmailService.Ethereal;
             const testAccount = await nodemailer.createTestAccount();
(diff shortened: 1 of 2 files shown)


verification/curator-service/api/src/clients/email-client.ts
```diff
@@ -14,30 +14,21 @@ export enum EmailService {
...
      */
-    static async create(
-        user?: string,
-        password?: string,
-    ): Promise<EmailClient> {
-        let service: EmailService;
-        let transport: Mail;
-        if (!user || !password) {
-            service = EmailService.Ethereal;
+    async initialize(): Promise<EmailClient> {
```
attwad, line 27: do you want to store the fact that it was initialized and just return this if this gets called another time instead of creating new stuff?

verification/curator-service/api/src/clients/email-client.ts
```diff
@@ -67,23 +58,26 @@ export default class EmailClient {
...
      */
-    send = (
+    send(
         addressees: string[],
         subject: string,
         text: string,
-    ): Promise<SentMessageInfo> => {
+    ): Promise<SentMessageInfo> {
+        if (this.transport === undefined) {
+            throw new Error('Client must be initialized prior to use.');
```
attwad, line 67: nit but best practice to explicitely say what's needed (did you forget to call initialize()?)

attwad (approved):

verification/curator-service/api/src/clients/email-client.ts
axmb: SG, done.

PR #986, opened 2020-09-11
fix all flake8 issues

attwad:
Basically ran autopep8 and `python3 -m flake8 **/*.py --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics` on the ingestion dir, it seems that we have a CI for that but it doens't error out when errors are encountered:

https://github.com/globaldothealth/list/runs/1100557888?check_suite_focus=true

I'd love to have an autopep8 pre-commit hook as well as we do for the typescript code, will look into that.


axmb (approved):
Excellent, thank you!

PR #989, opened 2020-09-11
Don't allow curators to delete more than 10,000 from UI

allysonjp715:
By showing an error dialog rather than a confirmation dialog if they try to delete more than 10,000 cases.

Also removes the text that "rows cannot be recovered" because they now can after #982


attwad (approved):

z023:
Typo in error message: 'cases selected to delete which is greater than the allowed maximum of 10000. An admin can **preform** the deletion if it is valid'

Issue #948, opened 2020-09-07
UX tweaks 


shroothless:
**Describe the bug**

- Tooltip icon should be in the search field 

- Move search bar up to nav bar
- Add download button
- Move pagination up
- Change type face for table rows (not headers) and pagination to Inter

**Screenshots**
https://screenshot.googleplex.com/DNCfDQAirMNvecn

See mocks here for details:
https://www.figma.com/file/TqqzDnXF4KTBAGUchpt4mQ/Global.Health-Marketing-Website?node-id=278%3A2

PR #990, opened 2020-09-11
Set font to Inter for table rows and pagination bar

allysonjp715:
#948

Screen Shot 2020-09-11 at 12 13 49 PM


attwad:
What does it fallback to if one doesn't have inter installed? is it bundled somehow by default? I don't see any fonts include directive in our html like `<link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Inter">`

allysonjp715:
> What does it fallback to if one doesn't have inter installed? is it bundled somehow by default? I don't see any fonts include directive in our html like `<link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Inter">`

Added a stylesheet for Inter. We should add one for Mabry Pro too, but I don't see it in fonts.googleapis.com. I've asked the designers - will add it in a separate PR once we have the url.

attwad (approved):
thanks!

PR #997, opened 2020-09-11
add workflow to update prod schema

attwad:
dev is handled automatically with another workflow, not touching it.

### DIFF
--- a/.github/workflows/update-db.yaml
+++ b/.github/workflows/update-db.yaml
@@ -20,18 +20,19 @@ jobs:
         with:
           node-version: "12.x"
 
-      - name: Update prod schema
+      - name: Update dev schema
         env:
-          CONN: ${{ secrets.DB_CONNECTION_URL_PROD }}
+          CONN: ${{ secrets.DB_CONNECTION_URL_DEV }}
           DB: "covid19"
           COLL: "cases"
           # Those are relative to the setup-db directory, not the working dir.
           SCHEMA: "../../data-service/schemas/cases.schema.json"
           INDEXES: "../../data-service/schemas/cases.indexes.json"
         run: npm --prefix data-serving/scripts/setup-db run setup
-      - name: Update dev schema
+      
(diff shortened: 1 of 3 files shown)


.github/workflows/update-db.yaml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: "12.x"
+
+      - name: Update dev schema
```
axmb: Believe this is done by `.github/workflows/case-data-update-dev.yml`.

.github/workflows/update-db.yaml
```diff
@@ -0,0 +1,41 @@
...
+
+jobs:
+  update-case-data:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - name: Set up Node.js
```
axmb: I remember Kelly having to install the MongoDB CLI to do this (and writing d.ts for it in the npm script). Can't recall if that was for this schema update portion, or something else, though. Might be good to test this works in the actions VM by temporarily committing a pull_request trigger.

axmb (approved):

.github/workflows/update-db.yaml
attwad: this was needed because she was running a different script that was doing a mongoimport, this script only deals with the schema and uses the node js mongo client so it should be fine (confidence lelvel 90%, merging this now ;) )

.github/workflows/update-db.yaml
attwad: I was editing this as you typed sorry ;) I noticed too late.

axmb: Ah, you've already removed!

PR #1000, opened 2020-09-11
Include Mabry Pro font and fix fonts across tables

allysonjp715:
Screen Shot 2020-09-11 at 4 53 22 PM


verification/curator-service/ui/src/components/App.tsx
```diff
@@ -95,6 +95,19 @@ const theme = createMuiTheme({
...
             },
         },
+        MuiTablePagination: {
+            root: {
+                border: 'unset',
+                fontFamily: 'Inter',
+                '& .MuiTablePagination-input': {
+                    fontFamily: 'Inter',
+                },
+                '&&& .MuiTypography-root': {
```
attwad, line 105: sure it's not &&&&&& or &&&&&&&&& ? ;-)

verification/curator-service/ui/src/components/SourceTable.tsx
```diff
@@ -535,21 +534,18 @@ class SourceTable extends React.Component<Props, SourceTableState> {
                             Pagination: (props): JSX.Element => {
                                 return (
                                     <div className={classes.tablePaginationBar}>
-                                        <Typography>
+                                        <Typography style={{ width: '100%' }}>
```
attwad, line 537: hmm shouldn't this be a className instead?

attwad (approved):
that users table though 😭

verification/curator-service/ui/src/components/SourceTable.tsx
allysonjp715: I figured it was small enough to use styles, but changed to classes for consistency.

attwad:
> that users table though 😭

ah no it is included, I thought it was missing, great! 👍

Issue #978, opened 2020-09-10
Add UI-configurable email alerts for automation


axmb:
**Describe the solution you'd like**

- Curators can specify email recipients for a given automated source.
- Email notifications are sent for source creation.
- Email notifications are sent for failed ingestion of the source.

**Describe alternatives you've considered**
We use CloudWatch alarms now, but it won't really suite the curation team. When a Lambda fails, an alarm can be sent along an SNS topic, but since we want different UI-configurable recipients for every source, we'd have a mess of SNS topics/alarms to deal with.

axmb:
@shroothless / @acansler -- any thoughts on email content for source creation or failed ingestion runs?

I'm just using a basic message now, but if you're keen to put together a couple HTML templates, I can switch to those. Not a big deal, IMO, but a nice to have.

shroothless:
What would a template look like? Can you send me a sample please..

axmb:
Yep! Sent over an email.

PR #1001, opened 2020-09-11
Set up email notifications, and fire for an initial (source creation) case

axmb:
Other notification scenarios, and UI, forthcoming.

Re: #978

### DIFF
--- a/verification/curator-service/api/src/controllers/sources.ts
+++ b/verification/curator-service/api/src/controllers/sources.ts
@@ -255,16 +255,21 @@ export default class SourcesController {
         source: SourceDocument,
         type: NotificationType,
     ): Promise<void> {
-        if (source.notificationRecipients?.length > 0) {
-            let subject: string;
-            let text: string;
-            switch (type) {
-                case NotificationType.Create:
-                    subject = 'New source configured for automation';
-                    if (source.name) {
-                        subject.concat(`: ${source.name}`);
-                    }
-                    text = `A new source was configured for automation in G.h List.
+        if (
+            !source.notificationRecipients ||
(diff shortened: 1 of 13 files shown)


verification/curator-service/api/src/controllers/sources.ts
```diff
@@ -195,6 +201,7 @@ export default class SourcesController {
                 source.toAwsStatementId(),
             );
             source.set('automation.schedule.awsRuleArn', createdRuleArn);
+            await this.sendNotifications(source, NotificationType.Create);
```
attwad, line 204: up to you but I wouldn't feel super bad if we were not waiting for the emails to be sent.

verification/curator-service/api/src/controllers/sources.ts
```diff
@@ -243,4 +250,39 @@ export default class SourcesController {
         }
         return;
     };
+
+    private async sendNotifications(
+        source: SourceDocument,
+        type: NotificationType,
+    ): Promise<void> {
+        if (source.notificationRecipients?.length > 0) {
```
attwad, line 258: early return in case length === 0 ? just to keep indent left/positiveness whenever possible.

attwad (approved):

verification/curator-service/api/src/controllers/sources.ts
axmb: Good point -- we probably don't need to wait. For now it might be nice, though, to simplify making sure they're working.

verification/curator-service/api/src/controllers/sources.ts
axmb: SG, done.

PR #1002, opened 2020-09-11
Add remainder of source-based notification emails

axmb:
Final notification scenario is for uploads that end in errors. I'll move/commonize a couple functions to do that, so I'll cut it as a separate PR.

Re: #978


verification/curator-service/api/src/controllers/sources.ts
```diff
@@ -127,7 +128,7 @@ export default class SourcesController {
         // Careful here, source.isModified('automation.schedule.awsScheduleExpression')
         // will return true even when just the parser is updated which is
         // error prone, prefer isModified() without dotted.paths if possible.
-        if (source.automation?.schedule?.isModified('awsScheduleExpression')) {
+        if (source.automation?.isModified('schedule')) {
```
axmb, line 131: This had to be changed. This is identical in practice (the only other schedule field is rule ARN -- which won't change) in the case of adding/changing automation. But, if automation is removed -- meaning the schedule is removed (since awsScheduleExpression is required) -- we'll never trigger this scenario since schedule? is falsey.

attwad: ack, thanks for the explanation

verification/curator-service/api/src/controllers/sources.ts
```diff
@@ -273,6 +277,18 @@ export default class SourcesController {
                     \tSchedule: ${source.automation.schedule.awsScheduleExpression}
                     \tParser: ${source.automation.parser?.awsLambdaArn}`;
                     break;
+                case NotificationType.Remove:
+                    subject = 'Automation removed for source';
+                    if (source.name) {
```
attwad, line 282: isn't name a required field?

attwad (approved):

verification/curator-service/api/src/controllers/sources.ts
axmb: It is indeed, good catch!

PR #1017, opened 2020-09-15
Set up NODE_DEBUG and longjohn to debug network issues

axmb:


data-serving/data-service/Dockerfile
```diff
@@ -38,6 +38,10 @@ USER node
 # Run with node env set to production.
 ENV NODE_ENV production
 
+# TODO: Remove this pending https://github.com/globaldothealth/list/issues/961.
+# Run with node debug mode set to http, net, and tls.
+ENV NODE_DEBUG http,net,tls
```
attwad, line 43: you can set that up in the data.yaml and curator.yaml files instead of hardcoding those in the dockerfile

attwad (approved):
we need an mpm package for printing stack traces </nervous laughter>

data-serving/data-service/Dockerfile
axmb: Sure -- does this LGTU?

attwad (approved):

data-serving/data-service/Dockerfile
attwad: perfect, thanks

Issue #790, opened 2020-08-13
Optimize queries


khmoran:
We are receiving MongoDB alerts like the below:

```
Query Targeting: Scanned Objects / Returned has gone above 1000covid19-map-cluster01-shard-00-02.sc7u9.mongodb.netAlert DetailsCREATED2020/08/13 16:38 GMTTYPESecondaryCLUSTERcovid19-map-cluster01CURRENT VALUE31,120.1 | Query Targeting: Scanned Objects / Returned has gone above 1000covid19-map-cluster01-shard-00-02.sc7u9.mongodb.net |   | Alert Details |   | CREATED | 2020/08/13 16:38 GMT |   | TYPE | Secondary |   | CLUSTER | covid19-map-cluster01 |   | CURRENT VALUE | 31,120.1 | OPEN
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
Query Targeting: Scanned Objects / Returned has gone above 1000covid19-map-cluster01-shard-00-02.sc7u9.mongodb.net
 
Alert Details
 
CREATED | 2020/08/13 16:38 GMT
 
TYPE | Secondary
 
CLUSTER | covid19-map-cluster01
 
CURRENT VALUE | 31,120.1
```

Adding indexes in #782 and #785 did not solve the issue.

khmoran:
It seems related to charts, ex. here's one query that took a while:

https://cloud.mongodb.com/v2/5ea89a90db26a511f1804cf8#metrics/host/04824987aea705eb892205a8b9d5f242/profiler/covid19.cases/command/1597336651604

Not sure if there are other views/indexes that would help

abhidg:
Many indices have been added since this issue was opened (#1875) which have fixed this.

PR #1021, opened 2020-09-15
do not show charts on homepage for non curators or admins

### LINKED ISSUE #1020: Disable charts on homepage for non curators
Charts are meant for curators to see the data aggregates, right now when someone external to the project goes to curator.ghdsi.org what they get is a bunch of empty white boxes.

we should just remove those if not logged in at all and not curators, pending a nicer homepage.

attwad:
Fixes #1020

Rationale:
- we should have a proper homepage instead
- doesn't show up when in incognito mode
- charts are potentially expensive until we figure out proper indexes for them (#790)

### DIFF
--- a/verification/curator-service/ui/cypress/integration/components/AppTest.spec.ts
+++ b/verification/curator-service/ui/cypress/integration/components/AppTest.spec.ts
@@ -15,11 +15,20 @@ describe('App', function () {
     });
 
     it('Shows charts on home page', function () {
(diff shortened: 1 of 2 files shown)


allysonjp715 (approved):

sratcliffe118:
For MVP I think the visualizations will meet the requirements here so I am fine with restricting this view to admins and curators for phase 1 launch

PR #1023, opened 2020-09-15
fix email env vars in dev

attwad:
Tried to reload dev today but it failed because the env vars weren't defined.
Just setting them to empty for now until a proper account is available.


verification/curator-service/api/src/util/validate-env.ts
```diff
@@ -59,7 +59,8 @@ export default function validateEnv(): Readonly<{
             devDefault: '',
         }),
         EMAIL_USER_PASSWORD: str({
-            desc: 'MongoDB URI provided to MongoClient.',
```
axmb, line 62: /shamecube

axmb (approved):

PR #1024, opened 2020-09-15
Fix flaky test by waiting on all requests

allysonjp715:


attwad (approved):
Nice, I just got hit by that I think...

PR #1027, opened 2020-09-15
remove reader from roles, all logged-in users can now see the linelist and sources

### LINKED ISSUE #1025: Get rid of reader role
Data is available publicly, we should just require logging-in to view the linelist data in the curator portal.
This will simplify things a bit as well as we have one less role check to do.

This has been requested and approved by @sratcliffe118

attwad:
Fixes #1025 

I will remove the reader roles from existing users with the mongo shell command:
`db.users.updateMany({}, {$pull: {roles: 'reader'}})`
so that updates pass validation.

### DIFF
--- a/verification/curator-service/auth.md
+++ b/verification/curator-service/auth.md
@@ -12,7 +12,6 @@ The different roles are:
 
 - admin
 - curator
-- reader
 
 A user can have no roles.
 
@@ -20,11 +19,11 @@ Here are the permissions granted to each roles:
 
 - Admin
   - List users based on their roles
-  - Assign/unassign role to users 
+  - Assign/unassign role to users
 - Curator
   - CRUD cases and ingestion sources
-- Reader
-  - Read-only access to cases and sources
+
+By default all logged-in users have read-only access to the linelist data and sources used to populate it.
 
 ## Unit-testing
 
--- a/verification/curator-service/ui/src/components/App.tsx
+++ b/verification/curator-service/ui/src/components/App.tsx
@@ -304,8 +304,7 @@ export default function App(): JSX.Element {
             text: 'Linelist',
             icon: <ListIcon />,
             to: { pathname: '/cases', state: { search: '' } },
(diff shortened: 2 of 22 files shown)


verification/curator-service/auth.md
```diff
@@ -12,19 +12,18 @@ The different roles are:
...
 - Admin
   - List users based on their roles
-  - Assign/unassign role to users 
+  - Assign/unassign role to users
 - Curator
   - CRUD cases and ingestion sources
-- Reader
-  - Read-only access to cases and sources
+
+By default all logged-in users have read-only access to the linelist data and sources used to populate it.
```
allysonjp715, line 26: Readers don't have access to sources

verification/curator-service/ui/src/components/App.tsx
```diff
@@ -304,8 +304,7 @@ export default function App(): JSX.Element {
             text: 'Linelist',
             icon: <ListIcon />,
             to: { pathname: '/cases', state: { search: '' } },
-            displayCheck: (): boolean =>
-                hasAnyRole(['reader', 'curator', 'admin']),
+            displayCheck: (): boolean => user.email !== '',
```
allysonjp715, line 307: Is it `''` or `undefined` when a user isn't logged in? Could this just be `displayCheck: (): boolean => user.email` ?

allysonjp715 (approved):

verification/curator-service/auth.md
attwad: they do, on purpose so that folks can see where the data came from (it was the case before, this didn;t change with this PR).

verification/curator-service/ui/src/components/App.tsx
attwad: the default value for user in the app component is an empty user with all fields set to their default value so I have to check for email here (it's done as well like that below for other parts of the App component), we could change to undefined but I'd prefer to do that in another PR.

PR #1029, opened 2020-09-15
make user undefined in app component instead of working with a user struct full of empty fields

attwad:


verification/curator-service/ui/src/components/App.tsx
```diff
@@ -531,12 +526,12 @@ export default function App(): JSX.Element {
...
                     <div className={classes.drawerHeader} />
                     <Switch>
-                        {user.email && (
+                        {user && (
                             <Route exact path="/cases">
                                 <LinelistTable user={user} />
                             </Route>
                         )}
-                        {hasAnyRole(['curator']) && (
+                        {user && hasAnyRole(['curator']) && (
```
allysonjp715, line 534: Can the user check be inside of hasAnyRole so this doesn't need to be added everywhere?

verification/curator-service/ui/src/components/SourceTable.tsx
```diff
@@ -480,6 +480,9 @@ class SourceTable extends React.Component<Props, SourceTableState> {
                                     <SourceRetrievalButton sourceId={row._id} />
                                 ),
                                 editable: 'never',
+                                hidden: !this.props.user?.roles.includes(
```
allysonjp715, line 483: This was removed because you can only access the Sources table if you are a curator

verification/curator-service/ui/src/components/App.tsx
attwad: no this is a typecheck for react, there is already a check in hasAnyRole for !user but apparently it is not enough.

allysonjp715: Gotcha. May not need it on this line anymore since we don't pass user into the SourceTable

allysonjp715 (approved):

verification/curator-service/ui/src/components/App.tsx
attwad: indeed, thanks. removed.

PR #1034, opened 2020-09-15
Increase resources available to prod jobs

attwad:
For #961

### DIFF
--- a/aws/README.md
+++ b/aws/README.md
@@ -192,7 +192,7 @@ To push a new release follow the [github UI flow](https://github.com/globaldothe
 
 Tag main with the `0.1.2` tag:
 
-```
+```shell
 git checkout main
 git tag 0.1.2
 ```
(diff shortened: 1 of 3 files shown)


axmb (approved):
Can you tl;dr how you determined these? More resources would definitely help with processing batch requests, but with Fargate there's no nice presized cut-offs to choose.

attwad:
> Can you tl;dr how you determined these? More resources would definitely help with processing batch requests, but with Fargate there's no nice presized cut-offs to choose.

We're not on fargate anymore, those are standard machines with fixed ram and CPU available, I looked at the `kubectl describe nodes` and `kubectl describe pods` to get those + a bit of trial and error to make them actually fit (had requested too much in the first commit).

FTR each node has:

```
Capacity:
 attachable-volumes-aws-ebs:  25
 cpu:                         2
 ephemeral-storage:           20959212Ki
 hugepages-1Gi:               0
 hugepages-2Mi:               0
 memory:                      2002020Ki
 pods:                        11
```

PR #1048, opened 2020-09-16
Add ability to search for verification status

### LINKED ISSUE #1010: Filtering of verification status in curator portal
We discussed the addition of filters to the curator portal GUI to facilitate processing  @allysonjp71 @axmb 

e.g. Allow for filtering on verification status so all verified entries can be dropped
e.g. Sorting by Confirmed date so all cases in same time period can be aggregated together more quickly 5

allysonjp715:
Fixes #1010


data-serving/data-service/schemas/cases.indexes.json
```diff
@@ -13,7 +13,8 @@
             "location.place": "text",
             "location.name": "text",
             "pathogen.name": "text",
-            "caseReference.sourceUrl": "text"
+            "caseReference.sourceUrl": "text",
+            "caseReference.verificationStatus": "text"
```
attwad, line 17: this is for full text search, you'll want to add an index below for this field to make it searchable using a keyword

allysonjp715: Oh got it, thanks. I was copying sourceUrl which didn't have one, so I added a keyword index for both.

attwad (approved):

PR #1051, opened 2020-09-16
Save page number of line list table in browser history

### LINKED ISSUE #1006: After opening a line list entry in the Linelist view on the website on page 'X' after closing it user should stay on same page 'X' not on page 1 as it is now
When I open a line list entry in the Linelist view on the website on let's say page 4 and close it again it defaults back to Page 1. This makes data validation a bit tricky.

1. Go to 'https://curator.ghdsi.org/cases'
2. Click on arrow to go to page 2 on line list view
Screenshot 2020-09-14 at 11 52 32

3. Open line list case.
4. Close line list case
5. You should be defaulted back to page one.

The same behaviour is true when using the filtered view.

It would be better to stay on that same page or same filtering view.

allysonjp715:
Fixes #1006

This way if a user:
 - goes to next page
 - clicks view case
 - closes view case

The page on the linelist will be the page they left, not necessarily the first page.

This will work once mbrn/material-table#964 is fixed. Tests can be added at that point.

### DIFF
--- a/verification/curator-service/ui/src/components/LinelistTable.tsx
+++ b/verification/curator-service/ui/src/components/LinelistTable.tsx
@@ -80,6 +80,7 @@ interface LocationState {
     editedCaseIds: string[];
     bulkMessage: string;
     search: string;
+    page: number;
 }
 
 interface Props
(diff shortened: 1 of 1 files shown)


iamleeg:
Closing this out as it's ancient, but feel welcome to add commits to the branch and resubmit if you still think this is important!

PR #1123, opened 2020-09-17
do not show Home button when user has no roles, redirect to /cases

### LINKED ISSUE #1106: Remove the home tab for reader access users
There is no content on the home tab for a user with read access to the G.h list (see screenshot)

Change: For reader profiles remove the home tab and land users on the line list tab
Screen Shot 2020-09-17 at 12 00 03 PM

attwad:
Fixes #1106

### DIFF
--- a/verification/curator-service/ui/cypress/integration/components/AppTest.spec.ts
+++ b/verification/curator-service/ui/cypress/integration/components/AppTest.spec.ts
@@ -53,7 +53,7 @@ describe('App', function () {
         cy.visit('/');
 
         cy.contains('Create new').should('not.exist');
-        cy.contains('Charts');
+        cy.contains('Charts').should('not.exist');
         cy.contains('Linelist').should('not.exist');
         cy.contains('Sources').should('not.exist');
         cy.contains('Uploads').should('not.exist');
--- a/verification/curator-service/ui/cypress/integration/components/AppTest.spec.ts
+++ b/verification/curator-service/ui/cypress/integration/components/AppTest.spec.ts
@@ -66,7 +66,7 @@ describe('App', function () {
         cy.visit('/');
 
         cy.contains('Create new').should('not.exist');
-        cy.contains('Home');
+        cy.contains('Home').should('not.exist');
         cy.contains('Linelist');
         cy.contains('Sources').should('not.exist');
         cy.contains('Uploads').should('not.exist');
--- a/verification/curator-service/ui/cypress/integration/components/AppTest.spec.ts
+++ b/verification/curator-service/ui/cypress/integration/components/AppTest.spec.ts
@@ -9,8 +9,8 @@ describe('App', function () {
         cy.visit('/cases');
         cy.url().should('eq', 'http://localhost:3002/cases');
 
-        cy.contains('Home');
-        cy.contains('span', 'Home').click();
+        cy.contains('Charts');
+        cy.contains('span', 'Charts').click();
         cy.url().should('eq', 'http://localhost:3002/');
     });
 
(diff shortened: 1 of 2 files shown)


verification/curator-service/ui/src/components/App.tsx
```diff
@@ -313,7 +320,8 @@ export default function App(): JSX.Element {
             text: 'Home',
```
allysonjp715, line 320: WDYT about calling this charts and making it visible to curators and admins? And always having the home page be the linelist page? Doesn't seem too useful of a homepage for non-readers either. We could then add a different homepage that is useful to everyone (pending UX).

I think this current logic also shows this home page for not-logged in users which we don't want as it's empty.

attwad: Done.

- FYI logged out users currently see this
image

not super useful but once we have an about page (/terms) seem a bit too "harsh" as a front page I think it'd be a good candidate.

- readers only go to linelist
- curators end up on the charts yes which is the "Home" page, so in that sense it makes sense to rename it to charts pending a proper home page from UX.

verification/curator-service/ui/cypress/integration/components/AppTest.spec.ts
```diff
@@ -66,7 +66,7 @@ describe('App', function () {
         cy.visit('/');
 
         cy.contains('Create new').should('not.exist');
-        cy.contains('Home');
+        cy.contains('Home').should('not.exist');
```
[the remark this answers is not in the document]
attwad: done

verification/curator-service/ui/src/components/App.tsx
```diff
@@ -310,10 +317,11 @@ export default function App(): JSX.Element {
     const location = useLocation<LocationState>();
     const menuList = [
         {
-            text: 'Home',
+            text: 'Charts',
             icon: <HomeIcon />,
             to: '/',
-            displayCheck: (): boolean => true,
+            // Readers only see the linelist, home button is useless to them.
+            displayCheck: (): boolean => user !== undefined && !isJustReader(),
```
allysonjp715, line 324: Seems simpler as `hasAnyRole(['curator', 'admin'])` and matches the other displayCheck logic

verification/curator-service/ui/src/components/App.tsx
```diff
@@ -567,6 +582,11 @@ export default function App(): JSX.Element {
                 >
                     <div className={classes.drawerHeader} />
                     <Switch>
+                        {isJustReader() && (
```
allysonjp715, line 585: `!hasAnyRole(['curator', 'admin'])`

verification/curator-service/ui/src/components/App.tsx
```diff
@@ -567,6 +582,11 @@ export default function App(): JSX.Element {
...
+                            <Route exact path="/">
+                                <Redirect to="/cases" />
```
allysonjp715, line 587: Could this redirect go inside the existing `<Route exact path="/">` at the bottom of this switch (for non curator/admins)?

verification/curator-service/ui/src/components/App.tsx
allysonjp715: Cool, thanks. Filed #1133 for design of a simple home page.

verification/curator-service/ui/src/components/App.tsx
attwad: done.

verification/curator-service/ui/src/components/App.tsx
attwad: yes I moved it there now.

allysonjp715 (approved):

Issue #1098, opened 2020-09-16
Pop-up for first-time users on Line List Platform/G.h Map


dakrownie:
[Link to beta testing blocker list](https://docs.google.com/spreadsheets/d/1wM6se-GPNic3xxyggXPEsk0W6qMo6zl_RRcEwU6OcoU/edit#gid=1506108781)

Need to add a first time use state, status: Design TBD

dakrownie:
Assigned to tommyshimko@google.com and Shroo

shroothless:
See mocks [here](https://www.figma.com/file/gIpIu2cyO3MWXQ86Npb0ab/G.h-Map?node-id=1085%3A1472)

We will need content from the marketing team. I've also added links to the terms of services and data dictionary

allysonjp715:
I have this code ready, but I don't think it makes sense to add to the build until we have the proper content to go in the dialog. Who will create that content?

sratcliffe118:
@dakrownie who from marketing can pick this up?

dakrownie:
matilde@google.com is handling this.
There is an email thread with @sratcliffe118 and @shroothless in copy.

allysonjp715:
I just filed #1133 and thought, what if this isn't a first time popup but instead the home page or about page? WDYT @shroothless ?

shroothless:
I don’t think there’s enough content there to be a home page yet? We could
add this content to an about page as well though.

allysonjp715:
Closing out in favor of the homepage as per https://www.figma.com/file/gIpIu2cyO3MWXQ86Npb0ab?node-id=1085:1472#41702084

PR #1126, opened 2020-09-17
Add API for the welcome popup

allysonjp715:
#1098

### DIFF
--- a/verification/curator-service/api/openapi/openapi.yaml
+++ b/verification/curator-service/api/openapi/openapi.yaml
@@ -834,15 +834,23 @@ paths:
           pattern: '^[a-f\d]{24}$'
     put:
       tags: [User]
-      summary: Updates a specific user's roles
+      summary: Updates a specific user. Can only update roles and hasSeenWelcomePopup.
       operationId: updateUser
       requestBody:
-        description: New roles for the user
+        description: New properties for the user
         required: true
         content:
           application/json:
             schema:
-              $ref: "#/components/schemas/RoleArray"
+              type: object
+              additionalProperties: false
+              properties:
+                hasSeenWelcomePopup:
+                  type: boolean
+                roles:
+                  type: array
+                  items:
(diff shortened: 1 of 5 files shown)


verification/curator-service/api/openapi/openapi.yaml
```diff
@@ -834,15 +834,23 @@ paths:
           pattern: '^[a-f\d]{24}$'
     put:
       tags: [User]
-      summary: Updates a specific user's roles
+      summary: Updates a specific user. Can only update roles and hasSeenWelcomePopup.
```
attwad, line 837: nit but I'd say this is implied by the arguments below

verification/curator-service/api/src/model/user.ts
```diff
@@ -14,6 +14,10 @@ const userSchema = new mongoose.Schema({
         enum: userRoles,
     },
     picture: String,
+    hasSeenWelcomePopup: {
+        type: Boolean,
+        default: false,
```
attwad, line 19: isn't the default false for a boolean?

attwad (approved):

verification/curator-service/api/openapi/openapi.yaml
allysonjp715: I like to spell it out in English too, in case someone missed the additionalProperties tag or they're quickly scanning through the docs.

verification/curator-service/api/src/model/user.ts
allysonjp715: I don't see any mention in the docs of what the default value for booleans is. It could potentially be `undefined` as well. However clients should be prepared to deal with an undefined value, so I removed the default here.

allysonjp715:
Going to wait to submit in case we want a home page rather than a popup: https://github.com/globaldothealth/list/issues/1098#issuecomment-694771101

attwad:
Can now close that PR I believe?

PR #1128, opened 2020-09-17
Log with Pino

axmb:
Gets us a pretty condensed standard log format.

image

Per my comment, prettifying isn't recommended in production. That said, we aren't request logging, so the overall load is modest. And the trade-off is that we don't have to pipe local/kubectl logs through (a globally installed for everyone) `pino-pretty`.

I'm happy to change that if folks prefer, but thought it'd be better to "just work" for now.

Includes converting a few statements (handling err in controllers) that are currently info-level to error-level.


verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -163,8 +164,8 @@ export class AuthController {
 
     /**
      * Configures OAuth passport strategy.
-     * @param clientID the OAuth client ID as gotten from the Google developer console.
-     * @param clientSecret the OAuth client secret as gotten from the Google developer console.
+     * @param clientID the OAuth client ID as gotten from the Google developer logger.
```
attwad, line 167: Who's that Google developer logger? can I talk to them? 😶

verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -163,8 +164,8 @@ export class AuthController {
...
+     * @param clientSecret the OAuth client secret as gotten from the Google developer logger.
```
attwad, line 168: nope

attwad (approved):
prettyprinting sounds reasonable given your comments, thanks

verification/curator-service/api/src/controllers/auth.ts
axmb: oh no

PR #1130, opened 2020-09-17
Upgrade deps/patch a high severity vulnerability

axmb:


attwad (approved):
Looks good but why are no checks run on such PR?

axmb:
> Looks good but why are no checks run on such PR?

Seems to be because our workflows are configured to run on PRs against `main`.

Might as well change that to all PRs, IMO. Anything semi-non-idempotent should be triggered on `push`, regardless.

PR #1137, opened 2020-09-18
Sort uploads by most recent first

### LINKED ISSUE #1116: Change upload list ordering from latest date to oldest
**Describe the bug**
Change upload list ordering from latest date to oldest

Currently set from oldest to newest so need to scroll through to end to check out the latest uploads for verification
Screen Shot 2020-09-17 at 1 03 24 PM

allysonjp715:
For #1116

### DIFF
--- a/verification/curator-service/api/src/controllers/uploads.ts
+++ b/verification/curator-service/api/src/controllers/uploads.ts
@@ -115,6 +115,7 @@ export default class UploadsController {
                             upload: '$uploads',
                         },
                     },
+                    { $sort: { 'upload.created': -1 } },
                 ]),
                 Source.aggregate([
                     { $unwind: '$uploads' },
--- a/verification/curator-service/api/src/controllers/uploads.ts
+++ b/verification/curator-service/api/src/controllers/uploads.ts
@@ -115,7 +115,7 @@ export default class UploadsController {
                             upload: '$uploads',
                         },
                     },
-                    { $sort: { 'upload.created': -1 } },
+                    { $sort: { 'upload.created': -1, sourceName: -1 } },
                 ]),
                 Source.aggregate([
                     { $unwind: '$uploads' },
--- a/data-serving/data-service/package-lock.json
+++ b/data-serving/data-service/package-lock.json
@@ -1765,11 +1765,12 @@
       }
     },
     "@types/express-serve-static-core": {
-      "version": "4.17.5",
-      "resolved": "https://registry.npmjs.org/@types/express-serve-static-core/-/express-serve-static-core-4.17.5.tgz",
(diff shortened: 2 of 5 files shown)


verification/curator-service/api/src/controllers/uploads.ts
```diff
@@ -115,6 +115,7 @@ export default class UploadsController {
                             upload: '$uploads',
                         },
                     },
+                    { $sort: { 'upload.created': -1 } },
```
axmb, line 118: This is probably cool if `created` has timestamp resolution, but we probably want an explicit secondary sort otherwise.

axmb (approved):

verification/curator-service/api/src/controllers/uploads.ts
attwad: you'll need an index as well for sorting, look at the schema for cases for an example (it should have a sort-related comment on it)

attwad: I meant this:
https://github.com/globaldothealth/list/blob/main/data-serving/data-service/schemas/cases.indexes.json#L162

but note that it's not going to be an issue unless we reach a significant amount of uploads to go through so probably find to do in another PR (we started seeing OOMs due to sorting without an index at around 170K cases).

allysonjp715: Got it. I'll add the index in a separate PR since it will take some figuring out how to set up the first index for the curator-service (and I want to get the build fixes in).  

In terms of a secondary sort, I think it's okay to default to the order that was happening before we added this (I believe based on order added to the DB). WDYT?

axmb: I'd say one of two things:

1. If `created` is a time (and not just a date), show the HH:MM and no need for a secondary sort.
2. If `created` is a date, like it's displayed, we should probably secondary sort by the source name. Just so it's easier to find the relevant upload and the table appears to be organized.

allysonjp715: Added secondary sort on source name

z023:
Uploads still listed as oldest first. Where is this issue at?

allysonjp715:
> Uploads still listed as oldest first. Where is this issue at?

We just hadn't pushed yet. I just pushed dev and you can now see the correct ordering there.

z023:
I'm currently seeing each uploads page is sorted by most recent rather than the whole uploads list being sorted by most recent.

allysonjp715:
> I'm currently seeing each uploads page is sorted by most recent rather than the whole uploads list being sorted by most recent.

Indeed, silly mistake, will sent PR.

z023:
Still seeing the same issue.

allysonjp715:
Hm this doesn't seem fixed even after a push. Will reopen #1116

z023:
Any progress with this?

allysonjp715:
#1303

allysonjp715:
Fix has pushed to prod now and it lgtm.

PR #1141, opened 2020-09-18
Mexico

calremmel:
Here's a first take for the Mexico parser!

Big question is what to do with the huge municipalities dictionary. Can I store this elsewhere, or is this the best solution?


ingestion/functions/parsing/mexico/input_event.json
```diff
@@ -0,0 +1,7 @@
+{
+    "env": "local",
+    "s3Bucket": "epid-sources-raw",
+    "s3Key": "5f59e8c47ef66004bc9c91ce/2020/09/10/0855/content.csv",
+    "sourceUrl": "http://datosabiertos.salud.gob.mx/gobmx/salud/datos_abiertos/datos_abiertos_covid19.zip",
```
attwad, line 5: is the original source really a zip?? in which case we'll have to implement unzipping before parsing. (worth filing an issue to track that if that's true)

attwad (commented):
For the huge dictionaries, it'd be better to store them as CSVs or JSON besides the code but not **in** the code itself for readability and maintainability. Lambdas bundle all files in the directory together so you can access them easily and load them in memory when you start parsing.

ingestion/functions/parsing/mexico/input_event.json
calremmel: Yes unfortunately it is! Chatted it over with AB a little, yeah we'll either have to implement that or rehost every time it is updated. I'll file an issue.

calremmel:
> For the huge dictionaries, it'd be better to store them as CSVs or JSON besides the code but not **in** the code itself for readability and maintainability. Lambdas bundle all files in the directory together so you can access them easily and load them in memory when you start parsing.

Thanks, that's the part I wasn't sure about. I'll store them as JSON in the same directory.

ingestion/functions/parsing/mexico/mexico.py
```diff
@@ -0,0 +1,2706 @@
...
+        demo["gender"] = convert_gender(gender)
+    if age:
+        demo["ageRange"] = {"start": float(age), "end": float(age)}
+    if nationality == "1":
+        demo["nationalities"] = "Mexican"
+    elif nationality == "2":
+        demo["nationalities"] = national_origin
+    if demo == {}:
+        return None
+    return demo
```
axmb, line 2620: If you'd like, can shorten ~equivalently to `return demo or None`

axmb (approved):
I don't feel strongly about storing the dictionaries. Remainder of the parser looks great, though we'll have to figure out the zip issue as discussed.

ingestion/functions/parsing/mexico/mexico.py
calremmel: Cool!!!!

iamleeg (approved):
Thanks @calremmel!

.gitignore
```diff
@@ -377,6 +377,7 @@ dmypy.json
 !.vscode/tasks.json
 !.vscode/launch.json
 !.vscode/extensions.json
+*.vscode
```
iamleeg, line 380: I'd normally put editor warts into my user's global `.gitignore` so that my projects don't end up with everybody's different editor configurations in there, but it's not a big deal.

calremmel: I didn't know that there was a global .gitignore! Super, super useful, thanks for this

attwad:
@Mougk the continuous integration tests were not passing for this PR, please do not merge into main stuff that's not green.
https://github.com/globaldothealth/list/runs/1162656692

Apparently the path to `dictionaries.json` has to be made relative, I will try to fix that now.

attwad:
@calremmel the fix FTR: https://github.com/globaldothealth/list/commit/a7df1b3d3d5ba01033c6de2099d64cb03a7b16ac

Mougk:
> @Mougk the continuous integration tests were not passing for this PR, please do not merge into main stuff that's not green.
> https://github.com/globaldothealth/list/runs/1162656692
> 
> Apparently the path to `dictionaries.json` has to be made relative, I will try to fix that now.

My bad. Was too ambitious late Friday night.

PR #1147, opened 2020-09-21
Japan parser update

AnyaLindstromBattle:
Updated Japan parser to be come comprehensive, namely:
1.confirmedPatient is true or false, we only want to import the true cases.
2. sourceURL and deathSourceURL should be put as extra sources url in caseReferences.additionalSources
3. deceasedDate should be an event
4. patientStatus should be parsed as well


ingestion/functions/parsing/japan/japan.py
```diff
@@ -1,3 +1,6 @@
+#Now change japan_test and sample data as well
```
attwad, line 1: remove comment please

ingestion/functions/parsing/japan/japan.py
```diff
@@ -43,23 +43,48 @@ def convert_location(raw_entry):
...
     """
-    date = datetime.strptime(raw_date, "%Y-%m-%d")
+    date = datetime.strptime(raw_date["dateAnnounced"], "%Y-%m-%d")
     return date.strftime("%m/%d/%YZ")
 
+def convert_death_date(raw_death_date: str):
+    if "deceasedDate" in raw_death_date:
+        death_date = datetime.strptime(raw_death_date["deceasedDate"], "%Y-%m-%d")
+        return death_date.strftime("%m/%d/%YZ")
+    else:
```
attwad, line 65: the else part just returning None is not needed

ingestion/functions/parsing/japan/japan.py
```diff
@@ -43,23 +43,48 @@ def convert_location(raw_entry):
...
+        return None
+
+def convert_additionalSources(additional_sourceURL: str):
+    additionalSources_list = []
+    additionalSources = {}
+    if "sourceURL" in additional_sourceURL:
+        additionalSources["sourceUrl"] = additional_sourceURL["sourceURL"]
+    if "deathSourceURL" in additional_sourceURL:
```
attwad, line 73: If you have both source and death urls the death one will overwrite the source one.
Consider rewriting this method so it's simpler:

```
sources = []
if "sourceURL" in additional_sourceURL:
  sources.append({"sourceUrl": additional_sourceURL["sourceURL"])
if "deathSourceURL" in additional_sourceURL:
  sources.append({"sourceUrl": additional_sourceURL["deathSourceURL"])
return sources or None
```

ingestion/functions/parsing/japan/japan.py
```diff
@@ -72,26 +97,35 @@ def parse_cases(raw_data_file, source_id, source_url):
...
-                            "end": convert_date(entry["dateAnnounced"]),
+                            "start": convert_date(entry),
+                            "end": convert_date(entry),
                         }
+                    },
+                    {
+                        "name": "outcome",
+                        "dateRange":
+                        {
+                            "start": convert_death_date(entry),
```
attwad, line 117: why always put death date here if outcome can be unknown?

ingestion/functions/parsing/japan/japan.py
```diff
@@ -72,26 +97,35 @@ def parse_cases(raw_data_file, source_id, source_url):
...
                     }
                 ],
                 "demographics": {
                     "ageRange": convert_age(entry),
                     "gender": convert_gender(entry)
                 },
                 "notes": detect_notes(entry)
-            } for entry in cases if entry["patientId"] != "-1")
-
+            } for entry in cases if (entry["patientId"] != "-1" and entry["confirmedPatient"] == True))
```
attwad, line 128: `== True` is not needed

ingestion/functions/parsing/japan/japan_test.py
```diff
@@ -24,6 +29,15 @@
                             "start": "03/13/2020Z",
                             "end": "03/13/2020Z"
                         }
+            },
+            {
+                "name": "outcome",
+                "dateRange":
```
attwad, line 35: as per my other comment, you should'nt put a dateRange with null start and end fields if the outcome is unknown

ingestion/functions/parsing/japan/japan.py
```diff
@@ -43,23 +40,48 @@ def convert_location(raw_entry):
...
     Convert raw date field into a value interpretable by the dataserver.
-
+​
     The date filtering API expects mm/dd/YYYYZ format.
     """
-    date = datetime.strptime(raw_date, "%Y-%m-%d")
+    date = datetime.strptime(raw_date["dateAnnounced"], "%Y-%m-%d")
     return date.strftime("%m/%d/%YZ")
 
+def convert_additionalSources(additional_sourceURL: str):
```
axmb, line 58: Isn't this `Dict`, rather than `str`? Likewise for `convert_outcome` and `convert_date`.

https://docs.python.org/3/library/typing.html

axmb, line 58: Total nit, but would be good to consistently use snake_case here (JSON fields which we don't control aside, of course). So `convert_additional_sources` and `additional_source_url`, e.g..

ingestion/functions/parsing/japan/japan.py
```diff
@@ -43,23 +40,48 @@ def convert_location(raw_entry):
...
+                    "value": "Death"}
+        elif raw_outcome["patientStatus"] == "Discharged" or raw_outcome["patientStatus"] == "Recovered":
+            return {"name": "outcome",
+                    "value": "Recovered"}
+        else:
+            return {"name": "outcome",
+                    "value": "Unknown"}       
+    else:
+        return {"name": "outcome",
+                "value": "Unknown"}
```
axmb, line 84: Can make this a little simpler to follow by deleting the first else (79-81) and removing the `else:` part of the second block (82).

axmb (approved):
Nice! Good catch re: confirmedPatient.

ingestion/functions/parsing/japan/japan.py
AnyaLindstromBattle: sorry yes, copy pasted some functions and forgot to change it... Thanks!

attwad (approved):
lgtm, modulo ab's comments being addressed

ingestion/functions/parsing/japan/japan.py
AnyaLindstromBattle: Isn't the input a str here?

attwad: No it's a dict

z023:
There are currently 729 cases coming through as Indonesia. ~70% of the IDs are duplicated. The ingestion has resulted in all their locations being set as a small place in Indonesia called 'Japan'. They were all identified in airport quarantine or are quarantine officers.

z023:
The 2 cases coming through as Missouri USA were also identified in airport quarantine and their residence is unknown.

ingestion/functions/parsing/japan/japan.py
```diff
@@ -43,23 +41,44 @@ def convert_location(raw_entry):
...
-def convert_date(raw_date: str):
+def convert_date(raw_date: Dict):
     """
     Convert raw date field into a value interpretable by the dataserver.
-
+​
     The date filtering API expects mm/dd/YYYYZ format.
     """
-    date = datetime.strptime(raw_date, "%Y-%m-%d")
+    date = datetime.strptime(raw_date["dateAnnounced"], "%Y-%m-%d")
```
calremmel, line 56: If you wanted, you could have the function accept entry["dateAnnounced"] as a string, rather than passing in the whole entry. That way, you could reuse this function to process death_date rather than having to repeat the logic. It looks like it works fine as is, though.

ingestion/functions/parsing/japan/japan.py
```diff
@@ -43,23 +41,44 @@ def convert_location(raw_entry):
...
+    return sources or None
+
+def convert_outcome(raw_outcome: Dict, raw_death_date: Dict):
+    if "patientStatus" in raw_outcome:
+        if raw_outcome["patientStatus"] == "Deceased":
+            death_date = datetime.strptime(raw_death_date["deceasedDate"], "%Y-%m-%d")
+            return {"name": "outcome",
+                    "dateRange":{
+                        "start": death_date.strftime("%m/%d/%YZ"),
+                        "end": death_date.strftime("%m/%d/%YZ")
```
calremmel, line 74: As before, you could probably fix it so that convert_date() handles this as well. Up to you!

z023:
There are still cases coming through as Indonesia. @AnyaLindstromBattle this might be something to discuss with Felix as he sorted the geocoding for Argentina when cases were coming through as Mexico.

AnyaLindstromBattle:
@z023 thanks for the heads up. How many cases are coming through as Indonesia? 

I believe this is partly what @attwad and @axmb were trying to sort out in #1278. i.e. what the change in that PR will allow is to specify specific cases which should be ignored (the Indonesia ones in this case). However this hasn't been merged as they were holding off until Japan was backfilled, has that been done? If so we can merge that PR and then I can change the parser accordingly.

z023:
There are currently 737 that have come through as Indonesia (see line list download below). As Japan is not too large the parser is able to ingest the whole dataset each day so no backfilling was needed.
[2020.11.02 Japan_Source_Indonesia_Cases.xlsx](https://github.com/globaldothealth/list/files/5476291/2020.11.02.Japan_Source_Indonesia_Cases.xlsx)

AnyaLindstromBattle:
Nice one, will merge that PR then. Ok as there are quite a few cases I'll have another look and chat to Felix about it as you suggested.

Issue #1107, opened 2020-09-17
Slow line list download speeds


sratcliffe118:
- Do we have a sense for the speeds when a user clicks the download button?
- I just tested on a 70-75Mb per second connection in UK for 137k entries and it is still running +2mins
- This feels too slow for users 

@Mougk Are there any download speed benchmarks / datasets we can look to?

attwad:
Took me 1.4min to start the content download of 130K cases:

![image](https://user-images.githubusercontent.com/1255432/93466686-f2efe600-f8ec-11ea-8bf8-4c8189d2316c.png)

That's quite a long time.

After that network seems to be the bottleneck as the servers aren't using any crazy CPU or RAM at all

```
kubectl top pods
NAME                            CPU(cores)   MEMORY(bytes)
curator-prod-7bd48b496f-tbrtb   3m           53Mi
data-prod-c469b5ccf-rsqtj       18m          44Mi
```

Download seems to have stopped altogether after 2.2min though and no sign of anything going on other than that spinning widget in the download button.

![image](https://user-images.githubusercontent.com/1255432/93467261-ce483e00-f8ed-11ea-860f-ea00b1e2da42.png)

I'm not sure if we can set some special http headers to start the download immediately in the browser instead of waiting for what seems like the complete file to be sent. @allysonjp715 in case she has any thoughts?

attwad:
This might be relevant: https://nodejs.org/pt-br/docs/guides/backpressuring-in-streams/

Also, we should probably gzip the CSV before sending it, it will reduce the size of the download and maybe make it faster?
The stream pipeline stuff of node described in this website looks nice, perhaps we can use it in the curator service as well to do this?

allysonjp715:
Just found this, could be part of the problem: https://github.com/axios/axios/issues/479

attwad:
Also, I've looked a gzip and nginx already does gzip compression of responses for us (visible in the dev tools nettwork tab), although not for the download it seems, we should be able to configure it so that it does:

![image](https://user-images.githubusercontent.com/1255432/93755807-3bb4e100-fc04-11ea-9d78-199ae6626767.png)

attwad:
> Just found this, could be part of the problem: [axios/axios#479](https://github.com/axios/axios/issues/479)

I don't think it applies to us though? We're using node on the backend, not in the browser.

allysonjp715:
[...]

PR #1151, opened 2020-09-21
Compress text/csv responses in nginx

attwad:
For #1107


allysonjp715 (approved):
Have you confirmed this doesn't change the resulting downloaded file?

attwad:
> Have you confirmed this doesn't change the resulting downloaded file?

Yes, note that it's nginx doing the compression so I wouldn't expect a bug in there honestly but just to be sure I've cleared a bit of data in dev (to fit in memory) and downloaded it, the file looks fine.

also that's quite a transfer improvement 👍 
image

Issue #1155, opened 2020-09-21
Add ability to ingest data sources stored in zip files


calremmel:
The dataset for Mexico is stored as a zip file, as follows:
http://datosabiertos.salud.gob.mx/gobmx/salud/datos_abiertos/datos_abiertos_covid19.zip

It would be great if it were possible for data ingestion to automatically handle data sources that are stored as zip files or other compressed file formats.

Alternatively, we could rehost data sources stored this way somehow, and link to an extracted version on our own servers, but that seems less than ideal.

attwad:
The retrieval function could uncompress the data when it stores it, let me look at how we could do that.

PR #1157, opened 2020-09-21
uncompress zip files when source URL points to one

attwad:
Advantage is that it requires no configuration per source, disadvantage is that it makes assumptions like having one file in the zip file that we're interested in. Fortunately I think that's the case for Mexico data that we're trying to ingest so it seems fine to me. I couldn't confirm though as I can't download the original zip file (need to vpn somewhere it seems, can't DL from Europe :( ).

For #1155

### DIFF
--- a/ingestion/functions/retrieval/retrieval.py
+++ b/ingestion/functions/retrieval/retrieval.py
@@ -77,7 +77,7 @@ def get_source_details(env, source_id, upload_id, api_headers, cookies):
             e, env, common_lib.UploadError.INTERNAL_ERROR, source_id, upload_id,
             api_headers, cookies)
 
-def raw_content(url:str, content: bytes) -> io.BytesIO:
+def raw_content(url: str, content: bytes) -> io.BytesIO:
     # Detect the mimetype of a given URL.
     print(f'Guessing mimetype of {url}')
     mimetype, _ = mimetypes.guess_type(url)
--- a/ingestion/functions/retrieval/retrieval.py
+++ b/ingestion/functions/retrieval/retrieval.py
@@ -1,8 +1,11 @@
 import codecs
+import io
 import json
+import mimetypes
 import os
 import sys
-import io
+import tempfile
+import zipfile
 from chardet import detect
 
 import boto3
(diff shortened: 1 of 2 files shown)


axmb:
> Advantage is that it requires no configuration per source, disadvantage is that it makes assumptions like having one file in the zip file that we're interested in. Fortunately I think that's the case for Mexico data that we're trying to ingest so it seems fine to me. I couldn't confirm though as I can't download the original zip file (need to vpn somewhere it seems, can't DL from Europe :( ).
> 
> For #1155

It's indeed a single CSV file.

ingestion/functions/retrieval/retrieval.py
```diff
@@ -1,8 +1,11 @@
 import codecs
+import io
 import json
+import mimetypes
 import os
 import sys
-import io
+import tempfile
+import zipfile
```
axmb, line 8: Sanity checking -- these are all python3 and don't require declaration in `requirements.txt`, right?

ingestion/functions/retrieval/retrieval.py
```diff
@@ -74,6 +77,25 @@ def get_source_details(env, source_id, upload_id, api_headers, cookies):
             e, env, common_lib.UploadError.INTERNAL_ERROR, source_id, upload_id,
             api_headers, cookies)
 
+def raw_content(url:str, content: bytes) -> io.BytesIO:
```
axmb, line 80: hyper-nit: `url: str`

ingestion/functions/retrieval/retrieval_test.py
```diff
@@ -373,3 +376,24 @@ def test_upload_to_s3_raises_error_on_s3_error(
 
     # We got the wrong exception or no exception, fail the test.
     assert not "Should have raised an exception."
+
+def test_decompress_zip_file():
```
axmb, line 380: nit, but `test_raw_content_unzips` and `test_raw_content_ignores_unknown_mimetypes`?

axmb (approved):

ingestion/functions/retrieval/retrieval.py
attwad: yup, python has "battery included" as they like to say

ingestion/functions/retrieval/retrieval.py
attwad: omg how can you... okay done ;)

ingestion/functions/retrieval/retrieval_test.py
attwad: done

Issue #1110, opened 2020-09-17
Line List data download user flow


sratcliffe118:
**Is your feature request related to a problem? Please describe.**
There is no visual indication as to the data download progress, time remaining or status. This plus the slow download rate makes it hard for the user to see what is going on and could result in multiple attempts to download thinking something is wrong, or give up and navigate away

**Describe the solution you'd like**
A UX treatment to show the progress of the download including the time or progress of the data download

UX treatment the user currently sees attached (the download button has a spinner and that is it)

![Screen Shot 2020-09-17 at 12 21 51 PM](https://user-images.githubusercontent.com/63205311/93463983-693a1b80-f8e0-11ea-9625-762cb8557564.png)

sratcliffe118:
@shroothless is this on your radar? Think this is a key user journey we need to make sure we have covered

shroothless:
I'm not following I think:  Step 1 Click download Step 2 .csv downloads in
browser

If we want to add some orientation hovers to call out what the download
button does we can add that?

sratcliffe118:
I just went through this process for 137k cases and the button was spinning for 5mins before i gave up.

Either we need to speed the download process up, or have some UX to show that the download is happening and how long it will take otherwise users will just give up as it is too slow

shroothless:
Got it, I can add UX treatment. Does it spin before the download starts?

sratcliffe118:
No, it starts spinning as soon as it's clicked

shroothless:
Got it, adding to my list.

allysonjp715:
The data download progress is now shown on the browser's download object (e.g. the bottom bar in chrome).

Issue #1109, opened 2020-09-17
Linelist very slow after clicking download button


sratcliffe118:
**Describe the bug**
Line list site becomes very slow once the download button is clicked

**To Reproduce**
1. Clicked download on the line list
2. navigated to home and then back to line list
3. Tabbed down a few pages

**Expected behavior**
Navigation and page loads are quick

attwad:
Is it though?
Perhaps my new computer is just too powerful.

![ezgif com-gif-maker](https://user-images.githubusercontent.com/1255432/93747847-95fb7500-fbf7-11ea-8989-ea18d4653a37.gif)

sratcliffe118:
Definitely quicker than mine.

Just tried again and it is better but still a degree of lag. Will downgrade to a P2 so we can keep track

allysonjp715:
After the most recent change the UX is snappy and users can move to other tabs. However they cannot interact with the linelist table until the entire download is complete (it shows the table as loading until it is complete). Perhaps this is due to a connection to the Cases database already being used and we can't make two queries on the database at the same time?

attwad:
We can definitely make two queries to the DB simulataneously, read queries shoudn't take a lock.

allysonjp715:
After the last PR the download is fully functional, and any part of the UI can be interacted with during the download with no extra lag. This can be viewed now on prod 👍

PR #1158, opened 2020-09-21
Use <form> to download cases in UI

allysonjp715:
This prevents the browser from storing the downloaded data in memory and greatly improves speed and performance:

For 9000 cases
Before change: 34.2 kB over 4.42s
Screen Shot 2020-09-21 at 4 26 07 PM

After change: 295B over 1.46s
Screen Shot 2020-09-21 at 4 28 48 PM

#1107 #1109 #1110

I think these bugs can be closed out after this PR but I want to double check in dev/prod with larger datasets.


verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -243,41 +241,35 @@ function RowMenu(props: {
...
+            >
+                <input name="query" value={props.search.trim()} />
+            </form>
+            <Button
+                variant="outlined"
+                color="primary"
+                onClick={(): void => formRef.current.submit()}
+                startIcon={<SaveAltIcon />}
+            >
+                Download
```
attwad, line 263: we can't get the loading spinner anymore?

attwad (approved):
one less dependency and faster? yesssss

verification/curator-service/ui/src/components/LinelistTable.tsx
allysonjp715: It's not needed anymore as the browser download object shows up quickly (the bottom panel in chrome) and shows the loading there.

attwad: nice! makes sense.

allysonjp715:
Unfortunately had to remove the cypress tests now that the HTTP response no longer contains the data. Cypress does not support interacting with file downloads: https://github.com/cypress-io/cypress/issues/5651

Issue #1124, opened 2020-09-17
Document a strategy for ingesting super-large (~100k+) data via ADI


axmb:
**Is your feature request related to a problem? Please describe.**
Countries that have all case data in a single file, and also have a large number of cases, can take a while to upload. With the current size of our prod server, the time it takes to upsert all of this data can exceed the time limit for our AWS Lambda parsing functions (15 minutes). This makes regular automation unable to backfill this data.

This is an issue for Peru, for example, which has 600K+ cases in a single file. To give a converse example, it probably isn't an issue for India, which has more cases than Peru, but has separate files each with tens-of-thousands of cases.

**Describe the solution you'd like**
A way to get this type of data into the platform.

**Additional context**
The simplest solution, and a workaround that's technically usable now, is to use the source `dateFilter` to break the data up (for instance, running it on individual dates). Unfortunately, this is pretty laborious, since we can't select a time range with the current date filter set up.

This is mostly an issue for backfilling data. In a steady state, we probably just want to examine recent history (say, a couple weeks or months) for case updates. But, now that many countries have hundreds-of-thousands of documented cases, we need to get this historical data into the platform.

Since this backfilling should be a one-time event, the simplest solution might be to support date-range date filters, and to manually change that in the UI, then trigger retrieval, to get relevant chunks of data into the platform. Alternatively, parsers for large data could trigger an exo-Lambda function (in our server, or an AWS batch job) to do this.

attwad:
I can think of a few things:

- parsers could store a hash of the file they retrieved and the last position they were at in the upload model, so that they can resume if needs be.
- date filters do not work for that problem currently as we can ingest stuff "up to a date" and "on a given date" but the "up to a date", we would require a from-to-date filter type which we do not have now to be able to ingest in chunks.
- parsers could be made to run locally as one-offs without any time-limit (should be very simple, we just need a main funtion that does more or less the same thing as what the lambda main does). That's not optimal in the sense that it requires manual intervention from a developer but if it's supposed to be a one-off until the date filters can be set on the source then it might be acceptable.

attwad:
I played a bit with adding QPS info to the logs of the parsing functions so that we have an idea of how fast they can go, here's what I have locally I think it's pretty useful, I'll fix the tests and send it for proper review later today:

```
Raw file retrieved at /tmp/tmp0_0ahaur
Sending 250 cases, total so far: 0
	Current speed: 90 cases/sec
	Max estimate: 80710 in the remaining 894 sec
[...]

PR #1159, opened 2020-09-22
Backfill feature

axmb:
Re: #1124

Felt a little complex/clunky to stuff this in the sources table, so trying this out in a separate modal accessible from the create menu.

Just drafted up the basic idea. I didn't include wiring through the supplied start/end dates through the Lambda functions.

image

The parsing Lambda execution is async, so by completing this form, a curator more or less gets the upload ID to monitor. There might be some trial-and-error around getting the right date range (e.g., including too many cases by accident), but at least a curator can do this themselves.

OTOH, if we think there are too many things that can go wrong with this, we should just implement some trivial date ranging directly in the parsing_lib, with the intention of just using that feature locally, and we can do local runs to backfill data.


verification/curator-service/api/openapi/openapi.yaml
```diff
@@ -202,6 +202,18 @@ paths:
         schema:
           type: string
           pattern: '^[a-f\d]{24}$'
+      - name: parse_start_date
+        in: query
+        description: The date on which to begin parsing/ingesting data.
```
attwad, line 207: those dates are "included" I believe, better be explicit here and/or in the openapi specs about it.

verification/curator-service/ui/src/components/AutomatedBackfill.tsx
```diff
@@ -0,0 +1,208 @@
...
+        values: AutomatedBackfillValues,
+    ): Promise<void> => {
+        try {
+            const response = await axios.post<RetrievalResult>(
+                `/api/sources/${values.caseReference?.sourceId}/retrieve
+                ?parse_start_date=${values.startDate}
+                &parse_end_date=${values.endDate}`,
+            );
+            setErrorMessage('');
+            setSuccessMessage(`Created upload ${response.data.upload_id}.`);
```
attwad, line 106: Here what would be nice would be to periodically fetch the number of cases from the upload while it's IN_PROGRESS, instead of having the curator do this manually. (For a future PR)

verification/curator-service/ui/src/components/AutomatedBackfill.tsx
```diff
@@ -0,0 +1,208 @@
...
+                                data-testid="submit"
+                                disabled={isSubmitting}
+                                onClick={submitForm}
+                            >
+                                Backfill source
+                            </Button>
+                            <Button
+                                className={classes.cancelButton}
+                                color="primary"
+                                disabled={isSubmitting}
```
attwad, line 196: nit but can you disable the modal X button as well? do we even want to do that?

verification/curator-service/ui/src/components/common-form-fields/Source.tsx
```diff
@@ -16,6 +16,7 @@ import { throttle } from 'lodash';
 interface SourceProps {
     initialValue?: CaseReference;
     hasSourceEntryId?: boolean;
+    disableFreeSolo?: boolean;
```
attwad, line 19: need more positive vibes 🧘 please~
s/disableFreeSolo/freeSolo

attwad:
I think this makes sense rather than in the sources table, good call.
With a dialog like that we can make things much simpler as well like having buttons to advance each date automatically for example and fully automatic backfill by looking at upload statuses/continuing automatically.

verification/curator-service/api/openapi/openapi.yaml
axmb: Good call, done.

verification/curator-service/ui/src/components/AutomatedBackfill.tsx
axmb: We could find a way I'm sure, but an escape hatch is probably good to keep. This came up with the other modal designs talking with UX.

verification/curator-service/ui/src/components/common-form-fields/Source.tsx
axmb: Yeah, I was being lazy about getting a default `false` behavior. Positive'd.

ingestion/functions/parsing/common/python/parsing_lib/input_event.json
```diff
@@ -4,11 +4,17 @@
...
     "sourceId": "5f0b9a7ead3a2b003edc0e7f",
-    "uploadIds": ["123456789012345678901234"],
+    "uploadIds": [
+        "123456789012345678901234"
+    ],
     "dateFilter": {
         "numDaysBeforeToday": 3,
         "op": "LT"
     },
+    "dateRange": {
```
attwad, line 14: should this be allowed with dateFilter set as well? seems contradictory

ingestion/functions/retrieval/valid_scheduled_event.json
```diff
@@ -1,4 +1,8 @@
 {
     "env": "local",
-    "sourceId": "5f6863d78a4a5200342cd2da"
+    "sourceId": "5f6863d78a4a5200342cd2da",
+    "parsingDateRange": {
```
attwad, line 4: that doesn't make sense in the retrieval event right?

verification/curator-service/api/src/clients/aws-lambda-client.ts
```diff
@@ -83,15 +83,30 @@ export default class AwsLambdaClient {
...
      * Invoke retrieval function lambda synchronously, returning its output.
+     *
+     * @param sourceId - ID of the source to be retrieved.
+     * @param parseStartDateString - YYYY-MM-DD on which to begin parsing (inclusive).
+     * @param parseEndDateString - YYYY-MM-DD on which to end parsing (inclusive).
      */
-    invokeRetrieval = async (sourceId: string): Promise<RetrievalPayload> => {
+    invokeRetrieval = async (
+        sourceId: string,
+        parseStartDateString?: string,
```
attwad, line 93: nit but as the start and end dates are to be used together I'd have bundled them in a struct to pass them as parameters

verification/curator-service/ui/src/components/AutomatedBackfill.tsx
```diff
@@ -0,0 +1,233 @@
...
+import axios from 'axios';
+import { makeStyles } from '@material-ui/core/styles';
+
+interface RetrievalResult {
+    bucket: string;
+    key: string;
+    upload_id: string;
+}
+
+interface User {
```
attwad, line 21: please use the user interface that's defined in User.tsx

verification/curator-service/ui/src/components/AutomatedBackfill.tsx
```diff
@@ -0,0 +1,233 @@
...
+     * purposes of backfill, users really just want to select a timezone-less
+     * date, as opposed to an instant in time.
+     *
+     * Converting to this format allows a proper range comparison for retrieval
+     * and parsing.
+     */
+    const isoDateString = (date: Date): string => {
+        const offsetMs = date.getTimezoneOffset() * 60 * 1000;
+        const msLocal = date.getTime() - offsetMs;
+        return new Date(msLocal).toISOString().split('T')[0];
```
attwad, line 109: ༼ ༎ຶ ෴ ༎ຶ༽

verification/curator-service/ui/src/components/AutomatedBackfill.tsx
```diff
@@ -0,0 +1,233 @@
...
+            <Formik
+                validationSchema={AutomatedBackfillSchema}
+                validateOnChange={false}
+                initialValues={{
+                    caseReference: undefined,
+                    startDate: null,
+                    endDate: null,
+                }}
+                onSubmit={async (values): Promise<void> => {
+                    await triggerBackfill(values);
```
attwad, line 153: if this returns a promise can you just return triggerBackfill(values)? instead of waiting for it?

attwad (approved):

ingestion/functions/retrieval/valid_scheduled_event.json
attwad: yeah no please ignore I read this before understanding that we weren't storing the parsing date range in the source as we do for the date filter, so it needs to be passed like that.

ingestion/functions/parsing/common/python/parsing_lib/input_event.json
axmb: SG, I've added a check.

verification/curator-service/api/src/clients/aws-lambda-client.ts
axmb: Done.

verification/curator-service/ui/src/components/AutomatedBackfill.tsx
axmb: Neat, thanks.

verification/curator-service/ui/src/components/AutomatedBackfill.tsx
axmb: SG, done.

verification/curator-service/ui/src/components/AutomatedBackfill.tsx
axmb: I know, I know. `Date` is a nightmare -- I promise this is the best, least-left-pad'd way.

PR #1165, opened 2020-09-22
Client-side rate limit geocode requests

### LINKED ISSUE #1131: Ensure we do not go above geocoding rate limits
https://docs.mapbox.com/api/#rate-limits
We can currently do max 600QPS to geocode but do not enforce this on our backend.
Considering we currently geocode things sequentially, I do not see us going over that limit anytime soon but still that'd be good to have.

We only have one task right now so an single rate limiter would be enough (I'm sure node has plenty of rate limiters on npm), I am unsure if the existing library that we're using to talk to mapbox services has somer ate limiting built-in or not.

attwad:
Fixes #1131

Will talk to @gtuli to raise the limits on the mapbox side now that we have that in place.

Note that this assumes 600 requests per second per task, it's not global rate limiting (no shared state between dev/prod tasks), we'll need a more complex solution if we want to support that.

### DIFF
--- a/data-serving/data-service/src/util/validate-env.ts
+++ b/data-serving/data-service/src/util/validate-env.ts
@@ -1,4 +1,4 @@
-import { CleanEnv, cleanEnv, port, str, bool } from 'envalid';
+import { CleanEnv, cleanEnv, port, str, bool, num } from 'envalid';
 
 export default function validateEnv(): Readonly<{
     DB_CONNECTION_STRING: string;
@@ -7,6 +7,7 @@ export default function validateEnv(): Readonly<{
     MAPBOX_PERMANENT_GEOCODE: boolean;
     MAPBOX_TOKEN: string;
     ENABLE_FAKE_GEOCODER: boolean;
+    MAPBOX_GEOCODE_RATE_LIMIT_PER_SEC: number;
 }> &
     CleanEnv & {
         readonly [varName: string]: string | undefined;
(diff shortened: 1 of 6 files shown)


data-serving/data-service/src/util/validate-env.ts
```diff
@@ -37,5 +38,11 @@ export default function validateEnv(): Readonly<{
             devDefault: true,
             default: false,
         }),
+        MAPBOX_GEOCODE_RATE_LIMIT_PER_SEC: num({
+            desc:
+                'number of requests per seconds allowed to mapbox geocode endpoint',
+            devDefault: 50,
```
axmb, line 44: Will we hit this if we test ADI in dev?

axmb (approved):

data-serving/data-service/src/util/validate-env.ts
attwad: devDefault is only used when NODE_ENV is not production so that's for local dev and CI.
dev-curator.ghdsi.org will get 600, as well as prod.

axmb: Of course -- MB, thanks.

gtuli:
@attwad and @axmb 
Just FYI, we had emailed Mapbox yesterday for increasing the geocoding rate limits from 600 requests per minute to 2000 per minute, and also asked if that can be increased not only for the beta testing but permanently for us. 
We are waiting on their response. I will keep you both posted. 
Thanks!

PR #1171, opened 2020-09-22
Flush out the release docs a bit more

allysonjp715:


attwad (approved):
Sorry I made a few changes since we discussed (I tend to just submit to main for small doc changes... my bad for the conflict)

PR #1174, opened 2020-09-22
Add landing page for logged out users

### LINKED ISSUE #1133: Add a landing page for logged out users
Logged out users currently see the following in the Curator Portal, which I would think was broken as a first time user.

image

Even a simple about page with instructions to log in would be good. @shroothless could you design and assign back to eng?

allysonjp715:
Fixes #1133

Also fixes text like "Linelist" -> "Line list" and "Terms of service" -> "Terms of use"

Screen Shot 2020-09-22 at 4 09 37 PM


verification/curator-service/ui/cypress/integration/components/AppTest.spec.ts
```diff
@@ -54,23 +46,26 @@ describe('App', function () {
 
         cy.contains('Create new').should('not.exist');
         cy.contains('Charts').should('not.exist');
-        cy.contains('Linelist').should('not.exist');
+        cy.contains('Line list').should('not.exist');
         cy.contains('Sources').should('not.exist');
         cy.contains('Uploads').should('not.exist');
         cy.contains('Manage users').should('not.exist');
+
+        cy.contains('Trustworthy line list data');
```
attwad, line 54: that's some bold check! :D

verification/curator-service/ui/src/components/App.test.tsx
```diff
@@ -41,31 +41,32 @@ it('renders without crashing when logged out', async () => {
...
     expect(mockedAxios.get).toHaveBeenCalledWith('/auth/profile');
+    expect(queryByTestId('profile-menu')).not.toBeInTheDocument();
 });
 
 it('has link to map', async () => {
     const axiosResponse = {
-        data: {},
-        status: 403,
-        statusText: 'Forbidden',
+        data: { name: 'Alice Smith', email: 'foo@bar.com', roles: ['admin'] },
```
attwad, line 56: why do you need to be logged in to get access to the map?

attwad (approved):

verification/curator-service/ui/src/components/App.test.tsx
allysonjp715: This was specifically checking for this button:
Screen Shot 2020-09-23 at 9 10 04 AM

Which is only visible after logging in. There is another link to the map present on the landing page which is covered in the landing page test.

PR #1179, opened 2020-09-23
"Englishization" of filters

### LINKED ISSUE #1153: Add location to Admin 1, 2 and 3 drop down options
**Describe the bug**
Currently the filter drop down shows Admin 1,2 and 3. Add in location to make it clear to the user the fields are in reference to the location level of the case.
Screen Shot 2020-09-21 at 12 32 08 PM

attwad:
Fixes #1153

image


verification/curator-service/ui/src/components/SearchBar.tsx
```diff
@@ -192,22 +192,28 @@ export default function SearchBar(props: {
...
-                ].map((text) => (
-                    <MenuItem key={text} onClick={(): void => clickItem(text)}>
-                        {text}
+                    { desc: 'curator email', value: 'curator' },
+                    { desc: 'gender', value: 'gender' },
+                    { desc: 'nationality', value: 'nationality' },
+                    { desc: 'occupation', value: 'occupation' },
+                    { desc: 'country', value: 'country' },
+                    { desc: 'outcome', value: 'outcome' },
+                    { desc: 'unique case ID', value: 'caseid' },
```
allysonjp715, line 201: nit: is 'unique' redundant when it's an ID? same for upload ID

allysonjp715 (approved):

verification/curator-service/ui/src/components/SearchBar.tsx
attwad: IDs are not necessarily "globally" unique, but yeah it's technical so not really relevant here, removed.

PR #1192, opened 2020-09-23
Increase node group instance size

axmb:
@attwad: Gaurav mentioned in Slack that it'd be fine to move to `r6g.large` -- thoughts on this resource allocation?

I've drafted this up for 2x`t3.small` --> 1x `r6g.large`, but:

- Two instances is better than one from a redundancy standpoint, but with the current set up we require both instances to meet our resource allocations, anyway.
- We lose 2 total vCPU by doing this -- though I can't tell if we're ever CPU limited.
- I haven't actually used the full 16 GiB of memory.

We could do something like 1x `t3.small` and 1x `t3.large` instead, though I'd have to check if there are consequences to running multiple managed node groups.

### DIFF
--- a/aws/cluster.yaml
+++ b/aws/cluster.yaml
@@ -12,10 +12,10 @@ availabilityZones: ["us-east-1a", "us-east-1b"]
 
 managedNodeGroups:
   - name: managed-worker-ng
-    minSize: 2
-    maxSize: 5
-    desiredCapacity: 2
-    instanceType: t3.small
+    minSize: 1
+    maxSize: 2
+    desiredCapacity: 1
+    instanceType: r6g.large
     labels: { role: worker }
     tags:
       nodegroup-role: worker
--- a/aws/cluster.yaml
+++ b/aws/cluster.yaml
@@ -12,10 +12,10 @@ availabilityZones: ["us-east-1a", "us-east-1b"]
 
 managedNodeGroups:
   - name: managed-worker-ng
-    minSize: 1
+    minSize: 2
     maxSize: 2
-    desiredCapacity: 1
-    instanceType: r6g.large
+    desiredCapacity: 2
+    instanceType: t3.large
     labels: { role: worker }
     tags:
       nodegroup-role: worker
--- a/aws/curator.yaml
+++ b/aws/curator.yaml
@@ -167,10 +167,10 @@ spec:
                   key: email_user_password
           resources:
             requests:
-              memory: "256Mi"
+              memory: "512Mi"
               cpu: "150m"
             limits:
-              memory: "512Mi"
+              memory: "1Gi"
               cpu: "250m"
 ---
 apiVersion: v1
--- a/aws/curator.yaml
+++ b/aws/curator.yaml
@@ -100,7 +100,7 @@ spec:
     matchLabels:
       environment: prod
       app: curator
-  replicas: 1
+  replicas: 2
   template:
     metadata:
       labels:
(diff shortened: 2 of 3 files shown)


attwad:
> * Two instances is better than one from a redundancy standpoint, but with the current set up we require both instances to meet our resource allocations, anyway.
> * We lose 2 total vCPU by doing this -- though I can't tell if we're ever CPU limited.
CPU was never an issue afaict so that lgtm
> * I haven't actually used the full 16 GiB of memory.
> 
> We could do something like 1x `t3.small` and 1x `t3.large` instead, though I'd have to check if there are consequences to running multiple managed node groups.

what would actually be a great improvement with more resource available is to run two instances of the prod service instead of one, this way we can rollout without any downtime, right now there's a few seconds where traffic gets dropped (nginx throws in a 503 I think) while the single task is reloaded that's definitely not great in production.

axmb:
> > * Two instances is better than one from a redundancy standpoint, but with the current set up we require both instances to meet our resource allocations, anyway.
> > * We lose 2 total vCPU by doing this -- though I can't tell if we're ever CPU limited.
> >   CPU was never an issue afaict so that lgtm
> > * I haven't actually used the full 16 GiB of memory.
> > 
> > We could do something like 1x `t3.small` and 1x `t3.large` instead, though I'd have to check if there are consequences to running multiple managed node groups.
> 
> what would actually be a great improvement with more resource available is to run two instances of the prod service instead of one, this way we can rollout without any downtime, right now there's a few seconds where traffic gets dropped (nginx throws in a 503 I think) while the single task is reloaded that's definitely not great in production.

That's a good point. I don't think dropping traffic for that small of a period is a big deal, but it might still be worth it for the ability to not have to carefully time deployments, and to avoid occasionally remediating issues/bugs from dropped traffic, unlikely as they may be.

Adjusted the set up slightly to a configuration that supports this (2x`t3.large` -- there wasn't enough vCPU in 1x `r6g.large`). Seems reasonable, but I'll defer to @SpencerMarks-IDHA and @gtuli for their expertise. For Spencer/Gaurav: the goal is to get more memory headroom (at least a 4Gi limit) for the `data-prod` spec. A nice bonus of this set up is adding a second replica for the prod services, as mentioned above.

gtuli:
I think, the setup with 2 X t3.large is good as it give us both replication and enough compute power per instance (2 vCPUs, 8 GiB). We can also use elastic load balancer for further improvement in the future.  
Are there any limitation of this architecture other than supporting two instances? 
@SpencerMarks-IDHA do you see any limitations here? Shall we get JR's opinion on it?

attwad:
> I think, the setup with 2 X t3.large is good as it give us both replication and enough compute power per instance (2 vCPUs, 8 GiB). We can also use elastic load balancer for further improvement in the future.
> Are there any limitation of this architecture other than supporting two instances?
> @SpencerMarks-IDHA do you see any limitations here? Shall we get JR's opinion on it?

FYI kubernetes already handles load balancing between the nodes (we already have two t3.small nodes that are balanced)

SpencerMarks-IDHA:
I think it is fine.

axmb:
Thanks, appreciate the feedback, all. Please approve when you're ready and I'll update our cluster.

attwad (approved):

gtuli (approved):

axmb:
Submitting so that we can do some testing -- can sync up on those results and talk if any changes are needed.

Issue #1043, opened 2020-09-16
Lookup existing geocodes before sending request to mapbox improvements


attwad:
**Is your feature request related to a problem? Please describe.**
Geocode requests get cached in memory so that we don't call mapbox too much and end up with a large bill, BUT that cache has a fixed size to not eat too much memory (500 items currently) and gets cleared eventually when a service restart happens.

![image](https://user-images.githubusercontent.com/1255432/93347313-3850dc80-f835-11ea-8603-c56456a6969e.png)

**Describe the solution you'd like**
We should store the query that was used for a geocode alongside it in Mongo DB to make it possible to query it again in the future, avoiding mapbox requests whenever possible.

attwad:
Okay so a few things as I am trying to sort things out:

1. this requires moving geocoding to data service which is **non trivial**.
2. this requires moving batch validation into the data service as well as part of batch upsert, which was a TODO in the codebase but is not required because of 1.
3. only then can I start thinking about this issue of reusing existing geocodes, that's going to take a few days for sure.

attwad:
Seems our contract cover us for the next 12 months, so removing the launch blocker bit. Pfew.

gtuli:
Just to clarify, our contract covers total 12 months. We have utilized almost 6 months of it and have six remaining. Having a local lookup engine does make sense long term. We may have weigh it with other pressing requirements before launch.

attwad:
1 and 2 from my earlier comment are done. We can now think about storing the geocode results and re-using them.

abhidg:
Closing as decided to develop separate geocoding API

PR #1196, opened 2020-09-24
store geocode query alongside location

attwad:
For #1043

### DIFF
--- a/verification/curator-service/ui/src/components/new-case-form-fields/LocationForm.tsx
+++ b/verification/curator-service/ui/src/components/new-case-form-fields/LocationForm.tsx
@@ -115,10 +115,15 @@ export function PlacesAutocomplete(
                 let newOptions = [] as Loc[];
 
                 if (results) {
-                    newOptions = [...newOptions, ...results];
-                }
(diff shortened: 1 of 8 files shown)


verification/curator-service/ui/src/components/new-case-form-fields/LocationForm.tsx
```diff
@@ -115,9 +115,13 @@ export function PlacesAutocomplete(
                 let newOptions = [] as Loc[];
 
                 if (results) {
-                    newOptions = [...newOptions, ...results];
+                    newOptions = results.map((l) => {
+                        return {
+                            query: inputValue,
+                            ...l,
```
allysonjp715, line 121: Does this need `...newOptions` too?

allysonjp715 (approved):

verification/curator-service/ui/src/components/new-case-form-fields/LocationForm.tsx
attwad: no it was defined as an empty array before so it's useless I think

Issue #1186, opened 2020-09-23
Ensure data conversion script is up-to-date


allysonjp715:
We're not sure if the script that converts data from spreadsheets to mongo data is completely up to date: https://github.com/globaldothealth/list/tree/main/data-serving/scripts/convert-data

attwad:
Seems like the documents pass the schema validation

```
./data-pipeline/convert_and_import_latest_data.sh

---------------
Running with:

    [-m] MongoDB connection string: mongodb://127.0.0.1:27017
    [-d] database: covid19
    [-c] collection: cases
    [-r] sample rate: 1
    [-s] schema path: /Users/timothe/healthmap/list/data-serving/scripts/data-pipeline/../../data-service/schemas/cases.schema.json
    [-i] indexes path: /Users/timothe/healthmap/list/data-serving/scripts/data-pipeline/../../data-service/schemas/cases.indexes.json
---------------


---------------
Cloning nCoV2019 repo to /tmp/nCoV2019
---------------

Cloning into '/tmp/nCoV2019'...
remote: Enumerating objects: 8, done.
remote: Counting objects: 100% (8/8), done.
remote: Compressing objects: 100% (5/5), done.
remote: Total 4203 (delta 1), reused 8 (delta 1), pack-reused 4195
Receiving objects: 100% (4203/4203), 2.70 GiB | 11.45 MiB/s, done.
Resolving deltas: 100% (1720/1720), done.
Updating files: 100% (1252/1252), done.

---------------
Converting the latest data archive to the MongoDB schema
---------------

Unzipping /tmp/nCoV2019/latest_data/latestdata.tar.gz
Converting 2676404 / 2676404 cases from latestdata.csv
Converting data to new schema and writing to cases.json
[====================================================================================================================================================] 100%
Great success! 🎉

---------------
Preparing the database for the new import
---------------


> setup-db@1.0.0 setup /Users/timothe/healthmap/list/data-serving/scripts/setup-db
> npm install && tsc && mongo $CONN --eval "let args={connectionString: '$CONN', databaseName: '$DB', collectionName: '$COLL', schemaPath: '$SCHEMA', indexesPath: '$INDEXES', deleteAllDocuments: '$DELETE_ALL_DOCUMENTS'}" dist/index.js

audited 152 packages in 0.823s

13 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

MongoDB shell version v4.4.0
[...]

PR #1199, opened 2020-09-24
Various fixes to import scripts

attwad:
For #1186

This gets us closer to being able to import docs AND have curators fix them but we're not there yet as since we made source creation mandatory before case creation can be done this will need to be added to the script (otherwise when editing a case we get errors like `ValidationError: case.caseReference.sourceId: Path `sourceId` is required., case.caseReference: Validation failed: sourceId: Path `sourceId` is required`

Still this is much better.


allysonjp715 (approved):
Thanks!

Issue #1198, opened 2020-09-24
Sources updates not all passing through


z023:
When parser date filtering is set up for a source but this date filtering needs clearing, pressing clear and saving the changes instead sets the date filtering to 'parse all data up to 1 day ago'.

Pressing clear should leave the date filtering column blank.

attwad:
Interesting, taking a look!

attwad:
Must be something with how sources are updated in the backend because the UI sends the right info, but for example parsers cannot also not be cleared (modifications are ignored).

attwad:
Okay this is now fixed in prod.

z023:
I'm still getting the same problem @attwad ??

attwad:
I checked in prod that I could clear the dates, did you save after clicking CLEAR ?

PR #1200, opened 2020-09-24
Fix inability to clear dateFilter fields on source updates

attwad:
It turns out that fields with undefined values are just dropped from the request body by the openapi validator, so we were sending {dateFilter: undefined} and within the source controller updater handler we were seeing req.body == {}

:-(

There doesn't seem to be an option in openapi to keep those fields, there is "nullable" but that just sets the field to null when we save the source document, it doesn't remove it from the document as we want.

So the manual-and-cumbersome alternative that I found was to check and set the field to undefined ourselves if we got an empty {} (non null and non undefined) dateFilter.

Also fixes a typo in the edit case cypress test (how was it working until now??) and better English on the displayed date filter.

For #1198

### DIFF
--- a/verification/curator-service/api/src/controllers/sources.ts
+++ b/verification/curator-service/api/src/controllers/sources.ts
@@ -97,6 +97,12 @@ export default class SourcesController {
                 });
                 return;
             }
+            // Undefined fields are removed from the request body by openapi
+            // validator, if we want to unset the dateFilter we have to pass an
+            // empty object and set it undefined ourselves here.
+            if (JSON.stringify(req.body.dateFilter) === '{}') {
+                req.body.dateFilter = undefined;
+            }
             await source.set(req.body).validate();
             await this.updateAutomationScheduleAwsResources(source);
             const result = await source.save();
--- a/verification/curator-service/ui/src/components/SourceTable.tsx
+++ b/verification/curator-service/ui/src/components/SourceTable.tsx
@@ -175,12 +175,17 @@ class SourceTable extends React.Component<Props, SourceTableState> {
                 this.state.url + oldRowData._id,
                 newSource,
             );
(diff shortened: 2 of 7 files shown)


verification/curator-service/api/src/controllers/sources.ts
```diff
@@ -97,6 +97,12 @@ export default class SourcesController {
                 });
                 return;
             }
+            // Undefined fields are removed from the request body by openapi
+            // validator, if we want to unset the dateFilter we have to pass an
+            // empty object and set it undefined ourselves here.
```
allysonjp715, line 102: :/ should we do this for other fields too?

verification/curator-service/ui/src/components/SourceTable.tsx
```diff
@@ -18,7 +18,6 @@ import ParsersAutocomplete from './ParsersAutocomplete';
 import SourceRetrievalButton from './SourceRetrievalButton';
 import TextField from '@material-ui/core/TextField';
 import axios from 'axios';
-import { isUndefined } from 'util';
```
allysonjp715, line 21: Is this util used elsewhere or can it be removed?

allysonjp715 (approved):

verification/curator-service/api/src/controllers/sources.ts
attwad: This is the only "struct", other singular fields like strings are passed as empty strings and get through as expected.
The automation struct is handled separately.

verification/curator-service/ui/src/components/SourceTable.tsx
attwad: I can't find any other occurences.
Not sure where this was coming from.

image

PR #1217, opened 2020-09-25
Add ingestion functions coverage to CI

attwad:


attwad:
/ragequit, codecov says it can't parse the report but provides no info as to why.

PR #1218, opened 2020-09-25
More recommendations and fixes around flake8 linting for lambda functions

attwad:
Somehow I had it disabled by vscode so I missed a few lints along the way, I now have it enabled and fixed all existing warning/errors.

Would be nice to have it setup the same by the whole team, PTAL, thanks!


AnyaLindstromBattle (approved):
Looks good, will switch to using VSCode and use the recommended settings.

axmb (approved):
Nice, thanks!

PR #1224, opened 2020-09-25
Use codec.getreader to read files

attwad:
This is slightly simpler than what we had before and has the other benefit of working.

Tested with the source `https://covid19.ssp.df.gov.br/resources/dados/dados-abertos.csv?param=[random]`, fetched the result from s3 and running vimdiff over the raw data vs the retrieved data showed no textual difference (just the BOM being removed as the source encoding is indeed utf-8-sig).


AnyaLindstromBattle (approved):
Nice one, looks good, thanks for that!

Issue #482, opened 2020-07-09
Import Germany case data


iamleeg:
https://npgeo-corona-npgeo-de.hub.arcgis.com/datasets/dd4580c810204019a7b8eb3e0b329dd6_0?page=26

z023:
Germany isn't being ingested. A parser issue?

attwad:
The source configuration was pointing to the download page and not the actual data so the parser wasn't parsing anything indeed.

was: https://www.arcgis.com/home/item.html?id=f10774f1c63e40168479a1feb6c7ca74
should be: https://www.arcgis.com/sharing/rest/content/items/f10774f1c63e40168479a1feb6c7ca74/data

attwad:
I've changed it and triggered a retrieval.

attwad:
Looking at the ingestion logs, it seems we have issue with geocoding the cases:

```
Error sending cases to server, status=207, response=
{
    "phase": "GEOCODE",
    "numCreated": 0,
    "numUpdated": 0,
    "errors": [
        {
            "index": 0,
            "message": "no geolocation found for Germany, Schleswig-Holstein, SK Flensburg"
        },
        {
            "index": 1,
            "message": "no geolocation found for Germany, Schleswig-Holstein, SK Flensburg"
        },
...
```

I confirmed in their playground that mapbox prefers location to be written as "SK Flensburg, Schleswig-Holstein, Germany" instead.

@calremmel can you fix this please?

z023:
Is the geo coding resolved? Germany still isn't ingesting.

attwad:
Still some issues with it it seems, let me double check why...

attwad:
Hmmm mapbox has no results for "SK Flensburg, Schleswig-Holstein, Germany" when filtered by district (admin 2), it does provide a good one when no restrictions are set however:

```
{"id":"place.6415657628979890","type":"Feature","place_type":["place"],"relevance":0.772222,"properties":{"wikidata":"Q3798"},"text":"Flensburg","place_name":"Flensburg, Schleswig-Holstein, Germany","bbox":[9.357329,54.75195,9.506826,54.82434],"center":[9.43667,54.78194],"geometry":{"type":"Point","coordinates":[9.43667,54.78194]},"context":[{"id":"region.11132476791043490","wikidata":"Q1194","short_code":"DE-SH","text":"Schleswig-Holstein"},{"id":"country.11437281100480410","wikidata":"Q183","short_code":"de","text":"Germany"}]}
[...]

PR #1225, opened 2020-09-25
Do not restrict Germany geocoding calls to districts

attwad:
Cf. https://github.com/globaldothealth/list/issues/482#issuecomment-698932214


ingestion/functions/parsing/germany/germany.py
```diff
@@ -87,7 +87,10 @@ def parse_cases(raw_data_file: str, source_id: str, source_url: str):
                     "caseReference": {"sourceId": source_id, "sourceUrl": source_url},
                     "location": {
                         "query": ", ".join((row[_ADMIN2], row[_ADMIN1], "Germany")),
-                        "limitToResolution": "Admin2",
+                        # We could set "limitToResolution" to "Admin2" but even
+                        # though inputs are districts mapbox doesn't seem to have
+                        # relevant results when "district" restriction is
+                        # enforced so we do not set limitToResolution on purpose.
```
axmb, line 93: Do you think we should pass `limitToResolution: "Admin2, Admin1, Country"`, or is it better to just elide the field? The impact of the former is avoiding too-fine-grained erroneous geocodes.

attwad: right this feel safer thanks, done.

axmb (approved):

calremmel (approved):

PR #1226, opened 2020-09-25
Get rid of OOM issues in batchUpsert

allysonjp715:
This also starts running schema validations on batchUpsert as `Case.bulkWrite()` did not previously run schema validation. That is the reason there are several peripheral changes in this PR.

Simple tests indicate there may be a speed increase (if nothing else affected the tests):
Before these changes, 1k cases took 29s.
After these changes, 1k cases took 10s.

Before these changes, 10k cases caused OOM most times.
After these changes, 10k cases took 1.6 minutes.

The vast majority of the time (~90%) is spent on the `Case.validate()` calls. If that could be improved there would be large speed improvements overall.


data-serving/data-service/src/controllers/case.ts
```diff
@@ -301,41 +298,35 @@ export class CasesController {
...
-                        return {
-                            insertOne: {
-                                document: c,
-                            },
-                        };
-                    }
-                }),
-                { ordered: false },
-            );
+            // Use this method rather than Case.bulkWrite() as that method
```
axmb, line 301: Any thoughts as to why this method seems to be easier on memory? `bulkWrite` seems to be the recommended API for this (re: [tip](https://docs.mongodb.com/manual/reference/method/db.collection.initializeUnorderedBulkOp/)), but I assume you held everything else constant and witnessed OOMs with bulkWrite vs. this method.

data-serving/data-service/src/controllers/preprocessor.ts
```diff
@@ -16,7 +16,7 @@ const createNewMetadata = (curatorEmail: string): any => {
         revisionNumber: 0,
         creationMetadata: {
             curator: curatorEmail,
-            date: Date.now(),
+            date: new Date().toISOString(),
```
axmb, line 19: Assuming this is related to the OOM'ing -- was moving this around as a `Date` object using a bunch more memory than this approach?

verification/curator-service/ui/src/components/BulkCaseForm.tsx
```diff
@@ -245,8 +246,8 @@ class BulkCaseForm extends React.Component<
                 name: 'hospitalAdmission',
                 dateRange: c.dateHospitalized
                     ? {
-                          start: c.dateHospitalized,
-                          end: c.dateHospitalized,
+                          start: toUTCDate(c.dateHospitalized),
```
axmb, line 249: nit: this is actually more like `toUTCDateString`
hyper-nit: why `toString()` here and `toISOString()` above? I'd personally rather use `toISOString()`

axmb (commented):
Is it possible to try some timing tests here? It won't be totally faithful to prod latencies, but can give a ballpark figure.

I expect latency will remain relatively constant, since:

- I've tested the validation serialization before, and that didn't make much of a difference
- Geocoding should be mostly (I think?) cache hits in prod.
- I can't imagine bulkWrite/initialize[...] are too different.

But it'd be good to be sure.

If you're seeing OOMs below 10k cases, LMK, because I can pull/run the tests myself at 10k on my machine.

data-serving/data-service/src/controllers/case.ts
allysonjp715: Yes everything else held constant it was OOM. The initializeUnorderedBulkOp method breaks it into much smaller groups. It also does schema validation which bulkWrite does not do (which caused the need for all the peripheral changes in this PR) 

From the docs of bulkWrite: 
The number of operations in each group cannot exceed the value of the maxWriteBatchSize of the database. As of MongoDB 3.6, this value is 100,000. This value is shown in the isMaster.maxWriteBatchSize field.

This limit prevents issues with oversized error messages. If a group exceeds this limit, the client driver divides the group into smaller groups with counts less than or equal to the value of the limit. For example, with the maxWriteBatchSize value of 100,000, if the queue consists of 200,000 operations, the driver creates 2 groups, each with 100,000 operations.
https://docs.mongodb.com/manual/reference/method/db.collection.bulkWrite/#execution-of-operations

From the docs of db.collection.initializeUnorderedBulkOp: 
Each group of operations can have at most 1000 operations. If a group exceeds this limit, MongoDB will divide the group into smaller groups of 1000 or less. For example, if the bulk operations list consists of 2000 insert operations, MongoDB creates 2 groups, each with 1000 operations.
https://docs.mongodb.com/manual/reference/method/db.collection.initializeUnorderedBulkOp/#execution-of-operations

data-serving/data-service/src/controllers/preprocessor.ts
allysonjp715: It was a schema validation failure. Schema validation wasn't being run with bulkWrite. It was easier to change this than to change the schema, cause then everywhere it's referenced in the FE would have to change.

verification/curator-service/ui/src/components/BulkCaseForm.tsx
allysonjp715: BTW this is also how we do it in single case entry form.

Essentially, we need to display all the dates in UTC, but no one is concerned with timezone when they are entering the data. So if I type in a date and then use `.toISOString()` it may change the date value because it's assuming I intentionally used my own time zone. So `toUTCDate` changes the date to be in the UTC timezone, but keep the value of the date.

I agree it's unfortunate but this is what we've come up with so far. I suppose you are right on the naming, but it's used in a bunch of places that I'd rather not change, at least not in this PR :)

data-serving/data-service/src/controllers/case.ts
axmb: Got it -- so we're avoiding the second most likely place to OOM, which is the Mongo heap.

axmb (approved):
I ran some local testing -- chatted with Allyson already, mostly resummarizing here for posterity.

First, I noticed we're OOM'ing earlier in the data service (at _main_, as compared to a month ago). I expect we should have tested bulk upload when we reconfigured the interaction between the curator and data service, @attwad -- whereas I'd typically test bulk upload with 10k, and OOM'd only at about 30k cases, I'm now testing with 4k cases to avoid the OOMs.

Second, for the tests I was able to run, this change is at least as fast (or faster, in some cases) as compared to main. The only issue I noticed is if an upsert is done with no effective changes (so uploading all cases with UUIDs, and the case data is identical to the existing stored cases), there's a Mongo error:

```
data_1       | [1601055146657] WARN  (MongoError/40 on 541afd2d1d71): Invalid Operation, no operations specified
data_1       |     MongoError: Invalid Operation, no operations specified
data_1       |         at Function.create (/usr/src/app/data-serving/data-service/node_modules/mongodb/lib/core/error.js:57:12)
data_1       |         at toError (/usr/src/app/data-serving/data-service/node_modules/mongodb/lib/utils.js:123:22)
data_1       |         at UnorderedBulkOperation.bulkExecute (/usr/src/app/data-serving/data-service/node_modules/mongodb/lib/bulk/common.js:1109:31)
data_1       |         at UnorderedBulkOperation.execute (/usr/src/app/data-serving/data-service/node_modules/mongodb/lib/bulk/common.js:1138:22)
data_1       |         at CasesController.<anonymous> (/usr/src/app/data-serving/data-service/src/controllers/case.ts:323:48)
data_1       |         at Generator.next (<anonymous>)
data_1       |         at fulfilled (/usr/src/app/data-serving/data-service/dist/controllers/case.js:24:58)
data_1       |         at processTicksAndRejections (internal/process/task_queues.js:97:5)
curator_1    | [1601055146688] ERROR (40 on 955b267ddd48): Request failed with status code 500
curator_1    |     Error: Request failed with status code 500
curator_1    |         at createError (/usr/src/app/verification/curator-service/api/node_modules/axios/lib/core/createError.js:16:15)
curator_1    |         at settle (/usr/src/app/verification/curator-service/api/node_modules/axios/lib/core/settle.js:17:12)
curator_1    |         at IncomingMessage.handleStreamEnd (/usr/src/app/verification/curator-service/api/node_modules/axios/lib/adapters/http.js:236:11)
curator_1    |         at IncomingMessage.emit (events.js:327:22)
curator_1    |         at IncomingMessage.EventEmitter.emit (domain.js:485:12)
curator_1    |         at endReadableNT (_stream_readable.js:1224:12)
curator_1    |         at processTicksAndRejections (internal/process/task_queues.js:84:21)
```

Which should be a quick fix, @allysonjp715.

data-serving/data-service/src/controllers/case.ts
```diff
@@ -213,15 +213,15 @@ export class CasesController {
...
-            // eslint-disable-next-line @typescript-eslint/no-explicit-any
-            cases.map((c: any, index: number) => {
-                return new Case(c).validate().catch((e) => {
-                    errors.push({ index: index, message: e.message });
-                });
-            }),
-        );
+        // Do not parallelize these requests as it causes an out of memory error
+        // for a large number of cases. However this does take a long time to run
+        // in parallel, so if Mongo creates a batch validate method that should be used here.
```
attwad, line 218: s/in parallel/sequentially? unless I am insunderstanding this comment...

data-serving/data-service/src/controllers/case.ts
```diff
@@ -301,41 +298,35 @@ export class CasesController {
...
+                }
+            });
+            const bulkWriteResult = await bulk.execute();
             res.status(200).json({
                 phase: 'UPSERT',
                 numCreated:
-                    (bulkWriteResult.insertedCount || 0) +
-                    (bulkWriteResult.upsertedCount || 0),
-                numUpdated: bulkWriteResult.modifiedCount,
+                    (bulkWriteResult.nInserted || 0) +
```
attwad, line 327: nit but I really prefer `numInserted` instead of `nInserted`
(yes it comes from a gopher I know but this is not go... ;) )

data-serving/data-service/src/geocoding/adminsFetcher.ts
```diff
@@ -83,6 +83,16 @@ export default class MapboxAdminsFetcher {
                         (await this.getName(feature.properties.id)) ||
                         undefined;
             }
+            // Necessary because the schema doesn't accept undefined values
```
attwad, line 86: we should change the schema to allow undefined values for those, leave a todo if you don't want to do it here?

verification/curator-service/ui/src/components/BulkCaseForm.tsx
attwad: (FYI cmd+shift+f > replace is your friend for multiple files replacements in vscode)

attwad (approved):

data-serving/data-service/src/controllers/case.ts
allysonjp715: yes, good catch, changed.

data-serving/data-service/src/controllers/case.ts
allysonjp715: It's a field on the bulkWriteResult that we can't change :/ it is strange that they change from numInserted to nInserted depending on which bulk method is used.

data-serving/data-service/src/geocoding/adminsFetcher.ts
allysonjp715: Added todo.

allysonjp715:
> Which should be a quick fix, @allysonjp715.

Ah, thanks. Fixed and added a test.

PR #1233, opened 2020-09-28
Geocode every location entries in new case

### LINKED ISSUE #1231: Geocode travel history as well
trel history locations are not geocoded right now, one cannot set the location.query in them and have it geocoded.

We should just geocode them as well.

attwad:
So far we were only looking at `case.location`, now we also look at `case.travelHistory.travel.location`.

Fixes #1231

### DIFF
--- a/data-serving/data-service/test/controllers/case.test.ts
+++ b/data-serving/data-service/test/controllers/case.test.ts
@@ -5,6 +5,7 @@ import { MongoMemoryServer } from 'mongodb-memory-server';
 import app from './../../src/index';
 import fullCase from './../model/data/case.full.json';
 import minimalCase from './../model/data/case.minimal.json';
+import caseMustGeocode from './../model/data/case.mustgeocode.json';
 import mongoose from 'mongoose';
 import request from 'supertest';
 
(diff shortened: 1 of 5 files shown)


data-serving/data-service/test/controllers/case.test.ts
```diff
@@ -492,6 +493,47 @@ describe('POST', () => {
...
+            .expect(200);
+        await request(app)
+            .post('/api/cases')
+            .send({
+                ...caseMustGeocode,
+                ...curatorMetadata,
+            })
+            .expect(201)
+            .expect('Content-Type', /json/);
+        expect(await Case.collection.countDocuments()).toEqual(1);
```
allysonjp715, line 524: Can you check that the created case has the right geocoded fields?

allysonjp715 (approved):

data-serving/data-service/test/controllers/case.test.ts
attwad: yup, done.

Mougk:
@attwad do we need to adjust our parsers for this or should geo-coding on travel history work fine as is?

attwad:
No parser was providing that info until now, Felix is adding the first one right now so nothing needs to be done retroactively.

PR #1234, opened 2020-09-28
Add a catch all redirect to /

### LINKED ISSUE #1205: Redirect to home page or add custom 404 page?
Currently only https://curator.ghdsi.org/ shows the welcome screen when logged out. https://curator.ghdsi.org/cases or anything else shows a blank screen. We should show the welcome screen on any https://curator.ghdsi.org/ URL when logged out.

allysonjp715:
#1205

### DIFF
--- a/verification/curator-service/ui/src/components/App.tsx
+++ b/verification/curator-service/ui/src/components/App.tsx
@@ -735,10 +735,12 @@ export default function App(): JSX.Element {
                                 <LandingPage />
                             )}
                         </Route>
-                        {/* Redirect any unavailable URLs to / */}
-                        <Route path="/">
-                            <Redirect to="/" />
-                        </Route>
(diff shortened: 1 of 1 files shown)


attwad (commented):
Looks okay but shouldn't we redirect to a custom 404 instead when something isn't found?

allysonjp715:
> Looks okay but shouldn't we redirect to a custom 404 instead when something isn't found?

Potentially a product decision? This is easier in the name of time and no UX for a 404 page.

attwad (approved):
okay let's submit this but let's also flag the issue with the product and UX tag and not fix it with this PR.

allysonjp715:
> okay let's submit this but let's also flag the issue with the product and UX tag and not fix it with this PR.

Sounds good will do

PR #1240, opened 2020-09-28
Get search results when clearing search input without having to click…

allysonjp715:
… into the input and hit enter


attwad (approved):
seemed a bit counter intuitive to me (gmail doesn't do it)  but it turns out docs does it... so ok I guess ;)

PR #1261, opened 2020-09-29
bump prod to 0.2.1

attwad:


allysonjp715 (approved):

allysonjp715 (commented):
Actually my PR may be not backwards-compatible if we want to bump a major version: https://github.com/globaldothealth/list/pull/1243

attwad:
oh I see... well if we drop all data anyway were' not "breaking" anything ;) maybe we can push 1.0.0 as a farewell push on Wednesday

Issue #1263, opened 2020-09-29
Basic load testing of curator portal


attwad:
**Is your feature request related to a problem? Please describe.**
Some time ago we gave reader access to every logged-in user, that has load implications that we should measure.

**Describe the solution you'd like**
Use something like locust or some open source load testing tool locally and on dev to measure our capacity.

PR #1264, opened 2020-09-29
authenticate bearer holders in mustBeAuthenticated method

attwad:
I noticed that doing a load test /api/cases was returning 403 and /api/sources was working fine, turns out we do not authenticate token bearer in the `mustBeAuthenticated` middleware as we do in the `mustHaveAnyRole` one. Now they are on par.

Now the load test passes fine locally, I need to push this to dev to be able to load test there.

For #1263

### DIFF
--- a/verification/curator-service/api/src/controllers/auth.ts
+++ b/verification/curator-service/api/src/controllers/auth.ts
@@ -26,8 +26,19 @@ export const mustBeAuthenticated = (
 ): void => {
     if (req.isAuthenticated()) {
         return next();
+    } else {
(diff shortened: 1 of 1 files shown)


allysonjp715 (approved):
Is it possible to add a test for this?

attwad:
> Is it possible to add a test for this?

Nope :( at least not within 2 days in any meaningful way because the only thing I can do "now" is mock passport altogether, at which point only 1% of the logic is tested so it's not a useful test ot have imo.

PR #1266, opened 2020-09-29
Load tests docs and initial results

attwad:
It seems we're CPU bound for the curator service and prod has 250m which limits it to 10QPS (readonly traffic, mostly listing cases) it seems.
We should definitely increase that limit in prod.

For #1263


loadtest/README.md
```diff
@@ -0,0 +1,111 @@
...
+pip3 install -r requirements.txt
+```
+
+Get access to serialized credentials stored in S3 or generate your own and put them in an S3 bucket that you can access, then set the required environment variables when running locust:
+
+```shell
+S3_BUCKET='epid-ingestion' S3_OBJECT='covid-19-map-277002-0943eeb6776b.json'
+```
+
+You can check the [ingestion docs](/ingestion/functions/README.md) for how to generate/get those creds.
```
axmb, line 60: Does it cover generating?

loadtest/README.md
```diff
@@ -0,0 +1,111 @@
+# Load testing
+
+This directory contains [locust](https://locust.io) load testing tasks to load test the Global.health APIs.
+
+Set-up and initial load testing was tracked in [this github issue](https://github.com/globaldothealth/list/issues/1263).
+
+## TL;DR results
+
+Dev starts to cpu-choke at ~10 QPS (simulated with ~20 users and a spawn rate of 1), it's using the allowed 250m vCPUs fully while the RAM usage doesn't increase at all (50Mb->60Mb).
```
axmb, line 9: Does cpu choke basically mean we move from "ballpark reasonable" latency to "totally unreasonable" latency at 10QPS? Since we don't have latency SLOs around which to frame stress test results, I think this description works if that's what you mean.

axmb, line 9: Do we want to try to add an extrapolation to prod in the tl;dr? Or maybe a quick napkin calculation based on vCPUs?

axmb (approved):
Neat.

loadtest/README.md
attwad: I plan on doing the extrapolation with the PR that increases the CPU available to prod yes.

attwad: yes I rephrased, that's what I meant, latency was getting unreasonable when CPU is over the limits.

loadtest/README.md
attwad: it points to the aws docs, which covers generating yes.

loadtest/README.md
```diff
@@ -0,0 +1,111 @@
...
+Data service resource usage at 10QPS did not increase significantly, CPU constraints seem to be on the curator service only.
+
+So basically: read-only traffic doesn't impact RAM usage much, curator service bottlenet is its CPU limits.
+
+## Initial setup
+
+Install Python 3.8, locust and the necessary dependencies with:
+
+```shell
+pip3 install -r requirements.txt
```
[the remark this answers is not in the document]
attwad: good call, done

PR #1267, opened 2020-09-29
Sao Paolo Brazil Parser

calremmel:
Here's a first take for the Sao Paolo parser! Unfortunately, this is not case-level data so it is not possible to dedupe.


ingestion/functions/parsing/saopaolo/saopaolo.py
```diff
@@ -0,0 +1,66 @@
...
+                                "start": when,
+                                "end": when,
+                            },
+                        },
+                    ],
+                }
+                for _ in range(num_confirmed_cases):
+                    yield case
+            except ValueError as ve:
+                print(ve)
```
attwad, line 62: maybe print *and* raise? otherwise that parser could just be parsing nothing and wouldn't fail (we wouldn't notice)

attwad (approved):

PR #1271, opened 2020-09-29
even out cpu limits across curator and data services

attwad:
Fixes #1266

### DIFF
--- a/aws/curator.yaml
+++ b/aws/curator.yaml
@@ -168,10 +168,10 @@ spec:
           resources:
             requests:
               memory: "512Mi"
-              cpu: "150m"
+              cpu: "500m"
             limits:
               memory: "1Gi"
-              cpu: "250m"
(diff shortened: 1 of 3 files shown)


aws/curator.yaml
```diff
@@ -168,10 +168,10 @@ spec:
           resources:
             requests:
               memory: "512Mi"
-              cpu: "150m"
+              cpu: "500m"
             limits:
               memory: "1Gi"
-              cpu: "250m"
+              cpu: "768m"
```
axmb, line 174: Do these new values fit? We're running two replicas of prod services, and we have 4 total vCPU.

axmb (approved):

aws/curator.yaml
attwad: yes they do (prod is reloaded now with these values)

PR #1272, opened 2020-09-29
Add a dateFilter for greater than X days ago

axmb:
For large sources with UUIDs, this is the best way to get recent data. We can't ingest the entire corpus in each run, and using `EQ` runs the risk of missing data (a risk that's necessary/mitigated by delays for data without UUIDs).


attwad (approved):

attwad (approved):
Please fix the display as well in https://github.com/globaldothealth/list/blob/7b30d6305f72aff6c91584479611c77fc6fe27cc/verification/curator-service/ui/src/components/SourceTable.tsx#L411

PR #1275, opened 2020-09-30
Don't disaggregate non-line-list India data

axmb:
Backfilling this data, I noticed that the vast majority isn't actually able to be disaggregated into our platform. While I'd previously assumed the patientNumber/entryID represented a unique case/report, that isn't true for the "bulk" data they include. We should/must avoid such cases (which are represented as rows with >1 numCases) for two reasons:

- Rows with >1 numCases don't contain demographic information (merely location/count). Which is a bummer, though not a deal breaker, however:
- They don't always represent new data; for instance, a state might publish the amount of currently hospitalized patients

AFAICT, for rows with numCases == 1 we can be sure:

- For cases that receive updates, a state patient number is used, which we can use ourselves to dedupe
- Rows without state patient numbers aren't updated

The second point is my best guess based on casually browsing the data. We could be totally sure by _only_ taking data with state patient numbers, but it'd further reduce the amount of usable data from India. @Mougk and @outbreakprepared for thoughts there.

### DIFF
--- a/ingestion/functions/parsing/india/india.py
+++ b/ingestion/functions/parsing/india/india.py
@@ -109,15 +109,33 @@ def parse_cases(raw_data_file: str, source_id: str, source_url: str):
         uuid_column = ""
         uuid_prefix = ""
         for row in reader:
+            # Only rows representing 1 case are line list style data with
+            # demographic information. Additionally, most rows representing
+            # bulk data are actually transposed daily-count data -- in other
+            # words, they represent things like the current number of
+            # hospitalized patients, which we can't disaggregate into our
+            # dataset.
+            if int(row["Num Cases"]) != 1:
+                continue
+
             # The column used to denote case UUID changes in April.
             # It resets back to 1 when this happens.
             # Prefix old values ("Patient Number") to distinguish.
             if not uuid_column:
                 if "Entry_ID" in row:
                     uuid_column = "Entry_ID"
+                    uuid_prefix = ""
                 else:
                     uuid_column = "Patient Number"
                     uuid_prefix = "P"
+
+            # Some states publish updates on patients. In these cases, the
+            # values are keyed on a special state-specific ID. If that's
+            # available, we should use that to dedupe.
+            if row["State Patient Number"]:
+                uuid_column = "State Patient Number"
+                uuid_prefix = ""
+
             case = {
                 "caseReference": {
                     "sourceId": source_id,
@@ -139,10 +157,8 @@ def parse_cases(raw_data_file: str, source_id: str, source_url: str):
                 "notes": row["Notes"] or None
             }
             update_for_status(row, case)
-            for i in range(int(row["Num Cases"])):
-                c = copy.deepcopy(case)
-                c["caseReference"]["sourceEntryId"] = f"{uuid_prefix}{row[uuid_column]}-{i + 1}"
-                yield c
+            case["caseReference"]["sourceEntryId"] = f"{uuid_prefix}{row[uuid_column]}"
+            yield case
 
 
 def lambda_handler(event, context):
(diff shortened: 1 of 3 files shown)


attwad (approved):

Mougk:
AFAICT your suggestion looks good. We should only ingest data where we can be absolutely clear that they are unique and have detailed information, even if that reduces numbers.

Mougk (approved):

beoutbreakprepared:
@axmb  could you share a specific example? I want to confirm interpretation is correct - some spot checks I saw of places where numCases >1, and currentstatus == "Hospitalized" these represent incident new cases in a specific geography. Agreed on the lack of specific details when they do so though. Note - I cannot be exhaustive in my manual spot checks, so would be good to see a specific example that gives you cause for concern

axmb:
> @axmb could you share a specific example? I want to confirm interpretation is correct - some spot checks I saw of places where numCases >1, and currentstatus == "Hospitalized" these represent incident new cases in a specific geography. Agreed on the lack of specific details when they do so though. Note - I cannot be exhaustive in my manual spot checks, so would be good to see a specific example that gives you cause for concern

@beoutbreakprepared -- I'd originally noticed after trying a backfill in dev. For the most recent file (September 22nd onwards), the sum of the "Num Cases" column is 1.7M:

```shell
Desktop
➜  head -n 2 raw_data16.csv
Entry_ID,State Patient Number,Date Announced,Age Bracket,Gender,Detected City,Detected District,Detected State,State code,Num Cases,Current Status,Contracted from which Patient (Suspected),Notes,Source_1,Source_2,Source_3,Nationality,Type of transmission,Status Change Date,Patient Number
297549,,22/09/2020,,,,Aizawl,Mizoram,MZ,107,Hospitalized,,,http://twitter.com/dipr_mizoram/status/1308217294968020993,,,,,,

Desktop
➜  awk -F',' '{sum+=$10;} END{print sum;}' raw_data16.csv
1696713
```

That said, I think your observation re: `Hospitalized` is correct in the general case.

```shell
Desktop
➜  grep "22/09/2020.*Kerala.*Hospitalized" raw_data16.csv | awk -F',' '{sum+=$10;} END{print sum;}'
4125
```

I've checked a few states; and the total for India is pretty close as well: 83.3k for 22-Sep.

So with that in mind, we could:

1. Ingest all cases with `Hospitalized` status, including disaggregating rows with >1 `Num Cases`, and,
2. Ingest cases that have a state patient number, regardless of status, such that we can update known corresponding cases with new outcomes/data.

These days, (1) mostly has no demographic data, but earlier on when counts were lower (e.g. June), some states _did_ post individual records with `Hospitalized` which lacked state patient numbers.

With that in mind, shall we implement ingestion for the above two criteria?

beoutbreakprepared:
@axmb are there hidden endpoints? https://api.covid19india.org/ I can only see up to 14 [22nd August through to date], but you're referencing a 16 file correct? Apologies - i'm really bad at working with json files

axmb:
> @axmb are there hidden endpoints? https://api.covid19india.org/ I can only see up to 14 [22nd August through to date], but you're referencing a 16 file correct? Apologies - i'm really bad at working with json files

No worries -- we're using [the CSV endpoint](https://api.covid19india.org/documentation/csv/) (it's more efficient to parse).

beoutbreakprepared:
@axmb - ok, much easier. So yeah - Hospitalized, Recovered and Deceased are not mutually exclusive [which i think got you on the 1.7m odd. If you just subset to Hopsitalized is approx 800k which is consistent-ish with national cases

axmb:
Cool, I've updated to that specification (PTAL).

z023:
The source url hasn't changed has it? There is still an error with ingestion of the first India CSV I set up.

axmb:
> The source url hasn't changed has it? There is still an error with ingestion of the first India CSV I set up.

Nope, source URL hasn't changed. Let's hold off on backfilling until we get this change in.

Issue #1013, opened 2020-09-14
Add functionality to exclude ID's from auto-ingestion


Mougk:
Some cases (due to their uniqueness or complexity) will be erroneous and need curators to manually change them. Auto-ingestion will overwrite them with current workflow. 

Solution proposed:
Enable easy to configure way for curators to add **ID's** and **fields** that should not be included in auto-ingestion.

attwad:
@Mougk I'm not sure I get the "ID and Fields" combination here, do you mean that curators should be able to say "for these cases IDs, do not ingest these fields" ? or something else? please clarify, thank you.

Mougk:
Yes: "for these case IDs, do not ingest these fields" OR if there has been a manual change to a field by a curator, block updating manually changed fields for specific IDs during automated ingestion.

attwad:
Okay thanks, that's an interesting use-case.

We haven't thought of a way to do that, we don't even know if something's ingested manually or automatically (the automated "ingester" account is jut another user treated as any other curator).

The simplest approach would be to have curators be able to "lock" a whole case (making it read-only basically), that's simple but doesn't cover the field granularity that you requested.

We could also consider curators just leaving that case "unverified" with a note saying "wrongly ingested case, pleaes ignore" or something like that. That's possible with the current system.

The proper "id+fields" match would require writing a specific design document on how to do this as there are quite a few edge cases I believe so it's non trivial.

@Mougk wdyt?

Mougk:
Would the "lock" suggestion allow manual changes from curators on that file? If so, that's great! 

Id+fields would need to wait.

@attwad thoughts? 

@z023 and @outbreakprepared FYI

attwad:
I was thinking the curator would "lock" the case after having done the manual changes that were needed yes.

z023:
But what if there is new information for the case that ingestion would have added? It would be missed if the case was locked?

axmb:
It'd likely be missed, @z023. If the total number of erroneous cases is low, that might be an acceptable consequence.

Alternatively, if the total number of bad cases is quite low, we could special case them in the parser. Because we're avoiding updates for cases specifically with UUIDs, it'd be simple enough to check for those IDs and skip in the parser. This is the simplest solution, IMO, allows (with a reduction in simplicity) to do field-wise skipping or special casing.

attwad:
Yes that's the most straightforward approach, I can send an example for Japan now.

axmb:
[...]

PR #1278, opened 2020-09-30
ignore bad known ids for japan

attwad:
For #1013


axmb (approved):
Looks good, but we should hold off on submitting until Japan is backfilled.

iamleeg:
Closing this out as it's ancient, but feel welcome to add commits to the branch and resubmit if you still think this is important!

PR #1280, opened 2020-09-30
add example parser and better docs

### LINKED ISSUE #1265: Local Dev Environment Documentation Updates
There are a few points in the documentation of the local dev environment where I would find updates or further clarification very helpful for both new contributors and for me personally.

1. There are a couple of new secrets to include in the dev/.env file that it would be great to list with the others
2. For the dev/.env file, it would be great to list points of contact for getting secrets. I think this should probably be me, Spencer, and Gaurav for full coverage?
3. I find the process for adding and testing a new source a little difficult to follow, particularly given that it has changed a little from a couple of months ago. Perhaps we could clarify the steps to the point a total fool like me can follow them?

This could be a good opportunity to go through other documentation updates that might be helpful, so I've assigned anyone who might also regularly engage with the local dev documentation for writer parsers for their input as well.

attwad:
Fixes #1265


ingestion/functions/README.md
```diff
@@ -119,6 +119,114 @@ execution of code.
...
+├── __init__.py      # Required to make this a proper python package, usually empty.
+├── example.py       # Write your parsing code here.
+├── example_test.py  # Always add unit tests.
+├── input_event.json # AWS Cloudwatch event used when testing locally, will be described below.
+├── requirements.txt # Any special third-party dependency that your parser requires, this file is required even if it's empty.
+└── sample_data.csv  # Some sample data used in unit tests, usually copied verbatim from a real source.
+```
+
+At minima, a parser must generate a list of cases that conform to the openAPI
+specifications. If you have a local stack running, go to the [OpenAPI UI](http://localhost:3001/api-docs) to check the structure of a `Case` object. Otherwise you can always [check it online](https://curator.ghdsi.org/api-docs/) as well.
```
axmb, line 139: nit: can you set line breaks at 80 characters?

ingestion/functions/README.md
```diff
@@ -119,6 +119,114 @@ execution of code.
...
+
+FTR a minimal case can look like this:
```
axmb, line 141: This is documentation -- it's de facto for the record. :-P

ingestion/functions/README.md
```diff
@@ -119,6 +119,114 @@ execution of code.
...
+   "query": "Quebec, Canada",
+   "limitToResolution": "Admin1,Country"
+}
+```
+
+Travel history locations can be geocoded in the same way, travel history should only contain travels prior to the case's location.
+
+Prefer sending queries that go from smallest to biggest regions as mapbox can get confused by the former and geocode to a totally different country that the one you were expecting.
+
+Fields and nested structs should be referably not set (or set to `None`) rather than set to an empty value (for example unknown age shouldn't be set to `''` and unknown demographics altogether shouldn't be set to `{}`).
```
axmb, line 205: sp: preferably

axmb (approved):
LGTM, but certainly wait for Chris to TAL!

ingestion/functions/README.md
attwad: as discussed, not easily in vscode it's just sugar display but not actually introducing newlines upon formatting:(

Mougk:
Also adding @iamleeg to review

calremmel:
This is such a big improvement, thanks so much for doing this!

calremmel (approved):

PR #1285, opened 2020-09-30
fix peru geocoding by going from smaller to bigger regions

attwad:
Before:  https://curator.ghdsi.org/api/geocode/suggest?q=ASILLO,%20AZANGARO,%20PUNO,%20Peru

After: https://curator.ghdsi.org/api/geocode/suggest?q=PUNO,%20AZANGARO,%20ASILLO,%20Peru


allysonjp715 (approved):
Could you add documentation somewhere that this is suggested to do for all parsers?

attwad:
> Could you add documentation somewhere that this is suggested to do for all parsers?

Good call, I have a ongoing PR for parsers docs, I'll add it there.

PR #1298, opened 2020-09-30
Increase memory size for Argentina and Peru parsers

axmb:
```
2020-09-30T18:10:40.015-04:00 | Retrieving raw data from s3://[...]/content.csv
2020-09-30T18:10:53.456-04:00 | [Errno 28] No space left on device
```


axmb:
Dialing this back to 1024 memory size. The error above is actually disk space -- a result of AWS sharing `/tmp/` folders between Lambda executions.

attwad (approved):

PR #1311, opened 2020-10-02
add catch for if location code not in dictionaries

calremmel:
Now if a location code doesn't exist in the dictionary, the query string will simply contain 'MÉXICO'

I also changed the order of the query string to more closely match the way addressed are actually used in the country.

### DIFF
--- a/ingestion/functions/parsing/mexico/mexico.py
+++ b/ingestion/functions/parsing/mexico/mexico.py
@@ -32,15 +32,16 @@ _MUNICIPALITIES = maps["municipalities"]
 
 def convert_location(state_code: str, municipality_code: str):
     """
-    Codes beginning with 9 denote missing information
+    Convert state and municipality codes into location query.
     """
     query_list = []
-    if municipality_code[0] != "9":
+    missing_value_prefix = "9"
+    if municipality_code[0] != missing_value_prefix:
         try:
             query_list.append(_MUNICIPALITIES[state_code + municipality_code])
         except KeyError:
             print(f"Municipality code missing: {municipality_code}")
-    if state_code[0] != "9":
+    if state_code[0] != missing_value_prefix:
         try:
             query_list.append(_STATES[state_code])
         except KeyError:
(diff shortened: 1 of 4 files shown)


attwad:
Is this list so unmaintainable that we have to ignore unknown locations? How about running the parser locally, logging all missing locations and fixing them in the json dictionary? Seems like this will lead to better data but I haven't looked at the dataset so will leave the final decision to you 👍

Mougk:
@calremmel how many locations could not be found for Mexico? Agree with @attwad that adding these locations in lookup table would be best option.

calremmel:
@Mougk @attwad Looks like there are five missing locations, none of which correspond to anything that quickly comes up on Google:

09106
17056
17063
19066
17053

I'm currently downloading the 2019 shapefile from INEGI, which is enormous and will take 40 minutes from now, haha. If these codes are anywhere, they'll be in there.

If not, the state codes (first two digits) are all perfectly valid, so in those cases perhaps we can go an admin level up in the worst case?

Mougk:
very clever. going one level up sounds reasonable. thanks for looking into this @calremmel

attwad:
+1 to using the level up as a fallback!

calremmel:
Ok! I checked, and these appear to be entry errors. I'll have the parser fall back to the state level when this happens. Thanks all!

ingestion/functions/parsing/mexico/mexico.py
```diff
@@ -18,26 +18,34 @@
...
-_STATES = maps['states']
+_STATES = maps["states"]
 
-_MUNICIPALITIES = maps['municipalities']
+_MUNICIPALITIES = maps["municipalities"]
 
 
 def convert_location(state_code: str, municipality_code: str):
     """
     Codes beginning with 9 denote missing information
```
iamleeg, line 35: Documenting this is good, but the docstring for the function shouldn't (just) be this. Things I might do:

 - fill out the docstring, so that this message is a note at the end of a description of the function's behaviour
 - reorganise the code to make the intent clearer:
```python
missing_data_prefix = "9"
if !municipality_code.startswith(missing_data_prefix):
```

iamleeg (approved):

axmb (approved):

iamleeg (approved):

PR #1313, opened 2020-10-06
Handle a number of remaining edge cases in India's data set

axmb:
Some of the early CSV data is pretty loosely formatted. This cleans up a couple odds-and-ends, and allows the first set (`raw_data1.csv`) to be ingested without error. Each commit title tl;dr's the issue being addressed, but to expand slightly:

- I'd cleverly thought to avoid conflating an updated case's initial confirmation date with the date on which an outcome is reported by omitting a `confirmed` event for updates. As it turns out, this won't work because the presence of a `confirmed` event is required, and despite `events` being an array, we're rewriting (rather than appending to) the array with batchUpsert.
- Our schema specifies that additional sources must be unique. Which makes sense, but for whatever reason some rows have duplicate sources listed within them.
- A few rows have age ranges listed.

### DIFF
--- a/ingestion/functions/parsing/india/india.py
+++ b/ingestion/functions/parsing/india/india.py
@@ -63,12 +63,27 @@ def convert_demographics(row):
 
     if row["Age Bracket"]:
         raw = row["Age Bracket"]
-        age = float(raw.split(" ", 1)[
-                    0]) / 12 if " months" in raw.lower() else float(row["Age Bracket"])
-        demo["ageRange"] = {
-            "start": age,
-            "end": age
-        }
+        # Handle ranges, e.g. "28-35"
+        if "-" in raw:
+            parts = raw.split("-", 1)
+            demo["ageRange"] = {
+                "start": float(parts[0]),
+                "end": float(parts[1])
+            }
+        # Handle months, e.g. "6 months"
+        elif " months" in raw.lower():
+            age = float(raw.split(" ", 1)[0]) / 12
+            demo["ageRange"] = {
+                "start": age,
+                "end": age
+            }
+        # Handle standard ages
+        else:
+            age = float(raw)
+            demo["ageRange"] = {
+                "start": age,
+                "end": age
+            }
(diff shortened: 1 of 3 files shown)


ingestion/functions/parsing/india/india.py
```diff
@@ -78,63 +93,47 @@ def convert_demographics(row):
...
+            "name": "hospitalAdmission",
             "dateRange":
             {
                 "start": convert_date(row["Date Announced"]),
                 "end": convert_date(row["Date Announced"])
             },
-            "value": "Recovered"
+            "value": "Yes"
         })
-    elif row["Current Status"] == "Deceased":
```
iamleeg, line 98: Super unimportant, was the change from 'elif' to 'if' important? I can't see that it makes a difference, and certainly I'm not suggesting it needs changing back, just interested if this was a motivated or autopilot change. You still have elif down in l127.

iamleeg (approved):

ingestion/functions/parsing/india/india.py
axmb: Just an autopilot mishap, apologies. These are mutually exclusive conditions, so despite the functional lack of difference, my intention was to use `elif`.

z023:
The first India CSV has ingested however looking at the data compared with the source data there are more cases in the line list than there should be and some recovered and death outcomes are coming through when it should just be hospitalised. So I think some cases are being duplicated. Here's a comparison of the numbers:
Screenshot 2020-10-08 at 11 56 04

axmb:
That's odd -- I'll take a look this week. I tested it locally, and ended up ingesting ~17.2k, IIRC.

The deceased/recovered statuses are expected; we _do_ ingest those cases if there's a state patient number provided. IIUC, the cases with state patient numbers typically share those state UUIDs between the outcome row (recovered/deceased) and the confirmation row (hospitalized), which is why we count them. If that's not the case, we can always cut those out.

PR #1314, opened 2020-10-07
Document error recovery steps

iamleeg:


AnyaLindstromBattle (commented):
I'm wondering if you should maybe include every single step we did, just for noobs like me. i.e. running 'docker-compose -f docker-compose.yml -f docker-compose.dev.yml exec mongo mongo covid19' from dev folder (with or without winpty prefix), then, in mongo shell, 'var email='your@email';' followed by 'var roles=['admin', 'curator'];' and finally 'db.users.updateOne({email: email}, {$set: {roles: roles}});'

iamleeg:
Good suggestion, thanks, I'll make that change.

PR #1315, opened 2020-10-07
Minor changes to Amapa parser

AnyaLindstromBattle:
Changes to date function to deal with empty fields and minor changes for readability and/or consistency with other parsers

### DIFF
--- a/ingestion/functions/parsing/brazil_amapa/amapa.py
+++ b/ingestion/functions/parsing/brazil_amapa/amapa.py
@@ -32,7 +32,7 @@ _COMORBIDITIES_MAP = {
     "Gestante": "pregnancy",
     "Doenças respiratórias crônicas descompensadas": "respiratory system disease",
     "Doenças renais crônicas em estágio avançado (graus 3, 4 ou 5)": "chronic kidney disease",
-    "Doenças cardí­acas crônicas": "heart disease"
+    "Doenças cardíacas crônicas": "heart disease"
 }
 
 
--- a/ingestion/functions/parsing/brazil_amapa/amapa.py
+++ b/ingestion/functions/parsing/brazil_amapa/amapa.py
@@ -15,24 +15,24 @@ except ImportError:
             'common/python'))
     import parsing_lib
 
-_DATE_CONFIRMED_INDEX = 0
-_HEALTHCARE_WORKER_INDEX = 1
-_PREEXISTING_CONDITIONS_INDEX = 2
-_METHOD_CONFIRMATION_INDEX = 3
-_CONFIRMATION_INDEX = 4
-_GENDER_INDEX = 5
-_NOTES_BAIRRO_INDEX = 6
-_MUNICIPALITY_INDEX = 7
-_AGE_INDEX = 8
-_ETHNICITY_INDEX = 9
-_NOTES_INDIGENOUS_GROUP_INDEX = 10
+_DATE_CONFIRMED = "dataNotificacao"
+_HEALTHCARE_WORKER = "profissionalSaude"
+_PREEXISTING_CONDITIONS = "comorbidades"
+_METHOD_CONFIRMATION = "tipoTeste"
+_CONFIRMATION = "resultadoTeste"
+_GENDER = "sexo"
+_NOTES_BAIRRO = "bairro"
+_MUNICIPALITY = "municipio"
+_AGE = "idade"
+_ETHNICITY = "racaCor"
+_NOTES_INDIGENOUS_GROUP = "etniaIndigena"
 
-commorbidities = {
+_COMORBIDITIES_MAP = {
     "Diabetes": "diabetes mellitus",
     "Gestante": "pregnancy",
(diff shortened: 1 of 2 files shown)


iamleeg:
Looks like the test failure is due to the preexisting conditions list.

AnyaLindstromBattle:
> Looks like the test failure is due to the preexisting conditions list.

yeah I always get encoding issues when there are accents involved - think it's due to Excel weirdness

ingestion/functions/parsing/brazil_amapa/amapa.py
```diff
@@ -15,19 +15,19 @@
...
+_METHOD_CONFIRMATION = "tipoTeste"
+_CONFIRMATION = "resultadoTeste"
+_GENDER = "sexo"
+_NOTES_BAIRRO = "bairro"
+_MUNICIPALITY = "municipio"
+_AGE = "idade"
+_ETHNICITY = "racaCor"
+_NOTES_INDIGENOUS_GROUP = "etniaIndigena"
+
+_COMORBIDITIES_MAP = {
```
iamleeg, line 30: This is a great change, it will be more reliable if the columns get reordered in the data.

iamleeg (approved):
Looks good!

ingestion/functions/parsing/brazil_amapa/amapa.py
AnyaLindstromBattle: yep, exactly

calremmel (approved):
Great changes - this should resolve the error I'm currently seeing in the log. Thanks, Anya!

z023:
@AnyaLindstromBattle Amapa is still showing as an error. Here is the latest upload ID 5f8c2f84b5daec0b585d2621

calremmel:
Looking at the logs, the error is being thrown from parsing_lib.py rather from the parser itself:

```
[ERROR] KeyError: 'start'
  File "/var/task/amapa.py", line 180, in lambda_handler
    return parsing_lib.run_lambda(event, context, parse_cases)
  File "/opt/python/parsing_lib.py", line 307, in run_lambda
    common_lib.complete_with_error(
  File "/opt/python/common_lib.py", line 98, in complete_with_error
    raise exception
  File "/opt/python/parsing_lib.py", line 292, in run_lambda
    count_created, count_updated = write_to_server(
  File "/opt/python/parsing_lib.py", line 120, in write_to_server
    batch = batch_of(cases, cases_batch_size)
  File "/opt/python/parsing_lib.py", line 104, in batch_of
    batch.append(next(cases))
  File "/opt/python/parsing_lib.py", line 230, in <genexpr>
    return (case for case in case_data if case_is_within_range(case, cutoff_date, op))
  File "/opt/python/parsing_lib.py", line 216, in case_is_within_range
    case_date = get_case_date(confirmed_event["dateRange"]["start"])
```

z023:
There's still a problem with Amapa. I triggered a retrieval but there are still no uploads or errors in the parser logs.

iamleeg:
@z023 According to the logs, it successfully imported 0 events. Apparently it found no new cases within the last three days. Looking at the raw data the latest case is from 2020-07-20: have they changed the location of their later data, or just stopped updating the report, maybe?

AnyaLindstromBattle:
Yes I was just looking at the data and the issue is, as Graham says, with the data not the parser. 

I did, however, notice another but separate issue. If I download the data from https://drive.google.com/uc?export=download&id=16JBgPV2Px93iD_NbuDo13b3ye4KGxlcW which is the link we are using the last data is from 2020-07-20 as you said, but if I go to the Amapa source website http://painel.corona.ap.gov.br/ and to Baixar CSV to download the document from the Google Drive, the data looks different. The last date is now 29/09/2020 (so we would still get 0 new events as above as this is still more than 3 days ago), and the column titles are different e.g. Data de Notificacao as opposed to dataNotificacao.

I assume we want to get the data that we are missing between July and September into the database. So for this we will need to: 
1. Get a new link to the more up-to-date data - I will sort this and send it to @z023 when I have it. @iamleeg I don't suppose you have any idea why the google drive link that we provided for ingestion didn't 'keep track' of the added data? I just want to make sure that the link I provide isn't a 'snapshot' of the data at that particular time, if that makes sense. 
2. Change the parser to deal with the new format - this should be quick as it's mostly a column name issue
3. Once that's done and merged, we will have to wait for backfilling to work to get the Amapa data ingested. Once that's done, and assuming the new link keeps up with any data added, this should hopefully work as an automated source. 

Sound good?

iamleeg:
If they changed the Google Drive document we would always have play catch-up with that, given the current architecture of the system (otherwise we'd have to make a virtual browser to visit the source website, follow the download link, then get that URL). What you suggest is a good plan.

AnyaLindstromBattle:
Ok cool. So what happens if they update the document again? Do we need to build in some kind of check every few months to manually go and see if the data has been updated and, if it has, update the URL?

iamleeg:
Theoretically we already have that, from the liveness chart on the curator dashboard.

PR #1316, opened 2020-10-07
Goias (Brazil) parser

AnyaLindstromBattle:
First look at Goias parser, feedback wellcome.


ingestion/functions/parsing/brazil_goias/goias.py
```diff
@@ -0,0 +1,189 @@
...
+_DIABETES = "diabetes"
+_LUNG = "doenca_respiratoria"
+_IMMUNOSUPPRESSED = "imunossupressao"
+_CARDIOVASCULAR = "doenca_cardiovascular"
+
+_COMORBIDITIES_MAP = {
+    "doenca_respiratoria": "lung disease",
+    "doenca_cardiovascular": "cardiovascular system disease",
+    "diabetes": "diabetes mellitus"
+}
```
iamleeg, line 34: I think a lot of these lookups, and the functions that use them, are the same in other Brazil parsers (and maybe Portugal if/when we have one). It'd be good to move them to a common file in `ingestion/functions/parsing/common` so that there's a single place to go to read and update them.

iamleeg (changes requested):

ingestion/functions/parsing/brazil_goias/goias.py
```diff
@@ -0,0 +1,189 @@
...
+    if age:
+        if age == ">= 80 anos":
+            demo["ageRange"] = {"start": 80, "end": 120}
+        # No age resolution provided below 10 years
+        elif age == "< 10 anos":
+            demo["ageRange"] = {"start": 0, "end": 9}
+        else:
+            # Age in format '20 a 29 anos'
+            age_range = age.partition(" a ")
+            demo["ageRange"] = {"start": float(age_range[0]), "end": float("".join([i for i in age_range[2] if not i.isalpha()]))}
```
calremmel, line 132: For `age_range[2]`, is that ever not two characters? If it is always two characters, would `float(age_range[2][:2])` do the trick?

I love the use of partition, which I had not heard of before (had only used split).

calremmel (changes requested):

ingestion/functions/parsing/brazil_goias/goias.py
AnyaLindstromBattle: Nice one, so if I understand correctly age_range[2][:2] takes only the first two characters of e.g. '29 anos' ?

ingestion/functions/parsing/brazil_goias/goias.py
AnyaLindstromBattle: Do you mean the comorbidities or the column headers? Or both?

ingestion/functions/parsing/brazil_goias/goias.py
calremmel: Yes that's what it should do!

ingestion/functions/parsing/brazil_goias/goias.py
iamleeg: Everything, really; there's a "words used in Portuguese health reports" library waiting to be extracted here :).

AnyaLindstromBattle: Sure. Shall we merge this parser and then I can go off and do that as it will involve me changing all the other parsers anyway? If you're happy with the rest of it that is.

PR #1319, opened 2020-10-08
fix RGDS Brazil date parsing format to take 4 digit year

calremmel:
RGDS parser was not completing due to the year format not matching. This appears to be a recent change on the source side, so good to note that there have been updates!


iamleeg:
Have they changed all the historical data or is this a change for new rows?

calremmel:
> Have they changed all the historical data or is this a change for new rows?

They have changed all historical data as well.

iamleeg (approved):
All sounds good then!

PR #1321, opened 2020-10-12
Distrito federal

AnyaLindstromBattle:
Edits to logic when adding comorbidities. It was setting preexistingConditions = True even if no comorbidities were added.


iamleeg (approved):
Looks good!

Issue #1318, opened 2020-10-08
Colombia parser timing out


calremmel:
The Colombia parser is continuing to time out during daily ingestion runs.

Other parsers (like Argentina) that have previously timed out have been enabled to complete by reducing the range of the date filter from 7 to 3. We've gone down to 2 in Colombia's case, and it is still not able to run.

Any sense for the best way forward here? Would it make sense to ingest a single day at a time, or perhaps increase the time limit again, or is there some other superior solution?

z023:
I've just triggered Colombia again to parse data after 1 day ago. If that doesn't work I'm stuck for ideas.

axmb:
Looking at the logs for the function -- we attempt to filter the cases by date, and there's no more progress reported after that. Later on in logs, we can see the function's consumed the entirety of its allocated memory.

```

  | 2020-10-13T22:01:18.473-04:00 | Raw file retrieved at /tmp/tmpavsor2_0
  | 2020-10-13T22:01:18.473-04:00 | Filtering cases using date filter {'numDaysBeforeToday': 1, 'op': 'GT'}
  | 2020-10-13T22:15:56.918-04:00 | END RequestId: [...]
  | 2020-10-13T22:15:56.918-04:00 | REPORT RequestId: [...] Duration: 900084.76 ms Billed Duration: 900000 ms Memory Size: 128 MB Max Memory Used: 128 MB Init Duration: 1054.26 ms
```

I'll bump the memory allocation a bit. IIRC, we purposefully don't load the entirety of CSV data into memory, but this seems like a good first step regardless.

calremmel:
Thanks so much for giving this a look! It does appear to have solved the issue, with 308 MB used in the end...

...though now the most recent cases in the dataset were posted on 10-02-2020, so it's still not uploading cases for that reason! That bears some investigation for the source - hopefully their update schedule is just in chunks of dates.

No longer timing out, so I think we're good here. Thanks all!

PR #1324, opened 2020-10-15
Increase memory allocation for Colombia parser.

axmb:
Re: #1318.


iamleeg (approved):
PR too long, can it be broken up?

Just kidding, this is fine.

PR #1326, opened 2020-10-20
Describe using AWS logs in documentation

iamleeg:
I added documentation to the parser README on using the GHDSI portal together with CloudWatch logs to find the cause of a failure in a parser.


calremmel (approved):
I think this looks good. I've been using the online console, but that's fairly self-explanatory and the workflow is the same.

PR #1330, opened 2020-10-22
increase Mexico parser memory allocation

calremmel:
Parser is running out of memory, hopefully this should do the trick.


iamleeg (approved):
Default is 128MB, so this should be quite a bit of overhead!

PR #1331, opened 2020-10-22
Allow both date filter and date range, only using range

iamleeg:
@z023 reported that backfill hasn't worked since Oct 2. What I found is that when you request a backfill, you set the date range on the retrieval request (which is fed through to the parser). The parser lib loads the definition for the source, _including its date filter_. But this means that the parser lib tries to use _both_ the range and the filter when limiting the dates of cases, and boom. Backfill always errors.

I could have fixed this by removing the filter if the range is specified, but instead I chose to relax the "one or the other" assumption and instead document that the range always trumps the filter. That's because the range has always been explicitly requested by the user for this specific event, so should always be treated as deliberate.


AnyaLindstromBattle (approved):
Nice catch.

axmb (approved):
Makes sense, thanks Graham.

The mutual exclusivity invariant was added mid-PR-review, and this (important) case definitely slipped our minds at the time.

PR #1334, opened 2020-10-28
Changes to Paraiba to hopefully solve error in ingestion logs

AnyaLindstromBattle:
I have implemented the same fix as for Amapa to hopefully overcome issue with missing date of confirmation.


iamleeg (approved):
This will "unstick" the parser if there are dates in the wrong format, at the cost of skipping records where that's true. It'd be great to follow up on this work by looking into _why_ some of the dates can't be parsed, and making the parser more tolerant.

AnyaLindstromBattle:
From what I could see in the data the only issue was empty entries, all the entries where there was a date it was in the same format.

iamleeg:
Sounds good then!

PR #1335, opened 2020-10-29
Deal with non-unique additional sources entries

AnyaLindstromBattle:
Hopefully solves issue with validation failing


iamleeg (approved):
It'd be great to add a unit test that demonstrates that this _does_ actually uniquify the sources, as the code trick used is a bit tricky to understand.

AnyaLindstromBattle:
Sure, will do that now. What I tend to do is just edit the sample_data.csv and check that it still passes as expected. But I agree that it would be better to do it separately. I could either create a separate unit test as you did for Amapa, or just add an additional entry into sample_data so that the existing unit test parses two cases (one of which needs uniquifying). Any preference?

iamleeg:
Either of the above is fine :)

PR #1337, opened 2020-10-29
Add a test case where additional sources need to be made unique

AnyaLindstromBattle:
Added a test case where uniqueness of additional sources has to be enforced by the parser. Also updated unit test format to match what is being used in the newer parsers.


iamleeg (approved):
Thanks for this, as a slight workflow enhancement because there are two separate "fixes" here (the test, and supporting the extra fields) it would be preferable to have two PRs to make it easier to review the different changes separately.

PR #1339, opened 2020-10-30
Espirito Santo (Brazil) parser

AnyaLindstromBattle:
My first take on the Espirito Santo parser. Main thing to note is that, although yes/no information regarding national and international travel is provided, because of lack of information regarding how many days prior to confirmation the travel occurred I had to put it in notes (i.e. we cannot be sure that the travel was 30 days prior to confirmation as is expected in the schema).


ingestion/functions/parsing/brazil_espirito_santo/espirito_santo_test.py
```diff
@@ -0,0 +1,56 @@
...
+        sample_data_file = os.path.join(current_dir, "sample_data.csv")
+
+        result = espirito_santo.parse_cases(sample_data_file, _SOURCE_ID, _SOURCE_URL)
+        self.assertCountEqual(list(result), [
+            {
+                "caseReference": {"sourceId": _SOURCE_ID, "sourceUrl": _SOURCE_URL},
+                "location": {
+                    "query": "SERRA, Espirito Santo, Brazil"
+                },
+                "events": [{
```
calremmel, line 26: It might be nice to use as a test case someone who either died or recovered, just to test more of the logic.

ingestion/functions/parsing/brazil_espirito_santo/espirito_santo.py
```diff
@@ -0,0 +1,326 @@
...
+    "5ª à 8ª série incompleta do EF (antigo ginásio ou 1º grau)": "Years 5 to 8 of elementary school not completed",
+    "Ensino médio incompleto (antigo colegial ou 2º grau )": "High school not completed",
+    "Educação superior incompleta ": "Higher education not completed",
+    "Educação superior completa": "Higher education completed",
+    "Analfabeto": "Analphabet",
+    "4ª série completa do EF (antigo primário ou 1º grau)": "Fourth year of elementary school completed"
+}
+
+
+_NONE_TYPES = set(["Não", "Ignorado", "-", "Não Informado", "Não se aplica", None])
```
calremmel, line 76: I quite like this solution. Very readable lower down. Let's keep our eyes out for ramifications around including None in that list - if it shows up in an unexpected place, the behavior can be pretty weird.

ingestion/functions/parsing/brazil_espirito_santo/espirito_santo.py
```diff
@@ -0,0 +1,326 @@
...
+    demo = {}
+    if gender:
+        demo["gender"] = convert_gender(gender)
+    if age:
+        if age == "90 anos ou mais":
+            demo["ageRange"] = {"start": 90, "end": 120}
+        else:
+            # Age in format '20 a 29 anos', except for '05 a 9 anos' which has a leading 0
+            age_range = age.partition(" a ")
+            if age_range[0] == "05":
```
calremmel, line 232: Is this extra case necessary? I believe `float("05")` should return `5.0`. Python tends to automatically handle leading zeroes - this is quite annoying in the case of zip codes, but useful here.

iamleeg: I agree with @calremmel here. Write a unit test for the age parsing, then remove this code and check whether the test still passes :).

ingestion/functions/parsing/brazil_espirito_santo/espirito_santo.py
```diff
@@ -0,0 +1,326 @@
...
+        date = datetime.strptime(raw_date, "%Y-%m-%d")
+        return date.strftime("%m/%d/%YZ")
+
+
+def convert_gender(raw_gender: str):
+    if raw_gender == "M":
+        return "Male"
+    if raw_gender == "F":
+        return "Female"
+
```
iamleeg, line 103: I think we've really reached the point where functions like this should be in a common library, as multiple parsers all do this.

ingestion/functions/parsing/brazil_espirito_santo/espirito_santo.py
AnyaLindstromBattle: So initially I didn't have a specific case for the 05 and it did give me an error like "Python cannot deal with leading 0" when I was doing unit testing, hence why I added the case... Will try and reproduce and post here

ingestion/functions/parsing/brazil_espirito_santo/espirito_santo.py
```diff
@@ -0,0 +1,326 @@
...
+    if ethnicity == "Preta":
+        return "Black"
+    elif ethnicity == "Parda":
+        return "Mixed"
+    elif ethnicity == "Amarela":
+        return "Asian"
+    elif ethnicity == "Branca":
+        return "White"
+    elif ethnicity == "Indigena":
+        return "Indigenous"
```
iamleeg, line 219: @Mougk is "indigenous" expected here? I would have thought that we'd need more context (the indigenous Brazilians being different people from the indigenous Ainu in Japan, for example).

ingestion/functions/parsing/brazil_espirito_santo/espirito_santo.py
AnyaLindstromBattle: It looks as though there is a difference between Python 2 and Python in whether they accept leading 0's https://medium.com/techmacademy/leading-zeros-in-python-a-hidden-mystery-revealed-ee3e3065634d. Anyway will try and reproduce error, may shed some light

iamleeg:
```
>>> float("05 a 9 anos".partition(" a ")[0])
5.0
```
should be OK

iamleeg: @AnyaLindstromBattle ah yeah, there's a convention from C that integers with a leading 0 are treated as base 8 (the reason has something to do with a specific model of computers from the 1970s that isn't very exciting). Converting '09' into an int would fail with python 2, but not 3 (and this is all moot for the float type).

ingestion/functions/parsing/brazil_espirito_santo/espirito_santo.py
AnyaLindstromBattle: Yes I agree, it's next on my list of things I promise

ingestion/functions/parsing/brazil_espirito_santo/espirito_santo_test.py
AnyaLindstromBattle: Thanks, done

AnyaLindstromBattle:
@calremmel I have implemented your comments, thanks!

@iamleeg I am currently working on the library of common functions. I think we should maybe still merge this as I will have to go back and change all the parsers anyway and will do that in a separate PR. 

@Mougk thoughts on Grahams comment above?

Mougk:
Did you find a data dictionary describing when indigenous is used? See for example a full list of indigenous groups here: https://en.wikipedia.org/wiki/List_of_indigenous_peoples_of_Brazil

AnyaLindstromBattle:
@Mougk no unfortunately this data doesn't have a good dictionary that I could find. I have based those categories on the Brazilian Institute of Geography and Statistics (IBGE) racial categories:
"The Brazilian Institute of Geography and Statistics (IBGE), which has conducted censuses in Brazil since 1940, racially classifies the Brazilian population in five categories: branco (white), pardo (brown), negro (black), amarelo (yellow), and indigenous. As in international practice, individuals are asked to self identify within these categories." https://en.wikipedia.org/wiki/Race_and_ethnicity_in_Brazil#IBGE's_racial_categories

Mougk:
> @Mougk no unfortunately this data doesn't have a good dictionary that I could find. I have based those categories on the Brazilian Institute of Geography and Statistics (IBGE) racial categories:
> "The Brazilian Institute of Geography and Statistics (IBGE), which has conducted censuses in Brazil since 1940, racially classifies the Brazilian population in five categories: branco (white), pardo (brown), negro (black), amarelo (yellow), and indigenous. As in international practice, individuals are asked to self identify within these categories." https://en.wikipedia.org/wiki/Race_and_ethnicity_in_Brazil#IBGE's_racial_categories

Thanks for looking into it. Let's use 'indigenous' then.

PR #1340, opened 2020-10-30
Update to readme to include all finished parsers

AnyaLindstromBattle:
I've added all the parsers that were missing to the README file. Would be good if respective parser authors could double check I've put the right information down before approval.


calremmel (approved):
Everything I know of is here! Thanks for doing this. Will keep on top of mine going forward.

PR #1341, opened 2020-10-30
Make Colombia convert_date function return dates in format usable by date filter

calremmel:
This is a quick fix to make the convert_date function return dates in the format mm/dd/YYYYZ.

I added a flag to the function to make sure that the Z would not be included in notes meant to be read by humans. Let me know if you know of a more elegant way to do this.


fojackson8 (approved):
Yes good spot!

iamleeg (approved):
Looks great! Thinking about your question on tidying up the different formats, I might prefer an enumeration to a boolean (so `destination=data_service|notes` instead of `dataserver=True|False`) to make it a little clearer why the format changes in the implementation. But nothing that need hold up this PR.

PR #1342, opened 2020-11-02
Attempted fix for Colombia date confirmed issue

fojackson8:
Changed if statement to stop Colombia parser breaking when checking if the field 'fecha_diagnostico' is populated.


ingestion/functions/parsing/colombia/colombia.py
```diff
@@ -115,7 +115,7 @@ def parse_cases(raw_data_file, source_id, source_url):
                         }
                     },
                 ]
-            elif entry["Fecha diagnostico"] == '':
+            elif not entry["Fecha diagnostico"]:
```
calremmel, line 118: Nit: I believe this could be simply `else`.

ingestion/functions/parsing/colombia/colombia.py
```diff
@@ -104,7 +104,7 @@ def parse_cases(raw_data_file, source_id, source_url):
                 "demographics": convert_demographics(
                     entry["Edad"],
                     entry["Sexo"])}
-            if entry["Fecha diagnostico"] != '':
+            if entry["Fecha diagnostico"]:
                 case["events"] = [
```
calremmel, line 108: I think this is a great solution, and it should definitely fix the problem.

If you wanted to condense the if/else statement to just the part that it changes, rather than creating the entire events array all at once, you could consider initializing `case["events"]` as an empty list, then creating something like `confirmation_event = {"name": "confirmed"}`, and then just having `if field is there, confirmation_event["dateRange"] = "blah", else confirmation_event["dateRange"] = "other_blah"`.

But this is all entirely opinion and up to you!

ingestion/functions/parsing/colombia/colombia.py
```diff
@@ -104,7 +104,7 @@ def parse_cases(raw_data_file, source_id, source_url):
                 "demographics": convert_demographics(
                     entry["Edad"],
                     entry["Sexo"])}
-            if entry["Fecha diagnostico"] != '':
+            if entry["Fecha diagnostico"]:
```
calremmel, line 107: I think this is a great solution, and it should definitely fix the problem.

If you wanted to condense the if/else statement to just the part that it changes, rather than creating the entire events array all at once, you could consider initializing `case["events"]` as an empty list, then creating something like `confirmation_event = {"name": "confirmed"}`, and then just having `if field is there, confirmation_event["dateRange"] = "blah", else confirmation_event["dateRange"] = "other_blah"`.

But this is all entirely opinion and up to you!

calremmel (approved):

ingestion/functions/parsing/colombia/colombia.py
iamleeg: This will still `KeyError`. You need to test `if "key" in dict:` to avoid that.

iamleeg (changes requested):

iamleeg:
It'd be great to include a test covering the case that failed to motivate this fix, so that we can see that the fix addresses it.

ingestion/functions/parsing/colombia/colombia.py
fojackson8: Should no longer KeyError. Chris you're right this would be a more elegant solution. I haven't implemented that yet, but will add it in if needed next time

ingestion/functions/parsing/colombia/colombia.py
```diff
@@ -43,36 +41,56 @@ def convert_gender(raw_gender):
...
-def convert_demographics(age: str, sex: str):
     demo = {}
-    if age:
-        demo["ageRange"] = {
-            "start": float(age),
-            "end": float(age)
-        }
-    if sex:
-        demo["gender"] = convert_gender(sex)
+    if entry['Edad']:
```
iamleeg, line 67: As before, this will still raise `KeyError` if the entry is not in the dictionary. Please change these to the safer `if 'Edad' in entry:`.

iamleeg (changes requested):

ingestion/functions/parsing/colombia/colombia.py
iamleeg: @fojackson8 and I discussed this, we think it's better to fail if the columns aren't present because it means the team gets informed that the data cannot be parsed, rather than the parser "succeeding" to create no cases or use incomplete data.

iamleeg (approved):

iamleeg (approved):

z023:
The last upload for Colombia was on 10th Nov and produced an error but I don't think any errors have gone through to the parser logs in Slack. Could it still be a problem with date confirmed?

iamleeg:
According to the logs, the last time it ran was yesterday (Mon 16th) when it successfully ingested 0 cases: it's set to filter for cases within one day. The CSV file is too big for most spreadsheets but I think there hasn't been a new case added since 9th November, so the behaviour is correct.

z023:
Ah ok so if an ingestion was successful but there are no cases to add/update it won't show in the UI uploads list.

iamleeg:
I think they should, but evidently aren't, so I've raised #1355 to ask for the behaviour to be changed.

PR #1343, opened 2020-11-02
Peru

fojackson8:
Added line to check if "UUID" field is populated before parsing the case


calremmel (approved):
I think this makes good sense! There are similar checks in some of the other parsers, and if the ID is missing that's certainly a reason to skip inclusion.

Out of curiosity - what do the rows that are missing UUIDs look like? Is it all missing data in such a case, some sort of repetition of headers, or real data that for some reason doesn't get assigned an ID in the source?

ingestion/functions/parsing/peru/peru.py
```diff
@@ -73,27 +72,28 @@ def parse_cases(raw_data_file, source_id, source_url):
...
-                "location": convert_location(entry),
-                "events": [
-                    {
-                        "name": "confirmed",
-                        "value": conf_methods.get(entry['METODODX']),
-                        "dateRange":
-                            {
-                                "start": convert_date(entry["FECHA_RESULTADO"]),
-                                "end": convert_date(entry["FECHA_RESULTADO"])
+            if entry["UUID"]:
```
iamleeg, line 75: This needs to be `if 'UUID' in entry:`

```
$ python3
Python 3.8.6 (default, Oct  8 2020, 14:06:32) 
[Clang 12.0.0 (clang-1200.0.32.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> d = {'a': 1, 'b': 2}
>>> if d['c']:
...   print(d['c'])
... 
  File "<stdin>", line 1, in <module>
KeyError: 'c'
>>> if 'c' in d:
...   print(d['c'])
... 
```

iamleeg (changes requested):

iamleeg (approved):

PR #1344, opened 2020-11-02
Allow Colombia parser to run where diagnosis date was not recorded.

iamleeg:


iamleeg:
Ah, just saw there's already a PR for this.

PR #1347, opened 2020-11-06
Changes to parser to deal with port quarantine cases

### LINKED ISSUE #1345: Some Japan cases ingested as Indonesia
Some cases (approx 750) in Japan data geocoded to Japan, 61361, Sooko, Mojokerto, East Java, Indonesia

From looking at the data it seems all of the cases which are geocoded as Indonesia are 'Port Quarantine' cases and so do not have any other location information other than 'Japan' e.g. 
  {
    "patientId": "11098",
    "dateAnnounced": "2020-07-12",
    "ageBracket": 20,
    "gender": "F",
    "residence": "Gifu",
    "detectedPrefecture": "Port Quarantine",
    "notes": "Returning from Pakistan",
    "prefecturePatientNumber": "Port#386",
    "prefectureSourceURL": "https://www.mhlw.go.jp/stf/newpage_12379.html",
    "sourceURL": "https://www3.nhk.or.jp/news/html/20200712/k10012511461000.html",
    "confirmedPatient": true
  }

So what I assume is happening is that in the absence of any other location information the geocoder looks up Japan and returns Japan, 61361, Sooko, Mojokerto, East Java, Indonesia. What I am not too sure about, however, is why it doesn't look up Japan as a country and return only 'Japan' but instead goes down the administrative levels? As I am not too familiar with how the geocoder works would be good to get confirmation from someone who knows - @calremmel @iamleeg @fojackson8 ?

In a separate but related question, do we still want to ingest these cases @Mougk ? How useful is it to have cases with only the country level geocoding? And is this even possible or will they fail validation? There is some location information in the "residence" field but whether this is prefecture or city information is inconsistent so not sure how useful it is.

AnyaLindstromBattle:
I realize this is quite a crude fix so please let me know if there is a better way. 
In the absence of knowing exactly how the geocoding works I am also not 100% sure this will return the country Japan and not the neighbourhood in Indonesia called Japan, but `query: country` seems to have worked in the case of Argentina. 
Closes #1345

### DIFF
--- a/ingestion/functions/parsing/japan/japan.py
+++ b/ingestion/functions/parsing/japan/japan.py
@@ -37,13 +37,16 @@ def convert_age(raw_age):
 
 
 def convert_location(raw_entry):
-    query_terms = [
-        term for term in [
-            raw_entry.get("detectedCityTown", ""),
-            raw_entry.get("detectedPrefecture", ""),
-            "Japan"]
-        if term and term != "Unspecified"]
-    return {"query": ", ".join(query_terms)}
+    if raw_entry["detectedPrefecture"] == "Port Quarantine":
+        return {"query": "Japan"}
+    else:
+        query_terms = [
+            term for term in [
+                raw_entry.get("detectedCityTown", ""),
+                raw_entry.get("detectedPrefecture", ""),
+                "Japan"]
+            if term and term != "Unspecified"]
+        return {"query": ", ".join(query_terms)}
 
 
 def detect_notes(raw_notes):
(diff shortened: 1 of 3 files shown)


iamleeg:
Did you try an end-to-end test to see what this does?

AnyaLindstromBattle:
Yes all works as expected. Would you like me to add a test case where we can see it working?

iamleeg (approved):
Nice one, thanks! And thanks for adding the other test :+1:

PR #1351, opened 2020-11-13
Acre (Brazil) parser

AnyaLindstromBattle:
Here is my take on the Acre parser.

I had written this several weeks ago but then the website crashed; it seems to work again now but the data has changed somewhat and I have included some changes in the parser to accommodate that.

Note that I spoke to some of you about the duplicate UUIDs present in this dataset; this is now no longer relevant as the new data does not include a UUID entry. 

@Mougk I would appreciate if you could look at the data at http://covid19.ac.gov.br/monitoramento/notificacoes/csv, specifically the 'sintomas' and 'outrosSintomas' columns. The 'outrosSintomas' column has been added in this new version of the data, and includes some symptom information like 'Peito chiando' (wheezing chest) but also some pretty random information such as 'ABUSO SEXUAL' (presumably the patient has been the victim of sexual abuse) or 'ACIDENTE DE MOTO' (motorcycle accident). In addition, some of the extra symptom information does not match between the 'sintomas' and 'outrosSintomas' entries; for instance, 'Perda Paladar' (loss of taste) may be reported in the 'outrosSintomas' but 'Disturbios Gustativos' not reported in 'sintomas'. Could you please take a look at the entries and let me know what, if any, data you want me to ingest from 'outrosSintomas', or if you think the info in 'sintomas' is sufficient to characterise the symptoms of the patient?


AnyaLindstromBattle:
After discussions we have decided to only include the information in the 'sintomas' column, as these symptoms correspond to standardized symptoms used by the Brazilian ministry of health.

iamleeg (approved):
Nice!

calremmel (approved):
Looks great! Very clean.

PR #1353, opened 2020-11-16
WIP: [#1006] Fix: Issue with non-persisting pagination state

rpoltorak:
#1006 Fix

**What:**
Now, when we move between the case details and the list view, the value doesn't change (#1006). At first, I wanted the changes to be consistent with the search (also stored in the same location object), but I've decided to do that in a more straightforward way so it could be more maintainable in the future. Now, the current page number and page size are persisted in the App component to avoid resetting between list re-render events.

**Why:**
Issue #1006: When the user returns to the list from the case details and pagination resets to the first page.

**Checklist:**
- [x] Make a fix for page number
- [x] Make a fix for page size
- [x] Fix request duplication
- [x] Add unit tests for this particular use case to check if the correct page number is set.
- [x] Add cypress tests
- [ ] Review internally (HTD)


rpoltorak:
I'm closing this PR due to problems with the coverage report. Code coverage significantly dropped from 8X to 5X percent and CodeCov is pointing to completely unrelated places and files that weren't applied and even touched in the pull request. Pretty much sure that is kind of a bug, especially that CodeCov is totally ok with the 100% same changes in the new PR that I've made here to check this: #1362.

PR #1362, opened 2020-11-18
[#1006] Current page number & page size should be persisted while moving between case details and list view

rpoltorak:
#1006 Fix

**What:**
Now, when we move between the case details and the list view, the value doesn't change (#1006). At first, I wanted the changes to be consistent with the search (also stored in the same location object), but I've decided to do that in a more straightforward way so it could be more maintainable in the future. Now, the current page number and page size are persisted in the App component to avoid resetting between list re-render events.

The issue was about the page number, but the page size should be included in this fix as well. Otherwise, it doesn't make sense to save only the page number. There was also a bug that is very closely related to the pagination feature. When the user is changing the page size, the request to the backend is sending twice. This problem also fixed in this pull request.

**Why:**
Issue #1006: When the user returns to the list from the case details and pagination resets to the first page.

**Technical note:**
Events related to page & pageSize had to be moved to the props of the custom Pagination component in that (kind of middleware-like) way to make sure that the internal state change has to happen before material-table requests the data.

**Checklist:**
- [x] Make a fix for page number
- [x] Make a fix for page size
- [x] Fix request duplication
- [x] Add unit tests
- [x] Add cypress tests
- [x] Review internally (HTD)

### DIFF
--- a/verification/curator-service/ui/cypress/integration/components/LinelistTableTest.spec.ts
+++ b/verification/curator-service/ui/cypress/integration/components/LinelistTableTest.spec.ts
@@ -336,4 +336,29 @@ describe('Linelist table', function () {
         cy.contains('Germany');
         cy.contains('United Kingdom');
     });
+
+    it('Pagination settings stays the same after returning from details modal', function () {
+        for (let i = 0; i < 7; i++) {
+            cy.addCase({
+                country: 'France',
+                notes: 'some notes',
+                sourceUrl: 'foo.bar',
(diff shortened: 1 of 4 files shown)


AKorzeniowski (approved):
Everything seems fine to me, good job! 🎖️

iamleeg (approved):
Looks good, nice test coverage :+1:

calremmel (approved):
Looks great 👍

Issue #1283, opened 2020-09-30
Add more details in line list view


Mougk:
Based on UX research more detail on line list cases was desired:

<img width="1057" alt="Screenshot 2020-09-30 at 11 49 11" src="https://user-images.githubusercontent.com/8478551/94676268-f898f300-0312-11eb-8998-5dd6728ecaf0.png">

Suggestion: add latitude and longitude, add detail on hospitalisation date, date of symptom onset.

allysonjp715:
There was also a suggestion to give the user the ability to add/remove columns as they see fit. This will need UX guidance though.

Mougk:
Great, @katelynnobrien may be able to help with UX post fellowship. For now, any comment from @acansler and @shroothless would be good!

allysonjp715:
Comments from @acansler in #1290

WeronikaMichaluk:
Information from Chris: 
We would like to add the following columns:

1. latitude location.geometry.latitude
2. longitude location.geometry.longitude
3. hospitalization date events
a. where name = hospitalAdmission
4. date of symptom onset events
a. where name = onsetSymptoms

The last two will not be present for every record. Some sort of N/A or blank may be displayed where those are not present.

**Placement of columns**
After **country** so it is grouped with other geographic info:

1.latitude location.geometry.latitude
2.longitude location.geometry.longitude

After **outcome**: (where are 2 empty columns)
They might be empty for some of the cases.
3. hospitalization date events
a. where name = hospitalAdmission
4. date of symptom onset events
a.where name = onsetSymptoms

PR #1363, opened 2020-11-18
[#1283] Add more details to list view (table)

AKorzeniowski:
**What:**
The list view now shows latitude and longitude and hospitalisation date/period along with symptoms onset date (where applicable). 

**Why:**
Per global.health team request

Checklist:
- [x]  Add latitude and longitude to table with proper rounding
- [x]  Add hospitalisation period and symptoms onset date to table 
- [x]  Add unit tests for this particular use case to check if the correct page number is set.
- [x]  Internally reviewed (HTD)

### DIFF
--- a/verification/curator-service/ui/src/components/LinelistTable.tsx
+++ b/verification/curator-service/ui/src/components/LinelistTable.tsx
@@ -36,6 +36,7 @@ import { WithStyles } from '@material-ui/core/styles/withStyles';
 import axios from 'axios';
 import { createStyles } from '@material-ui/core/styles';
 import renderDate, { renderDateRange } from './util/date';
+import { round } from 'lodash';
 
 interface ListResponse {
     cases: Case[];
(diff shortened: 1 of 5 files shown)


rpoltorak (approved):
Everything looks good to me. Accepted!

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -700,6 +721,14 @@ class LinelistTable extends React.Component<Props, LinelistTableState> {
...
                                                     ?.administrativeAreaLevel1,
                                             country: c.location.country,
+                                            latitude: round(
+                                                c.location?.geometry?.latitude,
+                                                4,
+                                            ),
+                                            longitude: round(
+                                                c.location?.geometry?.longitude,
+                                                4,
+                                            ),
```
iamleeg, line 731: The only thing that gave me pause here was "what if the lat/long fields are not set" (`_.round` returns NaN if the first arg is undefined, and 0 If it is null) but the data don't include any records without latitude or longitude so it's OK. I wonder if it's beneficial to add an assertion here to document that expectation and break if it ever happens. Not worth holding up the review over though, this works well :+1:

iamleeg (approved):

calremmel:
Looks good to me!

calremmel (approved):

PR #1367, opened 2020-11-19
Deal with Buddhist calendar year

AnyaLindstromBattle:
Changed parser to deal with use of both Gregorian and Buddhist calendar year.


AnyaLindstromBattle:
I also changed the folder and file names from 'Thai' to 'Thailand' but this doesn't seem to have gone down well.

calremmel:
> I also changed the folder and file names from 'Thai' to 'Thailand' but this doesn't seem to have gone down well.

Hmm! Did you change the path in template.yaml?

ingestion/functions/parsing/thailand/thailand.py
```diff
@@ -0,0 +1,104 @@
...
+
+def convert_date(raw_date):
+    """
+    Convert raw date field into a value interpretable by the dataserver.
+
+    The date is listed in YYYY-mm-dd HH:MM:SS format, but the date filtering API
+    expects mm/dd/YYYYZ format.
+    """
+    date = datetime.strptime(raw_date, "%Y-%m-%d %H:%M:%S")
+    # Some cases are reported using the Buddhist calendar which is 543 years ahead of the Gregorian
```
calremmel, line 27: Fascinating! We have 540 years before this would cause any confusion, so this looks good to me.

calremmel (changes requested):
I'm guessing that CI is failing because template.yaml also needs the new module path. As soon as that is working, I think this looks great.

AnyaLindstromBattle:
Hmm, changed the template.yaml path doesn't seem to have solved it although you're right, I definetly did need to do that. I am just wondering how git actually deals with changes to folder names, as in does it basically create a new folder with the new name, or does it just change the name of the existing folder? If it creates a new folder, does this mean that I have to also add __init__.py and requirements.txt again? If so, this may be why it is failing.

AnyaLindstromBattle:
Ok so we seem to have solved that particular issue as the tests all pass now... However I don't want all the previous commits to the thailand parser to be lost (i.e. when it was called thai), so ideally want to 'merge' thai and thailand. But may have to speak to our resident Github wizard @iamleeg for help with that. Note to file: don't rename folders and files in this way again.

AnyaLindstromBattle:
Unfortunately had to delete the old thai folder and replace with this thailand one, so there will be a break in tracking these files.

iamleeg (commented):
Looks like git has now correctly followed the renames! 🎆

AnyaLindstromBattle:
You're right, nice!

iamleeg (approved):
Looks like you've addressed @calremmel's request so I'm approving and merging this.

Issue #1360, opened 2020-11-18
Search is crashing the app while using the pagination in Line List view


WeronikaMichaluk:
Describe the bug
We've found a problem with the line list pagination. It occurs when I switch to a page other than the first one and try to search for something.

To Reproduce

Go to the Line List view.
Change a page to the second one.
Put something in the search input.
Hit enter.
See the white screen.
Expected behavior
A user should be redirected to the first page (or the same that has been selected before) with a search filter applied.

Environment (please complete the following information):

macOS
Google Chrome
Version 86.0.4240.183
Additional context
Error details: 185: Maximum update depth exceeded

Developer hint
Probably some operation in a code related to the search is causing the infinite loop.

PR #1369, opened 2020-11-23
#1360 Fix blank page after using search in certain cases

AKorzeniowski:
What:
Using search after switching the page or viewing a case details would result in blank page. This was caused by an infinite loop trigged on search query change in those cases. Search is now done using a different approach and flips back to page one after query change - regardless of whether the result includes the page a user was on when doing the search.

Hnet-image

Why:
To improve user experience, as this situation occurred quite often during our tests of the application (#1360)

Checklist:

- [x] Fix crash when using search after page change 
- [x] Add Cypress tests
- [ ] Review internally (HTD)


AKorzeniowski:
Closing this due to CodeCov issue - a new PR should solve this (hopefully 🙏 ).

Issue #1015, opened 2020-09-14
Adjust structure of downloaded csv file


z023:
Not all the demographic information available when viewing individual cases in the line list is shown when you download the line list. I have added a tab (CaseDemosVSLineListDownload) in the G.h Data Checking Issues google doc (https://docs.google.com/spreadsheets/d/1AZ-RHootmz5NOVyJ4c37fs6lNU3IN27xUZhNJemxWnE/edit#gid=0) where I have listed the demographic categories that are in individual case information and also listed the columns I see when I download the line list. I have noted what is missing or needs changing in the line list.

allysonjp715:
Remaining requests are
- remove `[]` from the download when the value is an empty array 
- expand arrays of objects into individual fields in download (e.g. events)

allysonjp715:
For the first request, `[]` does not show up in the download from the UI, it only shows up in the daily exports cc @attwad

axmb:
Was this intentionally reassigned to me? Is there something particular I can help out with, @allysonjp715 ?

allysonjp715:
I was leaving it open for anyone to pick up when they had the time. I described what's needed in my last two comments.

axmb:
Got it. I'm not actively working this, but will pick it up if I work through other issues.

sratcliffe118:
@axmb @z023 do we see this as blocking the friends and family launch? or can we bump this to be fixed prior to public launch?

z023:
Could probably push it back on the priority list at the moment.

WeronikaMichaluk:
HTD need to see an example during the call to better understand the task

What is expected and what do you receive. Examples when there is not an empty array.
turn it into a comma-separated string?
turn each existing symptom into a column?

Chris to check and update us on it.

PR #1371, opened 2020-11-23
#1015: Adjust downloaded CSV file structure

rpoltorak:
This PR covers issue #1015

**What:**
As we agreed, I've made the following changes to the downloaded csv file:

- For both fields: **symptoms** and **nationalities**.
  - Changed array to comma-separated values for the following fields:  
   For example: `["chest pain","cough"] -> "chest pain, cough"`
  - If the array is empty, return an empty string: `[] -> ''`

- **Events** are now separated to the following columns:
  - _events.confirmed.value_
  - _events.confirmed.date_
  - _events.onsetSymptoms.date_
  - _events.firstClinicalConsultation.date_
  - _events.selfIsolation.date_
  - _events.hospitalAdmission.value_
  - _events.hospitalAdmission.date_
  - _events.icuAdmission.value_
  - _events.icuAdmission.date_
  - _events.outcome.value_
  - _events.outcome.date_
  
**Comment:**
A single **event** has the date range, but the start & end dates are always exactly the same. At least for now. That's the reason I've decided to go with one output date for each event. Of course, I can split them into separate columns if necessary.

**Why:**
#1015: The current CSV file structure needs improvement

**Checklist:**
- [x] Meet all the requirements mentioned above
- [x] Add unit tests
- [ ] Review internally (HTD)

### DIFF
--- a/data-serving/data-service/src/controllers/case.ts
+++ b/data-serving/data-service/src/controllers/case.ts
@@ -8,6 +8,7 @@ import axios from 'axios';
 import { logger } from '../util/logger';
 import stringify from 'csv-stringify';
 import yaml from 'js-yaml';
+import { parseDownloadedCase } from '../util/case';
 
 class GeocodeNotFoundError extends Error {}
 
@@ -76,6 +77,7 @@ export class CasesController {
                     }>).map((datum) => datum.name);
                     casesQuery
                         .cursor()
+                        .map(parseDownloadedCase)
                         .pipe(
                             stringify({
                                 header: true,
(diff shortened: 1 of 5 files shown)


AKorzeniowski (approved):
Nice, I like the extraction of case events parsing to a separate file! 🙌

iamleeg (approved):
Looks great to me, thanks! Unfortunately I don't remember the original motivation for having date ranges for all the events but you're right they usually start and end on the same date, in fact usually there's only one date in the upstream source data that we construct a range from >_<

PR #1377, opened 2020-11-26
Brazil Para parser

AnyaLindstromBattle:
First look at Para parser. Feedback welcome.


ingestion/functions/parsing/brazil_para/para.py
```diff
@@ -0,0 +1,241 @@
...
+    "doenças autoimunes": "autoimmune disease"
+}
+
+_NONE_TYPES = set(["Não", "null", "undefined", None])
+
+_SYMPTOMS_MAP = {
+    "Dor de Cabeça": "headache",
+    "Distúrbios Gustativos": "taste alteration",
+    # Symptom ontology does not have a specific term for smell alterations
+    "Distúrbios Olfativos": "disturbances of sensation of smell and taste",
```
iamleeg, line 55: @Mougk should we add this term to the ontology, and is this phrasing OK?

iamleeg (commented):
Looks good, just want to get answer from @Mougk about symptom ontology before approving.

Mougk:
Does loss of smell exist? 

In any case it should be added because it is a common symptom of COVID-19

AnyaLindstromBattle:
Loss of smell does not exist. In any case Distúrbios Olfativos translates to disturbances to sense of smell rather than specifically a loss of smell. So what I could do is add 'smell alteration'?

Mougk:
Sounds good. Can we also add 'loss of smell' in the dictionary?

AnyaLindstromBattle:
@Mougk In this case I won't add it to the dictionary as the data doesn't specifically report a 'loss of smell'. But in future if a dataset does I will make sure to.

fojackson8 (approved):
Yep all looks good

PR #1378, opened 2020-11-26
Added line to check if Date of Confirmation is populated before parsing case

fojackson8:
Added to if statement that filters cases: now checks if Date of Confirmation AND UUID are populated before parsing case


iamleeg (approved):
:shipit:

PR #1379, opened 2020-11-28
First commit to USA parser

### LINKED ISSUE #1349: USA Data Importer
United States data from CDC available.
 
https://data.cdc.gov/Case-Surveillance/COVID-19-Case-Surveillance-Public-Use-Data/vbim-akqf 

Data include patients onset data, when it was entered in the database, age, sex, race, ...

AnyaLindstromBattle:
First look at USA parser. 
Main thing I would like input on is which data we want to consider the date of confirmation. I see two options:
1. cdc_report_dt which is the 'Initial case report date to CDC'. As far as I can see this data is present for each case in the dataset.
2. pos_spec_dt which is the 'Date of first positive specimen collection'. This data is only present for approx 300,000 out of the nearly 6 million cases in the dataset. 

Normally I would say just go with cdc_report_dt (which is how the parser is currently set up) but I have noticed in some cases the two dates are separated by up to several months. So it looks as though there is sometimes a delay between the patient collecting a sample which is positive and being reported to the CDC. 
This leaves us with two options, (a) use the cdc_report_dt and accept we will sometimes suffer from the delay in getting the info to the CDC, or (b) go with pos_spec_dt and only take into account 300,000ish cases...


Mougk:
Option #1. That's the positive date when result was entered in the database.

AnyaLindstromBattle:
OK cool, that's how it's already set up so don't need to change anything.

AnyaLindstromBattle:
Any comments on this @iamleeg @calremmel @fojackson8 ?

fojackson8 (commented):
All looks good to me!

fojackson8 (approved):

PR #1388, opened 2020-12-02
Changes to ignore bad age entry

AnyaLindstromBattle:
Deal with entries in age column of '-'.


iamleeg (commented):
Again, a test would be good to make this problem more visible.

AnyaLindstromBattle:
@iamleeg done!

ingestion/functions/parsing/brazil_espirito_santo/espirito_santo.py
```diff
@@ -173,9 +168,11 @@ def convert_preexisting_conditions(lung: str, kidney: str, diabetes: str,
         comorbidities.append(_COMORBIDITIES_MAP["ComorbidadeObesidade"])
 
     if comorbidities:
+        preexistingConditions["hasPreexistingConditions"] = True
         preexistingConditions["values"] = comorbidities
-
-    return preexistingConditions
+        return preexistingConditions
+    else:
+        return None
```
iamleeg, line 176: Just noting that most of these changes aren't about bad age entries :). It helps to keep the PR description, commit messages, and changes in sync so that the project history is an accurate record of the changes made, and so that we reviewers know what we're looking for in the changes.

ingestion/functions/parsing/brazil_espirito_santo/espirito_santo.py
```diff
@@ -232,7 +229,7 @@ def convert_demographics(gender: str, age: str, ethnicity: str, healthcare_proff
             demo["ageRange"] = {"start": float(age_range[0]), "end": float(age_range[2][:2])}
     if ethnicity:
         demo["ethnicity"] = convert_ethnicity(ethnicity)
-    if healthcare_proffesional not in _NONE_TYPES:
+    if healthcare_proffesional == "Sim":
```
iamleeg, line 232: there's a typo here, `healthcare_professional`. It doesn't need to hold up the PR but would be nice to fix.

iamleeg (approved):

ingestion/functions/parsing/brazil_espirito_santo/espirito_santo.py
AnyaLindstromBattle: Ok thanks, will make sure I have separate commits in the future.

ingestion/functions/parsing/brazil_espirito_santo/espirito_santo.py
AnyaLindstromBattle: Cool, will merge for now so that Zoe can fix the ingestion, but will tackle this next week.

PR #1391, opened 2020-12-03
Take into account travel info

AnyaLindstromBattle:
Changed parser so travel info is parsed correctly and not just in notes


Mougk (approved):
looks good to me!

iamleeg (approved):

PR #1392, opened 2020-12-03
Paraguay parser

AnyaLindstromBattle:
First look at Paraguay parser


iamleeg (changes requested):
Could you add a test for the case where the age is 999 please?

ingestion/functions/parsing/paraguay/paraguay.py
```diff
@@ -0,0 +1,115 @@
...
+                try:
+                    case = {
+                        "caseReference": {"sourceId": source_id, "sourceEntryId": row[_UUID], "sourceUrl": source_url},
+                        "location": {
+                            "query": ", ".join(
+                                [row[_DISTRICT], row[_DEPARTMENT], "Paraguay"]
+                            )
+                        },
+                        "events": convert_events(
+                            row[_DATE_CONFIRMED]
```
calremmel, line 98: Having an event with a date confirmed is one of the few strictly required fields in order to include a case. If I understand this right, the function will simply return None if a date is missing here, and then continue the parsing as normal.

Perhaps instead it should skip the case if date is missing, or assign the date from some other default if anything is available?

calremmel (changes requested):

ingestion/functions/parsing/paraguay/paraguay.py
AnyaLindstromBattle: True, all the current entries in the Paraguay dataset currently have a date of confirmation, but I can still include this quality control step in case we get an empty field in a future case?

AnyaLindstromBattle: Done :)

iamleeg (approved):
Thanks!

PR #1393, opened 2020-12-03
First look at Rio de Janeiro parser

fojackson8:


iamleeg:
Unfortunately I can't convince FOSSA to show me _what_ the dependency issues are. I think it's a false positive on this PR, though, as you don't AFAICT introduce any new dependencies.

ingestion/functions/parsing/brazil_rio_de_janeiro/rio_de_janeiro.py
```diff
@@ -0,0 +1,149 @@
...
+
+            if entry['dias']:
+                notes.append(f'Number of epidemiological days = {entry["dias"]}')
+            
+            if entry['dt_coleta_dt_notif']:
+                notes.append(f'Date case was reported was {convert_date(entry["dt_coleta_dt_notif"],dataserver=False)}')
+
+
+            case["notes"] = "\n".join(notes)
+            cases.append(case)
```
iamleeg, line 144: Other parsers use ',' as the separator for notes. I don't think I need to hold up this review, but I do think we probably need to standardise.

ingestion/functions/parsing/brazil_rio_de_janeiro/rio_de_janeiro.py
```diff
@@ -0,0 +1,149 @@
...
+    """
+    Parses G.h-format case data from raw API data.
+    
+    Only include cases which include a 'dt_evento' as confirmed date field. This loses ~400 cases out of 350k.
+    
+    Using residential municipality ('municipio_res') as case location, as its the only location field other than 
+    'bairro' (neighborhood), which is missing in ~2/5 cases and cannot be easily geocoded with mapbox.
+    
+    
+    """
```
iamleeg, line 73: Thanks for the detailed doc comment :+1:

iamleeg (approved):

ingestion/functions/parsing/brazil_rio_de_janeiro/rio_de_janeiro.py
```diff
@@ -0,0 +1,149 @@
...
+    with open(raw_data_file, "r") as f:
+        reader = csv.DictReader(f,delimiter=';')
```
calremmel, line 75: nit, not worth holding up review: might want a space after the comma in the DictReader arguments for consistency.

ingestion/functions/parsing/brazil_rio_de_janeiro/rio_de_janeiro.py
calremmel: Yes, I agree that standardization is a good idea, and I think that ',' as the separator is safer, as `\n`, while it can make things look nicer in certain applications, in others it can cause a bit of weirdness depending on how it is interpreted.

calremmel (approved):
Very nice work! LGTM

ingestion/functions/parsing/brazil_rio_de_janeiro/rio_de_janeiro.py
```diff
@@ -0,0 +1,149 @@
...
+                    }})
+            else:
+                case["symptoms"] = {
+                    "status": "Asymptomatic",
+                }
+
+            if entry["sexo"]=='I':
+                notes.append(f'Gender not specified in case notes.')
+
+            if entry["bairro"]:
```
AnyaLindstromBattle, line 133: The wording for the neighbourhood entry in notes is a bit different in other parsers (at least the ones I have written). We should probably agree on a way to standardize this?

ingestion/functions/parsing/brazil_rio_de_janeiro/rio_de_janeiro.py
AnyaLindstromBattle: Agree with @calremmel on this one.

fojackson8: Ok good point, will use ',' for future parsers.

PR #1395, opened 2020-12-04
Increase memory allocation for Japan, Germany, Sao Paolo

calremmel:
Increase memory to solve timeouts for Germany and Sao Paolo and inability to load data for Japan.


iamleeg (approved):
Looks great, let's see if this fixes things!

PR #1399, opened 2020-12-07
#1376 Exclude cases from the auto-ingestion process (UI part)

### LINKED ISSUE #1376: Handle excluded cases in curator portal (UI)
**Note:**
This is a front-end part of #1013.

**User stories:**
- As a user, I want to exclude a selected one or more cases from the list, so they can be ignored in the parsing process. 
- As a user, I want to re-include a selected one or more cases from the list, so they can be re-included again in the parsing process.
- As a user, I want to write a note about the excluded cases, so I can later check what was the reason for the exclusion.
- As a user, I want to have a separate status indicator for excluded cases on the list, so they can be easily recognizable.

**Actions needed:**
- For a standard case, introduce additional action to the existing context menu (three dots). That action should open the modal/dialog after the click event.
- The modal/dialog should contain input for the note and the confirmation button.
- After clicking on the confirmation button, send the information about the excluded case (or cases) to the curator-service endpoint
- For an excluded case, introduce additional action to the context menu. That action should send the information about the case that should be re-included again
- Introduce the same exclude and re-include actions for multiple cases. We could introduce that functionality by adding a new icon to the context menu on the top of the list displayed when multiple cases are selected. In that case, the same note would be applied to each of the selected cases.
- Introduce a new status and indicator state for the excluded cases (first column in the list). So there would be 3 statuses instead of 2: 
  - U: Unverified
  - V: Verified
  - E: Excluded. It could be red(ish) to increase readability.
- Excluded cases displayed on the list could be greyed out (opacity: 0.8?)

rpoltorak:
Closes #1376

This is the UI part of the bigger task: #1013. This PR contains changes that are necessary to exclude and re-include cases from the frontend perspective. You can find the whole description here: #1376.

**Technical note:**
As we agreed that we want to treat the exclusion as a third status (together with 'Unverified', and 'Verified'), we decided to introduce the endpoint for changing the case status and handling all logic for consistency purposes. That required me to refactor the LineList component a little bit and make the status change more generic.

**Important:**
For now, Cypress tests were commented out because they depend on the backend. We need to merge the backend endpoints first.

**Checklist:**
- [x] Fulfill the requirements specified in the issue
- [x] Add cypress tests
- [x] Review internally (HTD)


AKorzeniowski (approved):
Looks fine to me, we just need to sort out the Cypress tests. 🚀

verification/curator-service/ui/src/components/CaseExcludeDialog.tsx
```diff
@@ -0,0 +1,81 @@
...
+                            The following cases will be ignored in the ingestion
+                            process:
+                            <ul>
+                                {caseIds.map((id) => (
+                                    <li key={id}>{id}</li>
+                                ))}
+                            </ul>
+                            <TextField
+                                name="note"
+                                placeholder="Please specify a reason of the exclusion"
```
calremmel, line 54: Perhaps this should be "Please specify reason for exclusion"

calremmel (approved):
This looks great. I made a small note about one piece of text, but the path for curators is very good. I like the clarity around the options for exclusion/exclusion, and the request for reasoning.

verification/curator-service/ui/src/components/CaseExcludeDialog.tsx
rpoltorak: Of course, you're right done!

PR #1400, opened 2020-12-09
#1375 Add status change endpoint (incl. excluding)

### LINKED ISSUE #1375: Implement an endpoint for excluding the cases (and changing other statuses)
As a part of #1013, the following actions would be needed:

- Introduce a new endpoint to change status of multiple cases (including excluding them).
- The new endpoint should be implemented in both: curator-service (as a proxy) and data-service.
- The new endpoint should take a collection of case IDs to change status of, along with note (applicable only for exclusion):
  - POST: `/api/cases/batchStatusChange`
  - JSON: `{
  "caseIds": [ "mongodb uuid", ... ],
  "note": "string"
}`
- Store the new status in the `caseReference` subdocument of a case in `cases` table.
- Extend the existing case model by allowing a note, exclusionDate, and additional "excluded" status (along with unverified/verified).
- Extend current /cases endpoint with the exclusion details:
`{ "cases": [ ..., { "exclusionData": { "date": "standard format", "note": "string" }, ... } ] }`

AKorzeniowski:
This would close #1375 
This is the backend part of the bigger task: #1013. Whole description available here: #1376.

Technical note:
As is mentioned in #1399,  a new endpoint for changing the case status and handling all logic has been introduced for consistency purposes. This will require additional work on the frontend to accommodate that change.

**Important**
Database schema update is needed for excluding (since we introduced a new subdocument in case schema). The appropriate command to be executed in Mongo shell can be found here: https://gist.github.com/AKorzeniowski/0337ca2573d33484b61526f55279466f

Checklist:

- [x] Allow changes of status via new endpoint
- [x] Save note and date when excluding case
- [x] Remove exclusion data when a case is marked as not excluded
- [x] Support case status updates through query (for marking cases from search results that span across multiple pages) 
- [x] Handle edge cases (wrong IDs, missing note for exclusion)
- [x] Add unit tests
- [x] Review internally (HTD)

### DIFF
--- a/data-serving/data-service/src/controllers/case.ts
+++ b/data-serving/data-service/src/controllers/case.ts
@@ -540,18 +540,28 @@ export class CasesController {
             return;
         }
 
+        let updateQuery = {};
+
         try {
-            const validIdsCount = await Case.countDocuments({
-                _id: { $in: caseIds },
-            });
-            if (validIdsCount != caseIds.length) {
-                res.status(422)
-                    .send({
-                        message:
-                            'At least one of provided case IDs was not found. No records changed.',
-                    })
-                    .end();
-                return;
+            if (!caseIds) {
+                updateQuery = casesMatchingSearchQuery({
+                    searchQuery: req.body.query,
+                    count: false,
+                });
+            } else {
+                updateQuery = {
+                    _id: { $in: caseIds },
+                };
+                const validIdsCount = await Case.countDocuments(updateQuery);
+                if (validIdsCount != caseIds.length) {
+                    res.status(422)
+                        .send({
+                            message:
+                                'At least one of provided case IDs was not found. No records changed.',
+                        })
+                        .end();
+                    return;
+                }
             }
         } catch (err) {
             if (err.name === 'CastError') {
@@ -562,6 +572,7 @@ export class CasesController {
                     .end();
                 return;
             }
+            logger.error(err);
             res.status(500).json(err).end();
             return;
         }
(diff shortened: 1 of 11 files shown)


rpoltorak (approved):
Look ok. Great job!

data-serving/data-service/src/controllers/case.ts
```diff
@@ -520,6 +520,93 @@ export class CasesController {
...
+     * Handles HTTP POST /api/batchStatusChange.
+     * Receives an array of MongoDB IDs, status to be set for them and optional note.
+     * Note is used only when excluding cases (status set to "Excluded").
+     */
+    batchStatusChange = async (req: Request, res: Response): Promise<void> => {
+        const newStatus = req.body.status.toUpperCase();
+        const caseIds = req.body.caseIds;
+
+        if (newStatus === 'EXCLUDED' && !req.body.note) {
+            res.status(422)
```
iamleeg, line 535: I rarely see 422 used in practice, nice one! :+1:

data-serving/data-service/test/controllers/case.test.ts
```diff
@@ -23,8 +23,11 @@ const invalidRequest = {
     demographics: { ageRange: { start: 400 } },
 };
 
+const realDate = Date.now;
+
 beforeAll(async () => {
     mongoServer = new MongoMemoryServer();
+    global.Date.now = jest.fn(() => new Date('2020-12-12T12:12:37Z').getTime());
```
iamleeg, line 30: ah, interesting. I've used time-warp libraries to mock the date in tests before, but this is much simpler.

iamleeg (approved):
Looks great, I love seeing "warnings" that the coverage goes up, and I learned something about testing with jest too, thanks!

PR #1416, opened 2020-12-14
Cuba

fojackson8:
Bug fix and improved notes for Cuba


ingestion/functions/parsing/cuba/input_event.json
```diff
@@ -1,7 +1,7 @@
 {
     "env": "local",
     "s3Bucket": "epid-sources-raw",
-    "s3Key": "test/cuba/test_content.json",
-    "sourceUrl": "s3://epid-sources-raw/test/cuba/test_content.json",
-    "sourceId": "5f36d00420a86b00357e2de9"
+    "s3Key": "5fd3fb36edbfbf003c59e352/2020/12/11/2315/content.json",
+    "sourceUrl": "s3://epid-sources-raw/5fd3fb36edbfbf003c59e352/2020/12/11/2315/content.json",
```
AnyaLindstromBattle, line 5: I think sourceUrl should be the actual URL you get the data from, so for Cuba it is https://covid19cubadata.github.io/casos.html I think? Or whatever URL you used for the retrieval and end-to-end testing. This makes it easier for the curators (and everyone else) to find the correct link to set up ingestion as it's not always the same as the one in the original issue that was set up.

AnyaLindstromBattle (changes requested):

ingestion/functions/parsing/cuba/input_event.json
fojackson8: Good spot, I forgot to push latest version of input_event.json file, should be fixed now.

AnyaLindstromBattle (approved):
Looks good to me!

ingestion/functions/parsing/cuba/cuba.py
```diff
@@ -17,13 +17,15 @@
     import parsing_lib
 
 
-def convert_date(raw_date):
+def convert_date(raw_date, dataserver=True):
```
calremmel, line 20: Nice solution!

calremmel (approved):
LGTM!

PR #1421, opened 2020-12-15
Don't attempt to collate non-'Hospitalized' cases in covid19india data.

axmb:
Per title, limit ingesting `covid19india` data to rows with the `Hospitalized` `currentStatus`.

We'd previously hoped to get data with other statuses by deduping on a state-specific UUID (in other words, allowing non-`Hospitalized` cases to be ingested on the basis that they'd be folded into an existing row with the same state-specific UUID), but that doesn't appear to have held up.

We'll get a slightly lower quantity of detailed information this way, but the data we get will be solid. There's one last hitch, which is that we can't reliably process rows with negative `numCases`. This is mostly a wash, however -- for the `raw_data16.csv` file, out of ~1.3M cases, the total of negative case counts for hospitalizations is 571.

On the whole, the linelist style data is a bit all over the place from this source. This is probably the best we can do without some sort of multi-file processing pipeline (or at least an offline pipeline where it's feasible to store and inspect an entire file's worth of data while we're processing the data).

### DIFF
--- a/ingestion/functions/parsing/india/india.py
+++ b/ingestion/functions/parsing/india/india.py
@@ -103,61 +103,22 @@ def convert_sources(row):
     return additionalSources or None
 
 
-def update_for_outcome(row, case):
-    if row["Current Status"] == "Hospitalized":
-        case["events"].append({
-            "name": "hospitalAdmission",
-            "dateRange":
-            {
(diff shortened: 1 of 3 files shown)


iamleeg (approved):
Looks good, shame we can't get the extra info but better to have dependable data. Thanks!

PR #1423, opened 2020-12-15
Expand Germany parser to include onset of symptoms and convert to dictReader

AnyaLindstromBattle:
Expanded the parser as @Mougk noticed date of symptoms entry was included. Also changed it it to a csv.dictReader as this is more robust to changes in data format.

### DIFF
--- a/ingestion/functions/parsing/germany/germany.py
+++ b/ingestion/functions/parsing/germany/germany.py
@@ -15,14 +15,16 @@ except ImportError:
             'common/python'))
     import parsing_lib
 
-_DATE_INDEX = 8
-_AGE_INDEX = 4
-_GENDER_INDEX = 5
-_CASECOUNT = 6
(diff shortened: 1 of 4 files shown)


ingestion/functions/parsing/germany/germany.py
```diff
@@ -75,8 +77,7 @@ def parse_cases(raw_data_file: str, source_id: str, source_url: str):
     Parses G.h-format case data from raw API data.
     """
     with open(raw_data_file, "r") as f:
-        reader = csv.reader(f)
-        next(reader)  # Skip the header.
+        reader = csv.DictReader(f, delimiter=",")
```
calremmel, line 80: nit: "," is the default for all objects in the csv module, so no need to specify. Not that it hurts!

calremmel (approved):
Looks excellent. Thanks for this!

PR #1424, opened 2020-12-15
WIP: #1385 Add endpoint for fetching list of excluded case IDs

### LINKED ISSUE #1385: Implement an endpoint for fetching the excluded case ids.
As a part of #1013, the following actions would be needed:

- Introduce a new endpoint to fetch excluded case ids.
- The new endpoint should be implemented in data-service.
- The new endpoint should return a collection of case ids:
  - GET: `/api/excludedCaseIds`
  - JSON: `{
  "cases": [ "sourceEntryId" ],
}`
- Obligatory filters: sourceId
- Optional filters: dateFrom, dateTo

AKorzeniowski:
This would close #1385
This is the second backend part of the bigger task: #1013. 

Checklist:

- [x] Fetch excluded case IDs based on source ID
- [x] Allow filtering by date range
- [x] Add tests
- [x] Review internally (HTD)

### DIFF
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -599,7 +599,7 @@ paths:
             format: date
       responses:
         "200":
-          $ref: "#/components/responses/200"
+          $ref: "#/components/responses/200CaseIdsArray"
         "400":
           $ref: "#/components/responses/400"
         "403":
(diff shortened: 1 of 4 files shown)


rpoltorak (approved):
Everything looks good, great job👍   No idea what's wrong with cypress tests as they are completely not related... I've tested it locally and it works.

PR #1429, opened 2020-12-16
Fix search guide dialog

rpoltorak:


calremmel:
Hi @rpoltorak ! There are a couple of failing tests, just let me know if they are ok to ignore. Many thanks!

rpoltorak:
> Hi @rpoltorak ! There are a couple of failing tests, just let me know if they are ok to ignore. Many thanks!

As you can see they are totally not related to that PR, so we could ignore them for this PR. BUT would be great to remember that and take a look and see what's the reason why they are failing later. This is weird and also occurs in the other PR here: https://github.com/globaldothealth/list/pull/1424.

PR #1450, opened 2020-12-30
Add catch for case exclusion when no date limits

calremmel:
The new case id exclusion endpoint requires either a date_range or a date_filter argument. For sources small enough that the entire dataset is ingested every time, this causes an error.

This PR adds a catch, retrieving all excluded case ids from the endpoint when no date limits are included for a source.


iamleeg:
Hi, sorry to miss out reviewing in time, I was still on leave. Just wanted to check is this case (no date filter _and_ no date range) supported in the rest of the parsing lib, i.e. in invoking the parser? I remember bugs before where this wasn't consistent (both were set and one was unused, I think) so just want to be certain that this case works consistently.

PR #1451, opened 2020-12-31
Thailand errors

AnyaLindstromBattle:
Edited Thailand parser to deal with errors when parsing gender and age. Also added nationalities as an entry in demographics.


calremmel (approved):
LGTM!

PR #1454, opened 2021-01-05
Changes to location parsing

### LINKED ISSUE #1407: Geocoding issues with DistritoFederal parser
There are some geocoding issues with the Distrito Federal parser which mean that certain cases are being geocoded to other Brazilian states (specifically with the most recent upload it is Parana and Espirito Santo). As you can see in the below screenshot it seems in both cases it is geocoding it to 'Rua Distrito Federal', so using Distrito Federal as a street name rather than the state (ignore the weird symbols this is because of Windows encoding).

image

The parser is set up to only ingest cases where state == ESPIRITO SANTO so this is not the issue. I have also looked at all the individual cases where the geocoding is failing to see if I can spot a pattern but so far I haven't found anything. For instance, for the second to last Parana case in the above screenshot, if the data is filtered by the case information it could correspond to any of the 4 following cases

image

As you can see there is no common feature, I thought it may occur because RA (municipality) was unpopulated but this is not the case.

As I was going through this I remembered a conversation I had with AB and Tim a while back about this which I think is related; essentially it seems some (or most, in fact) municipalities in Distrito Federal are not in Mapbox, see for instance Lago Sul. For these locations, there is only a match up until Distrito Federal (this may have something to do with the unique administrative organization of the DF state see https://en.wikipedia.org/wiki/Administrative_regions_of_the_Federal_District_(Brazil)). So I'm wondering if the unrecognized municipalities are confusing Mapbox, making it take Distrito Federal as a street name in some cases, and therefore returning the locations in the other states. 

I think the only solution to this is that we basically set the query as 'Distrito Federal, Brazil' for all cases, and include the provided municipality info in the notes. Do you agree?

AnyaLindstromBattle:
Changed parser to deal with most locations without geocoding as this was returning erroneous locations. 

Closes #1407

### DIFF
--- a/ingestion/functions/parsing/brazil_distrito_federal/dictionaries.json
+++ b/ingestion/functions/parsing/brazil_distrito_federal/dictionaries.json
@@ -1,0 +1,143 @@
+{
+    "comorbidities": {
+        "Pneumopatia": "lung disease",
+        "Nefropatia": "kidney disease",
+        "Distúrbios Metabólicos": "disease of metabolism",
+        "Obesidade": "obesity",
+        "Cardiovasculopatia": "cardiovascular system disease"
+    },
+    "admin_districts": {
+        "Plano Piloto": {
+            "latitude": -15.7939,
+            "longitude": -47.8828
+        },
+        "Gama": {
+            "latitude": -16.0194,
+            "longitude": -48.0669
+        },
+        "Taguatinga": {
+            "latitude": -15.8333,
+            "longitude": -48.0564
+        },
+        "Brazlândia": {
+            "latitude": -15.675,
+            "longitude": -48.2008
+        },
+        "Sobradinho": {
+            "latitude": -15.6531,
+            "longitude": -47.7914
+        },
+        "Planaltina": {
+            "latitude": -15.6192,
+            "longitude": -47.6525
+        },
+        "Paranoá": {
+            "latitude": -15.7686,
+            "longitude": -47.7797
+        },
+        "Núcleo Bandeirante": {
+            "latitude": -15.87,
+            "longitude": -47.9678
+        },
+        "Ceilândia": {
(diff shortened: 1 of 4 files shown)


ingestion/functions/parsing/brazil_distrito_federal/dictionaries.json
```diff
@@ -0,0 +1,143 @@
...
+        "Nefropatia": "kidney disease",
+        "Distúrbios Metabólicos": "disease of metabolism",
+        "Obesidade": "obesity",
+        "Cardiovasculopatia": "cardiovascular system disease"
+    },
+    "admin_districts": {
+        "Plano Piloto": {
+            "latitude": -15.7939,
+            "longitude": -47.8828
+        },
```
iamleeg, line 13: I'm OK with these being here but wanted to ask what the motivation is, does map box get these wrong?

ingestion/functions/parsing/brazil_distrito_federal/dictionaries.json
```diff
@@ -0,0 +1,143 @@
+{
```
iamleeg, line 1: Good idea to separate the data from the code like this :+1:

iamleeg (approved):
Looks good!

ingestion/functions/parsing/brazil_distrito_federal/dictionaries.json
AnyaLindstromBattle: Yeah exactly, see https://github.com/globaldothealth/list/issues/1407. essentially because the Distrito Federal in Brazil is not organized like other States but has it's own Administrative divisions, these are not in Mapbox and were geocoded wrong

iamleeg: thanks!

PR #1458, opened 2021-01-07
Update dependencies to address security vunerabilities.

iamleeg:
Github and npm are complaining particularly about the axios vulnerability, so here's a fix before I move on to changing the UI code. I've run all the automated tests locally and seen that they still pass, and we should wait for the CI run before approving this PR.


iamleeg:
Ah, just saw that dependabot has partially done this, I'll rebase this commit.

iamleeg:
Rebased and there's nothing new in package.json so this can go away.

PR #1459, opened 2021-01-08
Distinguish email notification failures from backend errors

iamleeg:
~~~Draft for the moment, as incomplete and not working.~~~

### DIFF
--- a/verification/curator-service/api/openapi/openapi.yaml
+++ b/verification/curator-service/api/openapi/openapi.yaml
@@ -168,6 +168,8 @@ paths:
       responses:
         "200":
           $ref: "#/components/responses/200Source"
+        "207":
+          $ref: "#/components/responses/207"
         "400":
           $ref: "#/components/responses/400"
         "403":
@@ -1823,6 +1825,8 @@ components:
         application/json:
           schema:
             $ref: "#/components/schemas/LocationArray"
+    "207":
+      description: Mixed response
     "400":
       description: Malformed request
     "403":
--- a/verification/curator-service/api/openapi/openapi.yaml
+++ b/verification/curator-service/api/openapi/openapi.yaml
@@ -168,8 +168,6 @@ paths:
       responses:
         "200":
           $ref: "#/components/responses/200Source"
-        "207":
-          $ref: "#/components/responses/207"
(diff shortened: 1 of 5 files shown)


iamleeg:
Upgrading this to ready for review: there's still some UX to iron out but in its current state it's both tested, and solves the immediate problem that's blocking adding new sources. My summary here of the various things I've written about the bug and its fix, to help reviewers and motivate future work.

## What was broken

There are three things that happen when you edit a source to change the AWS schedule.

1. send AWS the updated schedule information.
2. validate the change, and store the updated source (if valid) in mongodb as a source document.
3. send an email to the registered contacts for the source, informing them of the update.

Except for: those things didn't happen in that order, and if any of them failed it would block the others from happening! We were updating AWS, sending email, then saving the document. Due to an unrelated problem that I'll raise separately, the email wasn't being sent, so the schedule would be updated in AWS but there was no way to find out: no notification, and the portal didn't reflect it because the change wasn't being stored in mongo.

## What is now less broken

I've re-ordered the steps to happen in the listed order: if sending to AWS fails, then the portal isn't updated: if sending email notifications fails, it doesn't stop the portal updating. Additionally I distinguish the "email failed" error from other failures in the frontend, so they don't stop the portal updating.

## What is now more broken

If the email notification fails, it isn't shown in the error alert bar. I don't know why this is: the error is set in the component state, but then doesn't get shown.

abhidg:
The change to do notifications separately from updating AWS seems fine. I'm not familiar with mock to comment on that bit, there is one test failing: ` ✕ sends specified properties via email (1242 ms)`, is that a limitation of github actions?

iamleeg:
@abhidg that's a timeout connecting to the [ethereal service](http://ethereal.email/). That test actually connects to an SMTP black hole relay to send emails: I've talked about that on Slack (suggesting that we replace it with a local mock).

calremmel (approved):
I think this looks great, nice move for getting the curators able to move forward. We can tackle the other aspects (more specific error messages, fixing email notifications) when we have the luxury of time to do so.

Afraid I'm not qualified to comment on the javascript, but don't want to hold merging up any longer! Approved from my end.

PR #1463, opened 2021-01-11
Republic of korea

fojackson8:


ingestion/functions/template.yaml
```diff
@@ -416,6 +416,42 @@ Resources:
...
+      CodeUri: parsing/brazil_rio_de_janeiro/
+      Handler: rio_de_janeiro.lambda_handler
+      Description: Parse case data for the Rio de Janeiro State of Brazil
+      # Function's execution role
+      Policies:
+        - AWSLambdaBasicExecutionRole
+        - AWSLambdaReadOnlyAccess
+      Layers:
+        - !Ref ParsingLibLayer
+  NewZealandParsingFunction:
```
calremmel, line 431: Looks like the checks are failing on New Zealand, since there isn't a parser up for it yet. Perhaps you can remove that for the time being?

fojackson8: Good spot, thanks

fojackson8:
Note I've pushed these without locally invoking - tests on Github all passed successfully, but may be worth a more stringent check.

calremmel (approved):
LGTM!

PR #1464, opened 2021-01-12
New Zealand

fojackson8:


fojackson8:
Note the URL in the input_event.json is a recent copy of the data I uploaded to an s3 bucket for testing. The actual NZ URL is unstable, as it changes each day:

https://www.health.govt.nz/system/files/documents/pages/covid_cases_2021-01-11.csv

So as a source URL, we want to use:
https://www.health.govt.nz/system/files/documents/pages/covid_cases_$FULLYEAR-$FULLMONTH-$FULLDAY.csv

ingestion/functions/parsing/new_zealand/new_zealand.py
```diff
@@ -0,0 +1,135 @@
...
+    if raw_gender == "Male":
+        return "Male"
+    if raw_gender == "Female":
+        return "Female"
+    return None
+
+
+def convert_location(raw_entry):
+    if 'Managed Isolation' in raw_entry['DHB']:
+        return None
```
calremmel, line 49: Should this return `{"query": "New Zealand"}` rather than `None`? I believe that a location is required or the insertion will fail.

ingestion/functions/parsing/new_zealand/new_zealand.py
```diff
@@ -0,0 +1,135 @@
...
+    '''
+    demo = {}
+    if entry['Age group']:
+        if '90+' in entry['Age group']:
+            demo["ageRange"] = {
+                "start": 90,
+                "end": 120
+            }
+        else:
+
```
calremmel, line 66: nit: extra empty line

ingestion/functions/parsing/new_zealand/new_zealand.py
```diff
@@ -0,0 +1,135 @@
...
+                case = {
+                    "caseReference": {
+                        "sourceId": source_id,
+                        "sourceUrl": source_url
+                    },
+                    "location": convert_location(entry),
+                    "demographics": convert_demographics(entry)
+                }
+
+                if entry["Report Date"]:
```
calremmel, line 107: Are there situations where there is not a report date? A confirmed date is a required field, so if there are missing dates, perhaps those cases should be skipped?

calremmel (changes requested):
Nice work! Just a few questions and perhaps a couple of fixes.

ingestion/functions/parsing/new_zealand/new_zealand.py
fojackson8: @calremmel They're cases coming from 'Overseas', but they don't specify where. Following your suggestion, I've changed it so that Overseas cases arriving in New Zealand are geocoded as New Zealand, but this might be something we want to discuss. @AnyaLindstromBattle for visibility.

ingestion/functions/parsing/new_zealand/new_zealand.py
fojackson8: Fair point - all confirmed cases so far have had a report date. We now only include cases with a report date AND status=confirmed

ingestion/functions/parsing/new_zealand/new_zealand.py
calremmel: Sounds great! Thanks Felix. This looks good to me now as soon as you give me the go ahead on this point.

calremmel (approved):
LGTM!

ingestion/functions/parsing/new_zealand/new_zealand.py
fojackson8: Yep, that's fine from my side

PR #1473, opened 2021-01-14
Change Argentina notes delimiter from "\n" to ", "

calremmel:
Using newline as the delimiter for the notes was unfortunately causing some parsing issues for the download for the whole dataset! This is a quick fix, simply switching out newlines for commas to match the other parsers.


AnyaLindstromBattle (approved):
Sounds great, will also be good for consistency.

PR #1475, opened 2021-01-15
Import data from the old spreadsheet-based line list.

iamleeg:


iamleeg:
@mougk your list of countries without ADI to drop doesn't include Czech Republic, but we have a parser (under the country name Czechia): should we drop those cases too?

Mougk:
Yes, drop them.

iamleeg:
OK, ready for review now. If you compare the R script with the Python script, you can see that I didn't do the same thing regarding filtering the France cases separately. On inspection, it looks like this doesn't pick up any cases that aren't already filtered out in the first step.

abhidg:
1. China is present in the list of countries in `ghtransfer_15012020.R`, but not in the Python file.
2. The script seems to require only these, are the extra dependencies for the Jupyter notebook?

```python
requests==2.21.0
numpy==1.19.5
pandas==1.2.1
```
Here's the output from the Python code:

```
allData total cases: 2676310
detailedData total cases: 2556829
detailedDataWithoutSources total cases: 2001499
```

detailedData total cases is different from that in the Jupyter notebook, is this expected?

iamleeg:
1. Good spot, thanks!
2. yes, the rest of the requirements are for the notebook.

Case count: it's expected that the case count is different, because the methodology is different. @Mougk thought that the original import was too conservative (particularly because it dropped data for which we _aspired_ to have ingestion, but don't yet), so came up with slightly different rules. The notebook is the original script, and the python script uses the new filter method.

With China excluded, the case count is 1108685, which is much more like an expected value.

abhidg (approved):

PR #1476, opened 2021-01-15
Peru

fojackson8:
This should fix the age issue - we now filter out cases whose age > 120. I've included the case that broke it in the unit test sample_data


calremmel (approved):
Great change!

PR #1481, opened 2021-01-18
clean commit of scotland parser

fojackson8:


ingestion/functions/parsing/scotland/scotland.py
```diff
@@ -0,0 +1,103 @@
...
+    import parsing_lib
+
+
+def convert_date(raw_date: str, dataserver=True):
+    """
+    Convert raw date field into a value interpretable by the dataserver.
+    The date is listed in mddyy format,
+    """
+    date = datetime.strptime(raw_date, "%Y%m%d")
+    if not dataserver:
```
calremmel, line 25: I really like this solution!

ingestion/functions/parsing/scotland/scotland.py
```diff
@@ -0,0 +1,103 @@
...
+                                    {
+                                        "start": convert_date(entry["Date"]),
+                                        "end": convert_date(entry["Date"])
+                                    }
+                            }
+                        ],
+                        "demographics": convert_demographics(entry),
+                        "location": {"query": 'Scotland'}
+                    }
+                    yield case
```
calremmel, line 99: Rather than creating the case multiple times and yielding once apiece, perhaps you could consider creating the case once and yielding multiple times? i.e.

```
for _ in range(int(entry['DailyPositive'])):
    yield case
```

The advantage is that you only create the dictionary once, which is a little more lightweight

ingestion/functions/parsing/scotland/scotland.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+
+    with open(raw_data_file, "r") as f:
+        reader = csv.DictReader(f, delimiter=',')
+        cases = []
+        notes = []
+        for entry in reader:
+            if int(entry['DailyPositive']
+                   ) > 0 and entry['Sex'] != 'Total' and entry['AgeGroup'] != 'Total':
+                for k in range(0, int(entry['DailyPositive'])):
```
calremmel, line 80: nit: no need to have the 0 as the first argument, as that is the default anyhow.

calremmel (changes requested):
Great work! Just one change I think would be worthwhile, then good to go.

PR #1487, opened 2021-01-25
Task 1435 login screen

maciekz1996:
Changes:

* Added AWS Amplify
* Added default Google sign in button
* Added the ability for a user to sign up/in using different email account than gmail
* Implemented Verification codes functionality when signing up/in using email address
* Moved LandingPage component to src/landing-page
* Updated AppTest.spec.tsx in Cypress

Screenshots:

Screen Shot 2021-01-25 at 16 25 50

**AWS Amplify setup**

First run: `npm install -g @aws-amplify/cli`

Then inside `list/verification/curator-service/ui` -> `amplify pull` and follow the instructions

There are two environments available **develop** and **prod** (Both configured the same for now)

Until the AWS SES (Simple Email Service) is moved out of a sandbox only verified email addresses can receive verification codes.


sergioloporto (approved):

maciekz1996:
@calremmel Today I was working on making `amplify init` run by default on each pull request. I edited `curator-ui-node.yml` and I used `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` that are already stored as Github secrets, however the user that is connected with those keys seems to not have access to Amplify and because of that the `amplify init` command is failing (screenshot below). Could you grant this user whose keys are stored in Github secrets access to AWS Amplify ? Thanks in advance ! 

Screen Shot 2021-01-28 at 15 35 52

calremmel:
> @calremmel Today I was working on making `amplify init` run by default on each pull request. I edited `curator-ui-node.yml` and I used `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` that are already stored as Github secrets, however the user that is connected with those keys seems to not have access to Amplify and because of that the `amplify init` command is failing (screenshot below). Could you grant this user whose keys are stored in Github secrets access to AWS Amplify ? Thanks in advance !
> 
> Screen Shot 2021-01-28 at 15 35 52

That makes good sense, I will work on this now!

Issue #1468, opened 2021-01-13
HTD_ splitted issue On Data page, change default "Download" function to export CSV of daily cached full data set #1436


WeronikaMichaluk:
For MVP instead of initiating DL of data set based on applied filters, initiate a download of full daily cached data set and rename button "Download full data set". This will ensure a consistent and stable experience and allow users to sort, filter, and manipulate data via local applications while we further define and refine the Data page UX post-launch. This will entail:

- Designing new "Download full data set" button
-  Create interstitial modal explaining data set being downloaded is daily cached full G.h data set and may not exactly match record count in UI, total or filtered. (exact language TBD)

WeronikaMichaluk:
@joe-brilliant @calremmel 
I have splitted issue 1436 into Front and Backend - > this one is about Frontend. 

As for Designing button -> should we use UI UX library or you have any specific design for that?

WeronikaMichaluk:
@joe-brilliant @calremmel 
Could you please provide us with more details like some screenshots where does this task apply to? Could you please attach here some screenshots with some more description, please?
Where should changes be made?
We are not sure where is that mentioned button?
Could you please advise where should we create/design: **interstitial modal explaining data set being downloaded is daily cached full G.h data set and may not exactly match record count in UI, total or filtered. (exact language TBD)**

calremmel:
@WeronikaMichaluk Of course! The button is here:

![image](https://user-images.githubusercontent.com/40274075/104615083-df644a00-5656-11eb-8b4b-424a5322a39b.png)

@rkassa may have already created this button. @rkassa is that right? If not, it would be great for you to design.

The modal should appear when you click that button. The modal should have the following components:

1. Text along the lines of the following (will change but this is good for now).

> This download link provides access to the full Global.health line list dataset, cached daily at 12:00am UTC. Any cases added past that time will not be in the current download, but will be available the next day.

2. Link to download latestdata.tar.gz on S3 (not generated yet, link to be provided probably next week)

@joe-brilliant If you have any more directions for the modal text, please feel free to share here.

joe-brilliant:
Thanks all, text looks good!

zbigiman-htd:
I've made the dialog modal:

<img width="1791" alt="image" src="https://user-images.githubusercontent.com/69294239/105143860-b59a9f80-5afc-11eb-931e-1c386c8a0f9f.png">

@joe-brilliant is it OK?

PR #1489, opened 2021-01-27
Task 1468 download data modal

maciekz1996:
Changes:

* After clicking Download button in Toolbar modal informing that the full data set is going to be downloaded is shown
* Created new api endpoint `/api/cases/getDownloadUrl` responsible for generating signed url needed to download file from S3 Bucket

Screenshots:

Screen Shot 2021-01-27 at 15 01 24


sergioloporto (changes requested):
It looks like that according tot the [issue](https://github.com/globaldothealth/list/issues/1468), the button name should be "download full datase".
And probably "dataset" instead than "data set". Like this:

image

Need to confirm with @calremmel the correct working: dataset or data set.

verification/curator-service/api/src/index.ts
```diff
@@ -130,6 +131,7 @@ const awsEventsClient = new AwsEventsClient(
     awsLambdaClient,
     env.SERVICE_ENV,
 );
+const s3Client = new AWS.S3({ region: 'us-east-1', signatureVersion: 'v4' });
```
sergioloporto, line 134: Is it right, or should it be us-east-2?

sergioloporto (changes requested):

verification/curator-service/api/src/index.ts
calremmel: us-east-1 is correct! many of the s3 buckets are on 1

sergioloporto (approved):

calremmel:
@maciekz1996 This looks great. I noticed that the Cypress tests failed for Chrome - looks like for some reason it couldn't run them. Any thoughts for why? Let me know if it is safe to ignore.

maciekz1996:
@calremmel The Cypress tests were all passing before changing button text from "Download" to "Download full dataset". I noticed that Cypress test window is small and because of that input in SearchBar becomes too small to be focused by Cypress `cy.type(...)` due to wider button with longer text next to it. Let me know if I should try to resolve this issue.

maciekz1996:
@calremmel The issue with Cypress tests failing is now fixed

PR #1494, opened 2021-01-29
562 manual data entry from line list

sergioloporto:
- Fixed text according to the excel
- added the source URL validation logic


maciekz1996 (changes requested):
Everything looks fine, just two things that I noticed:

1) Tooltip in Event section is too long and some of the content is not visible (maybe it would be a good idea to make this tooltip scrollable ? 🤔 )
2) I saw that two tests are failing (CaseForm.test.tsx and NewCaseForm.spec.ts)

maciekz1996 (approved):

calremmel:
@maciekz1996 Looks great to me! There is a conflict with main, and I'm not totally familiar with that part of the codebase. Could you give it a look?

PR #1496, opened 2021-02-01
Deal with non-standard date format

AnyaLindstromBattle:
This should solve [ERROR] ValueError: Unhandled data: time data '2020/11/23 0' does not match format '%Y/%m/%d %H:%M:%S'

Also will make the parser more resistant to future errors.


fojackson8 (approved):
Agreed

PR #1499, opened 2021-02-02
Task 1441 hide drawer

maciekz1996:
Changes:

* Navigation drawer on the left is hidden by default for non curator users
* Linelist text changed to COVID-19 Linelist


sergioloporto (approved):

calremmel (approved):
Looks great!

Issue #1502, opened 2021-02-03
Filter Data using Variant of Concern


calremmel:
Users should be able to filter cases in Data by Variant of Concern.

- [x] Users should be able to filter for specific Variants
- [x] Users should be able to filter for any Variants (i.e. not null or equiv)

iamleeg:
Can we move the third point to #1501 @calremmel @rkassa @joe-brilliant? Otherwise there's a cyclic dependency where we can't build the map until filtering is done, but we need to have the map feature before we can do filtering.

calremmel:
@iamleeg Of course you're right, that makes no sense there. I'll move it now!

PR #1513, opened 2021-02-04
1502 voc filter

iamleeg:
Going to open this up to review now with 1/3 requirements from the checklist at #1502 implemented to get input on my plan for the other 2/3. Variant field is parsed in the search on the backend, and available in the search bar on the frontend.

 - there's a request for a "non-null" search, i.e. for any variant. Our current search doesn't allow for that, so I was thinking of testing for the string `*` as a search value and special-casing that. `variant:foo` would execute the query `Case.find().where('variant.name').equals('foo')` as is currently the case, but `variant:*` would execute `Case.find().where('variant.name').exists()`. This would work for all other properties, e.g. `admin3:*`.
 - map: I haven't looked at the map code yet at all. I know it's going to be easy to add the variant to the redirect URL for the line list, but what I'm wondering is how the UX could work? How will a user indicate on the map that they want to look at, e.g., P.1 in Brazil?

Thanks!

### DIFF
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -62,7 +62,7 @@ paths:
                 with a comma to union them. Only equality operator is supported.
                 Supported keywords are: curator, gender, nationality, occupation,
                 country, outcome, caseid, uploadid, sourceurl, verificationstatus, 
-                admin1, admin2, admin3
+                admin1, admin2, admin3, variant
       responses:
         "200":
           $ref: "#/components/responses/200CaseArray"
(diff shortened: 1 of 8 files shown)


iamleeg:
I'm going to assert that this is a self-contained pull request even though it doesn't completely implement #1502, because doing that would require the whole map VOC visualisation to be added before this can be completed. If we land this PR now, then users can start using the VOC fields to search the dataset (which curators can already fill in), and then we can add the map filtering once we have a UI for variants in the map.

calremmel (approved):
LGTM! I like the * syntax for the not null filtering. We can pull more folks into the VoC selection convo for the map.

PR #1514, opened 2021-02-04
remove stale packages from package-lock

iamleeg:
this fixes the problem with checking out `ssh://git@github.com/eligrey/FileSaver.js.git`


calremmel (approved):
So glad this is a quick one - thanks a ton Graham!

PR #1516, opened 2021-02-04
let me use colons in my git URL rewriting pls

iamleeg:


calremmel (approved):
Ah I see!!! Good move, thanks for this!

PR #1531, opened 2021-02-08
Fixed age filter, improved geocoding queries

fojackson8:
Peru was breaking on age filter I added last commit (to remove cases with age > 120). We no longer filter out these cases, just don't populate the Age field.

Second issue was mapbox geocoding may not have been accurate for certain places within Lima department. The Peru dataset has district, province and department location fields, and we have lots of cases with the following location query: "LIMA, LIMA, LIMA, Peru", which is possible as Lima is both a province and a department, and the capital of Lima province is Lima. I've accounted for this, but geocoding of Peru will need more QC


calremmel:
Looks really good! I'm curious - what error message were you getting when checking for age > 120? It looks like you were checking for existing age, which is great, which should have solved for values of None, for example. Are there ever strings in age?

Also, for entries with age > 120, do we feel like the rest of the data in those cases is reliable? Or reliable enough?

fojackson8:
I was previously checking whether Age < 120 for each case, and if not then filtering out that case. That breaks when Age==None.

For cases with age > 120, I think the rest of the data in these cases (only 2 of them) is pretty reliable. There is a larger issue of how reliable age is however - there are ~30 cases with reported ages over 110.

PR #1539, opened 2021-02-09
Germany geocoding error

AnyaLindstromBattle:
Issue was that German locations were only geocoding to admin1. Because of issues with German admin3 locations and mapbox previously identified by @calremmel I have added location information based on lat/long info from https://npgeo-corona-npgeo-de.hub.arcgis.com/datasets/esri-de-content::kreisgrenzen-2019

Note that the only locations not covered by this are the subdivisions within Berlin which are officially admin2 (not admin3) and so therefore no information for them was present in the source above. These will be geocoded using mapbox as Berlin, Germany.

### DIFF
--- a/ingestion/functions/parsing/germany/dictionaries.json
+++ b/ingestion/functions/parsing/germany/dictionaries.json
@@ -1,0 +1,2011 @@
+{
+    "name_id_dict": {
+        "1001": "Flensburg",
+        "1002": "Kiel",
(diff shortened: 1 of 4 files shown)


Mougk (approved):
these look good to me. thanks anya

calremmel (approved):
Nice solution, thanks for going to that length!

PR #1543, opened 2021-02-10
1537 Apply search filters automatically after delay

maciekz1996:
**Changes**

* When search query is typed into SearchBar filters are automatically applied after given duration (1 second for now)
* Created custom hook `useDebounce`
* Stopped setting SearchBar as disabled when data is loading


sergioloporto (approved):

calremmel (approved):
LGTM!

PR #1547, opened 2021-02-11
1488 large data ingestion

abhidg:
This PR adds large file ingestion support by saving temporary files to a EFS volume which is attached to the retrieval function.

**Note**: At the moment tests **do not** check whether the EFS is mounted properly, as I wasn't sure whether that would work in the Github CI environment. There is a `tempdir` parameter that controls where temporary files are saved that is set to `/tmp` for Github CI.


calremmel (approved):
This looks great! I like the solution for local testing.

Issue #1560, opened 2021-02-16
Fix ingestion issues relating to backfill


abhidg:
Various parsers fail on ingestion when trying to backfill by @AnyaLindstromBattle 

- [x] Czechia: chunking puts an extra 'd' in the second line which causes it to crash
- [ ] Data download at source stage fails for Argentina
- [x] Germany: ECONNREFUSED 10.100.59.189:80 error sending cases to server
- [x]  Brazil Rio Grande do Sul: Some chunks throw IndexError (see below)
- [ ] Mexico, Germany, Sao Paolo: Timeouts for some chunks
- [ ] Investigate uneven chunk sizes for Mexico
- [ ] RGDS ECONNRESET error
- [ ] USA fix ingestion script
- [x] re-ingest missing line list exported cases
- [x] update data dictionary and add to downloads
- [x] update data agg script to serve VOC data to map
- [x] tackle geocoding fix
- [ ] repair and improve error logger

Mougk:
@AnyaLindstromBattle can you try Mexico/Germany again to see whether there are timeouts for the chunks now that the chunk size has been reduced?

AnyaLindstromBattle:
I just tried backfilling Germany and got the following - 
![image](https://user-images.githubusercontent.com/62887048/108206672-9ce2d100-711e-11eb-90a2-59faac72b8d7.png)
Think the same as the permissions issue w Mexico?

calremmel:
**Status**

The ingestion infrastructure [currently works as follows](https://github.com/globaldothealth/list/tree/main/data-serving/data-service)(https://github.com/globaldothealth/list/tree/main/ingestion/functions/parsing):

1. A [retrieval function](https://github.com/globaldothealth/list/tree/main/ingestion/functions/retrieval) (AWS Lambda in Python) retrieves data (csv, json) from an external source and stores it in local filespace
2. It proceeds to divide it into chunks, uploads those chunks to S3, and then triggers simple [parsing functions](https://github.com/globaldothealth/list/tree/main/ingestion/functions/parsing) (also AWS lambda) on the chunks to mould them into a common format. These invocations are spaced out as much as possible in context of a 15 minute limit.
3. Each of these chunks are submitted to MongoDB Atlas via a [REST API written in node.js]().

**The Problem**

While the retrieval, chunking, upload to S3, and parsing are all working fine, upserting is not. Different submissions receive the follows error responses, which are a bit inconsistent:

1. socket hang up
2. ECONNREFUSED
[...]

PR #1562, opened 2021-02-16
Use correct header offset #1560

abhidg:
The Czechia parser was receiving a header with a spurious extra 'd'
in the second line which broke the parser.

This was happening due to incorrect calculation of the header
offset. Previously the header offset was calculated _after_
decoding, which is fine for ASCII or UTF-8, but in other encodings
there may be a BOM or in the Czechia case, additional characters
like \ufeff which made the offset off by one.

This commit correctly calculates the header offset at the byte
level before decoding.


calremmel (approved):
Makes sense! LGTM

PR #1566, opened 2021-02-17
Deal with age >120

AnyaLindstromBattle:
Validation error as Mexico had cases with age > 120


fojackson8 (approved):
Yep, looks good to me, fits well with recent Peru fix (#1531 )

PR #1569, opened 2021-02-18
Hard-code USA location

AnyaLindstromBattle:
using 'query: Unites States' in mapbox was returning Cloud, Cloud, Kansas. We want only United States to be the name.


fojackson8 (approved):
Solid change

PR #1576, opened 2021-02-19
Updated urls with production ones

maciekz1996:
Updated all the urls with production ones in Curator UI


calremmel (approved):
LGTM! Thank you Maciej

PR #1590, opened 2021-02-22
Deal with new age category Scotland parser

AnyaLindstromBattle:
Looks like they added a new age category which wasn't covered by the convert_age function


fojackson8 (approved):
nice one!

PR #1605, opened 2021-02-23
retrieval: Fix USA geocoding

abhidg:
Longitude was positive!﻿


calremmel:
Hahaha thanks @abhidg ! I special cased it for the aggregate export for the map just for today, but I can take that out once there is a chance to rerun. Amazing getting this in BTW

PR #1606, opened 2021-02-24
Temporary script for ingesting VoC data [Now ready for merge]

fojackson8:
Please don't actually merge this code, just use for one time ingestion then delete


fojackson8:
UPDATE: This can now be merged, as we've set a public URL for the voc data

iamleeg (approved):

ingestion/functions/parsing/variants/variants_test.py
```diff
@@ -0,0 +1,232 @@
...
+_SOURCE_URL = "placeholder_URL"
+
+
+class VariantsTest(unittest.TestCase):
+    def setUp(self):
+        self.maxDiff = 5000
+
+    def test_parse(self):
+        '''
+        Includes a row where province and district are unspecified, where it should return just
```
AnyaLindstromBattle, line 15: What does this mean?

AnyaLindstromBattle (approved):
Just had one question about a note in variants_test.py  but otherwise looks good.

ingestion/functions/parsing/variants/variants_test.py
fojackson8: Good catch was a copying error, now removed

Issue #1520, opened 2021-02-05
Exclude certain sources from the case list and data download


iamleeg:
**Is your feature request related to a problem? Please describe.**
We use certain data sources where our license/terms of use permit sharing aggregate data, but not individual records. Those sources should be excluded from the case counts, the line list in the portal and the curator-service API, but should still be used in calculating top-level metrics for the map visualisation.

There should be some explanatory text somewhere (@mougk to provide) explaining the situation and any disparity in numbers seen across the site.

We think the filtering should probably be based on source ID or URL so that we know any records from a specific source won't be included.

fojackson8:
Given that each case has an associated source_URL, presumably providing a list of source_URLs to be excluded would be a sufficient input here?

iamleeg:
Yes, indeed.

Mougk:
Suggested Text to be added: 

"Metadata from GISAID (https://www.gisaid.org/) are not available for download through this website and only aggregated counts are included in the visualisations. See data acknowledgements here: LINK TO DATA ACKS SECTION (https://dev-globalhealth.pantheonsite.io/ackowledgement/)"

Sound good?

iamleeg:
thinking of adding a toggle to the sources editor in the curator UI for "include cases from this source in the line list" and then working out everything else that needs to respect that toggle; will work on that until someone tells me why it's a bad idea :)

iamleeg:
This has been merged and deployed.

PR #1620, opened 2021-03-03
Filter sources

iamleeg:
Whew! This was quite the ride: implementing this change involved switching large chunks of the data service from `Collection.find()` to `Collection.aggregate()` because finding cases now involves a join on the sources collection to ensure you're allowed to expose this case.

I'm putting this up now because the approach basically works, and I just need to colour it in. So I'd like a review to ensure we all agree this is a good way to implement it.

 - [x] existing tests of the cases endpoint pass using the new implementation
 - [x] list endpoint uses new implementation
 - [x] test that cases which should be excluded actually are excluded

I have _not_ altered the upsert/delete operations, because then the parsers that work with excluded sources couldn't modify any records.

This PR addresses #1520.

### DIFF
--- a/data-serving/data-service/package-lock.json
+++ b/data-serving/data-service/package-lock.json
@@ -3679,7 +3679,6 @@
       "dependencies": {
         "anymatch": "~3.1.1",
         "braces": "~3.0.2",
-        "fsevents": "~2.1.2",
         "glob-parent": "~5.1.0",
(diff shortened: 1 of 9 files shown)


abhidg (approved):
Looks good to me! I'm not an expert on Mongoose, but the test cases seem pretty clear and they pass.

Issue #1608, opened 2021-02-25
Enable users to download filtered results


calremmel:
A critical next feature is to allow Data users to download the filtered results of their queries, rather than only the entire dataset. Currently, the download API endpoint times out on larger queries, and the download format it returns is out of date.

An ideal solution has the following requirements:

- [x] Users can download the results of small queries immediately.
- [x] Users can select from several different download formats.
- [x] Users can receive a link by email to download the results of larger queries.

There are multiple pieces to implementing such a solution. Here is an initial proposal for the work involved:

**UI**

- [x] Download modal presents 3 delivery options:
  1. Download up to {limit} rows immediately
  2. Download more than {link} rows through link delivered by email
  3. Download entire dataset immediately
- [x] Download modal allows choice of csv, tsv, and json formats
- [x] Large query users should be instructed to add info@global.health or appropriate to their contacts list or check spam so that they don't miss their download link

**Backend**

- [x] /download API endpoint is updated to return current data format
- [x] /download API endpoint returns a maximum of {limit} records, where limit is defined by max that can be returned quickly
- [x] /downloadLarge API endpoint submits query to be run in queue using AWS Batch
- [x] AWS Batch job should export and parse results in fashion identical to total data export job
- [x] Parsed results of large query should be saved compressed as temporary s3 object
- [x] Link to temporary s3 object should be sent to query user by email
- [x] Query results should expire and be cleaned up after 3-7 days

iamleeg:
> "/download API endpoint is updated to return current data format"

Is there documentation on the expected data format anywhere? The download endpoint gets its configuration from data-serving/scripts/export-data/case_fields.yaml so simple to update.

iamleeg:
OK I've thought through and designed the large-download-via-Batch API, and ended up not needing Batch :). I'd love some feedback on this idea @calremmel @abhidg @rkassa @sergioloporto @maciekz1996 to make sure I didn't overlook something.

## Strawman data flow

Let's follow the design where we use Batch, and see the implications.

### 1. curator places request for large download

This is done via the curator-service API at `/cases/downloadLarge?q=blah`.

### 2. this batch job triggered

curator-service submits a job to the batch queue. Now here's a problem: the batch queue is going to run _outside_ the authenticated user's context in curator-service, but on behalf of them. How do we ensure that the user is actually permitted to see the query results, and that someone isn't submitting jobs to spam someone's inbox with multi-gigabyte CSV files? Read on!

[...]

PR #1625, opened 2021-03-04
1608 download filtered results UI modal

sergioloporto:
#1608 

**Changes**
* User has the ability to download full dataset that is stored in S3 and updated daily (for now only in .csv)
* User has the ability to download filtered data instantly if applied query returns less than 10000 cases (csv, tsv and json)
* For bigger queries background worker will fetch the results and after compressing it upload to AWS S3 to newly created `covid19-filtered-downloads` bucket. The file will be automatically removed after 3 days. The user will receive an email containing pre signed url to this file (also valid for 3 days) allowing him to download the results. (csv, tsv and json)

Email that is being sent with the download link
Screen Shot 2021-06-01 at 11 47 12

Alert on data page informing user that the email was sent
Screen Shot 2021-06-03 at 10 37 48


data-serving/data-service/src/controllers/case.ts
```diff
@@ -74,6 +74,13 @@ export class CasesController {
      * Handles HTTP POST /api/cases/download.
      */
     download = async (req: Request, res: Response): Promise<void> => {
+        if (req.body.query && req.body.caseIds) {
+            res.status(400).json({ message: 'Bad request' });
+            return;
+        }
```
iamleeg, line 80: Nice, good to test this constraint before continuing

data-serving/data-service/src/controllers/case.ts
```diff
@@ -105,39 +112,105 @@ export class CasesController {
...
+                    })
+                    .limit(queryLimit);
+            } else {
+                matchingCases = await Case.aggregate(
+                    casesIgnoringExcluded,
+                ).collation({
+                    locale: 'en_US',
+                    strength: 2,
+                });
+            }
```
iamleeg, line 131: Not a big deal, but could be changed to:

```
let query:any = Case.aggregate(casesIgnoringExcluded)collation({locale: 'en_US', strength: 2, });
if (queryLimit) { query = query.limit(queryLimit); }
const matchingCases = await query;
```

verification/curator-service/api/src/workers/downloadAsync.js
```diff
@@ -14,64 +17,130 @@ try {
...
+
+                    resolve(url);
+                });
+            });
+
+            // Send signed url via email to user
+            const mailMessage = `
+            <p>Please visit the link below to download list of cases in response to your query <strong>${workerData.query}</strong>.</p>
+
+            <a href="${signedUrl}">Click here to download the data</a>
```
iamleeg, line 104: Nice, a much better way of handling the large files

iamleeg (approved):
Nice work, thank you!

data-serving/data-service/src/controllers/case.ts
maciekz1996: Yeah, the code will be cleaner, thank you for spotting that!

Issue #1607, opened 2021-02-25
Enable pagination to scale to larger data


calremmel:
Currently, the pagination system cannot handle the ingestion of 10mil+ more cases. The cause of this is MongoDB's process for aggregating and counting records. Returning 50 records from a source is not a problem, regardless of source size, but counting the number of records from a source can be.

Material table requires a total count in order to operate, so the count step cannot be skipped. It would also be ideal to return some kind of sense for users for the results of common queries.

**Proposed Solution (please suggest improvements!)**

- [ ] Create MongoDB collection with record counts for common queries (and no query), cached 1/day
- [ ] GET /cases api endpoint should return for total these numbers if the query is located in this collection
- [x] Otherwise, endpoint should limit results to reasonable limit (e.g. 5k, 10k) and table should display "of many" for total record count _if above that limit,_ while using limit internally. See this solution: https://stackoverflow.com/a/32808615

This issue is a prerequisite for ingesting large sources like USA.

maciekz1996:
@calremmel I definitely agree with your solutions above 👍  

Maybe for the bigger queries where we would limit the query internally and just display "of many" above the table we could allow the user to click on the "of many" or some other button like "see all" for example and then display a popup saying that because there are so many results this query can take a long time to execute. This way the user still has the ability to see all the data on Data page even from big data sources and he also knows that our app is not stuck, just loading big query. 

Of course it all depends on how long those queries would take for data sources as big as ours, if it's up to a minute or so then I think it would work, but longer than that it may not be such a good idea.

Let me know what you think about this idea

maciekz1996:
@calremmel I was researching this issue and found out that currently we are using `Model.countDocuments(query)` in `data-serving/data-service/src/controllers/case.ts (line 788)` which as you mentioned scans the entire collection in order to count the documents. There is however another function `Model.estimatedDocumentCount()` that scans collection metadata instead of the whole collection itself and as suggested in MongoDB Docs it is better suited for getting the number of documents in large collections. I think this could potentially resolve this issue when there are no filters applied (getting the whole cases collection) 🤔 

When the filters are applied there is unfortunately no way to use `estimatedDocumentCount()`

calremmel:
@maciekz1996 That sounds like a great solution! Since having no filters applied is such a common query, I think it is definitely worth a try. Perhaps we could try having the endpoint use `Model.estimatedDocumentCount()` when there are no filters, and `Model.countDocuments(query)` otherwise?

[...]

PR #1626, opened 2021-03-04
1607 scale pagination

maciekz1996:
#1607 

⚠️ Before pushing to production this solution should be tested on larger data source to see if the performance is good enough

**Changes**
* Added `count_limit` param to `GET /api/cases` query 
* Document count is being limited do `count_limit` param making the queries faster and Linelist pagination displays 'of many' as total results count
* Updated `openapi.yaml`

### DIFF
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -52,6 +52,7 @@ paths:
                       type: integer
                       format: int32
                       minimum: 100
+                      default: 10000
                       maximum: 50000
                 - name: q
                   in: query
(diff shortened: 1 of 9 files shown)


sergioloporto (approved):

iamleeg:
Tried with 1.1M cases in a local db, does seem perceptibly faster :+1:

iamleeg:
Unfortunately I think you'll need to rework the back-end changes before this can merge :(. I switched from `collection.find()` to `collection.aggregate()` in #1620 to support excluding the GISAID data, so there isn't a count query any more it's a phase in the aggregation pipeline.

iamleeg (changes requested):
Needs to integrate with the aggregation pipeline, sorry :(

maciekz1996:
@iamleeg Ok, I will work on that 👍

iamleeg (approved):
Thank you!

Issue #1622, opened 2021-03-03
Make a Comprehensive Case Schema


fojackson8:
Cases are ingested into the database as JSON objects, which have a set of fields (keys) populated relating to case demographics, location, dates etc. The specific fields that are accepted as input are defined with a schema. We need a comprehensive documentation of this schema. 

We need a list of all accepted fields pertaining to a case. Each field has a key and a value. We need the precise spelling of the key, a description of what this field should contain, and the form of the expected value. Some fields are expecting a list, others a boolean, others accept any string. It would be great to have this explicitly stated. An example of a case with all fields correctly populated would be a big help.

We almost have this on [these API schema docs](https://curator.ghdsi.org/api-docs/#/Case). **This just needs to be updated and moved somewhere else**.

This may also be useful for responding to user questions.

iamleeg:
The canonical URL for the swagger doc is https://data.covid-19.global.health/api-docs/ though currently I'm having bad times trying to access it. Looking at the doc it pulls from I see that the variant info isn't there so I've [added that](https://github.com/globaldothealth/list/pull/1627), please comment on the PR if there's any other useful info missing :+1:

calremmel:
@fojackson8 @iamleeg Does this still need to be moved somewhere else, or are the updated Swagger docs sufficient?

If this needs to be accessible elsewhere, perhaps it could be matched with the data dictionary somehow. That might be confusing for folks who don't understand the difference between the MongoDB schema and the exported .csv format, but if it is of interest, perhaps you can discuss whether it ought to be a component in the UI, simply another file included with downloads, etc.

PR #1627, opened 2021-03-04
Add undocumented fields to the case schema

iamleeg:
Proposed fix for #1622, not sure if complete though so @fojackson8 / @AnyaLindstromBattle any other information missing?

### DIFF
--- a/verification/curator-service/api/openapi/openapi.yaml
+++ b/verification/curator-service/api/openapi/openapi.yaml
@@ -460,7 +460,7 @@ paths:
                     application/json:
                         schema:
                             allOf:
-                                - $ref: '#/components/schemas/Case'
+                                - $ref: '#/components/schemas/NewCase'
                             required:
                                 - caseReference
             responses:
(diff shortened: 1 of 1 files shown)


calremmel (commented):
Certainly looks good to me! I think that this is the only new field aside from `limitToCountry` which is in there under POST /cases. Anya/Felix please feel free to correct.

iamleeg:
`limitToCountry` is actually documented but I can add examples, and there are parts of the location (geometry, country, geo resolution) that need additions. If you don't see `limitToCountry` documented in the swagger UI, perhaps it isn't getting updated on deployment?

iamleeg:
Ah, got it: the location fields are documented for PUT but not for POST.

fojackson8:
A lot of the VoC data has number of cases > 1. Is number of cases a documented field here?

iamleeg:
@fojackson8 number of cases is already documented for POST (i.e. create new cases), where it works, but not for PUT (i.e. upsert potentially new or existing cases) where it isn't available.

fojackson8 (approved):
I think that's everything, thanks for adding!

PR #1632, opened 2021-03-12
Add variant field to CSV export

iamleeg:
Putting this up as a straw man fix to "/download API endpoint is updated to return current data format" part of #1608, because I don't think I understand what the full implications of that are and can get told when people flame me in code review :)


abhidg (approved):
Looks good to me!

iamleeg:
The file has been deleted in main, so I'm abandoning this.

PR #1639, opened 2021-03-18
985 add ability to filter cases by dates

sergioloporto:


verification/curator-service/ui/src/components/App.test.tsx
```diff
@@ -189,4 +189,47 @@ describe('<App />', () => {
...
+
+            // fireEvent.click(screen.getByText('Confirmed date'));
+
+            console.log('theSearchField2222222', theSearchField);
+
+            expect(getByRole('textbox', { name: /searchbar/i })).toHaveValue(
+                'gfifykfhkgfkhfk',
+            );
+        });
+    });
```
sergioloporto, line 234: @rkassa @scarpino Hello, I'd like to ask you or James if you can take a look at it at your best convenience.
I am trying (in various ways) to:
-check if the searchbox is empty
-check if it has a specific value

However the tests always pass. No matter if the search textfield is empty, and I tell Jest to experct a random string that for sure is not there. Test always passes. I am wondering if there is a problem with my code, or with something else.

sergioloporto: resolved already

iamleeg:
I'm confused by the test failures. They don't appear to be new, but I'm trying to bisect on main and haven't found the first good commit yet going back to November O_O. The tests that fail seem fairly fundamental so I can't believe that they've _always_ been broken, but I haven't isolated what introduced the failure yet.

iamleeg:
> I'm confused by the test failures. They don't appear to be new, but I'm trying to bisect on main and haven't found the first good commit yet going back to November O_O. The tests that fail seem fairly fundamental so I can't believe that they've _always_ been broken, but I haven't isolated what introduced the failure yet.

I'm no longer confused by this, it's not a problem on this PR :)

data-serving/data-service/src/controllers/case.ts
```diff
@@ -91,14 +91,11 @@ export class CasesController {
             res.setHeader('Pragma', 'no-cache');
             axios
                 .get<string>(
-                    'https://raw.githubusercontent.com/globaldothealth/list/main/data-serving/scripts/export-data/case_fields.yaml',
+                    'https://raw.githubusercontent.com/globaldothealth/list/main/data-serving/scripts/export-data/functions/01-split/fields.txt',
                 )
                 .then((yamlRes) => {
                     const dataDictionary = yaml.safeLoad(yamlRes.data);
```
maciekz1996, line 97: `dataDictionary` variable should be removed as it is no longer used

iamleeg (approved):
Looks great, and thanks for patching up the tests!

maciekz1996 (changes requested):
Filtering by date works great, well done 🥇 I just noticed that the search guide is not updated to reflect newly added filtering keywords, and also I commented on one file where one variable declaration can be deleted. 

Screen Shot 2021-03-25 at 14 43 51

PR #1640, opened 2021-03-21
Include country field in location information

AnyaLindstromBattle:
Added required country field to location information.


AnyaLindstromBattle:
e2e testing passes but note that 'PANGOA, SATIPO, JUNIN' indicated as unknown location so may want to add it to dict @fojackson8

fojackson8 (approved):
LGTM

PR #1644, opened 2021-03-25
No location in travel history

iamleeg:
This unblocks the VoC ingestion because not all the VoC reports have detailed travel history, but it was decided that we should keep as much incomplete info as possible.


AnyaLindstromBattle (approved):
Looks good to me. Any reason it's failing tests?

iamleeg:
Yes, the tests were already broken on main. @sergioloporto fixed in his PR, and I have merged and rebased that PR so this should run cleanly, then I'll merge.

PR #1645, opened 2021-03-25
1297 sorting linelist view

maciekz1996:
**Changes**
* Added Select input that let's user to choose which field to sort by
* Added sorting order select input
* Added `sort_by` and `order` parameters to GET `/api/cases/` endpoint
* Updated and added tests

**Screenshots**
Screen Shot 2021-03-25 at 12 15 00
Screen Shot 2021-03-25 at 12 15 17

### DIFF
--- a/verification/curator-service/api/test/clients/email-client.test.ts
+++ b/verification/curator-service/api/test/clients/email-client.test.ts
@@ -16,19 +16,19 @@ describe('initialize', () => {
 
 describe('send', () => {
     // This email service is not currently in use
-    // it('sends specified properties via email', async () => {
-    //     jest.setTimeout(15000);
+    it.skip('sends specified properties via email', async () => {
+        jest.setTimeout(15000);
 
-    //     const client = await new EmailClient().initialize();
-    //     const address = 'foo@bar.com';
-    //     const result = await client.send(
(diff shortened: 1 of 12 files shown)


sergioloporto:
Looks great to me. Tested different filter combinations and pagination and works fine

sergioloporto (approved):

verification/curator-service/api/test/clients/email-client.test.ts
```diff
@@ -15,19 +15,20 @@ describe('initialize', () => {
 });
 
 describe('send', () => {
-    it('sends specified properties via email', async () => {
-        jest.setTimeout(15000);
+    // This email service is not currently in use
```
iamleeg, line 18: I suggest not commenting the test and using `it.skip` so it's documented in the test output that we've disabled this. It's a trivial change and I'll push the commit myself.

iamleeg:
Looks good, do you have an idea about the performance of sorting? It's been my experience that mongodb is _much_ faster for sorted results if it has appropriate indexes, and wonder if they need adding in this case.

maciekz1996:
@iamleeg I've added one index for age field in mongodb the rest of them were already there

PR #1648, opened 2021-03-25
Use mkstemp() instead of NamedTemporaryFile

abhidg:
NamedTemporaryFile has some issues on Windows:
https://bugs.python.org/issue14243


AnyaLindstromBattle (approved):
Looks great thanks for taking this on.

PR #1649, opened 2021-03-27
Mystery of the vanishing cases

abhidg:
The case: Spreadsheet imported cases have been mysteriously disappearing in the middle of the night, from the streets of MongoDB. We thought the deleter had gone away for good, but then a recent spate of disappearances showed that they might still be lurking somewhere in the codebase.

Solution: (Really hoping it's this, the evidence is enough for a solid case, but who knows what accomplices they have?)

https://github.com/globaldothealth/list/blob/main/.github/workflows/case-data-update-dev.yml
sets up MongoDB and tries to get the latest spreadsheet data (convert_and_import_latest_data.sh)

This file fetches the latest data and converts it. Presumably this fails somewhere, but then the script goes on to setup-db, which by default deletes all cases where importedCases exists:
https://github.com/globaldothealth/list/blob/966454b84b8d331f07c44b411011487d6ff985e2/data-serving/scripts/setup-db/src/index.ts#L47

As for why this only occurred sometimes, the workflow is triggered by any update to the schema which happened yesterday and before that, on 4 Feb 2021 (https://github.com/globaldothealth/list/actions/workflows/case-data-update-dev.yml)

Removing this workflow should fix this, but we should think about archiving this code path entirely.


abhidg:
Huh. I think this _does_ solve why the cases disappeared on 4 Feb (the 23m runtime), but the script no longer exists in the repository, and the most recent run actually failed. So what caused the latest disappearance?

iamleeg:
WAT

iamleeg:
Ok, that script _should_ be running only in dev (it uses the dev mongo URL secret). If we are saying that it deleted data in production, then we are saying that secret is misconfigured, so:

1. does anything else use it?
2. is the appropriate course to delete the scripts or to correct the secret, or both?

abhidg:
This is failing on -dev, and is no longer needed anyway as we are not using dev to view the old linelist data, merging.

PR #1760, opened 2021-04-22
1608 batch download

iamleeg:
There are still open questions on this:

 - I can't see how I'd add automated tests (I've been testing by making API calls to the curator service, and checking my inbox) and I'd welcome guidance from reviewers.
 - SES has an attachment limit of 10MB, how do we want to deal with that?
 - Is the approach to error reporting sufficient?
 - How will the extra configuration parameters (SES region and sender address) get populated in k8s?

### DIFF
--- a/verification/curator-service/api/src/controllers/cases.ts
+++ b/verification/curator-service/api/src/controllers/cases.ts
@@ -84,10 +84,12 @@ export default class CasesController {
      downloadAsync = async (req: Request, res: Response): Promise<void> => {
         try {
             const user = req.user as UserDocument;
+            const url = this.dataServerURL + '/api' + req.url.replace('Async', '');
             const worker = new Worker('./src/workers/downloadAsync.js', { 
                 workerData: {
                     query: req.body.query as string,
-                    email: user.email, // work out how to get the user's email
+                    email: user.email,
+                    url,
             }});
             /* we don't care what happens when the worker finishes, but for debugging
              * I'm going to log this. */
--- a/verification/curator-service/api/src/controllers/cases.ts
+++ b/verification/curator-service/api/src/controllers/cases.ts
@@ -85,14 +85,23 @@ export default class CasesController {
         try {
             const user = req.user as UserDocument;
             const url = this.dataServerURL + '/api' + req.url.replace('Async', '');
+            // the worker needs access to the AWS configuration.
+            const region = process.env.AWS_SERVICE_REGION;
+            const accessKeyId = process.env.AWS_ACCESS_KEY_ID;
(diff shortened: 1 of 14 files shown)


verification/curator-service/api/src/controllers/cases.ts
```diff
@@ -73,6 +74,47 @@ export default class CasesController {
...
     };
 
+    /** DownloadAsync spawns a worker thread that forwards the request to the data service
+     * and emails the streamed response as a csv attachment. It will only email to the
+     * logged-in user, and it replies immediately with 204. Any error will be reported
+     * in the email, except the error of not being able to email which we can only log.
+     * 
+     * Question for code reviewers: do we want to retry on errors? If so we need to introduce
+     * a message bus rather than just using background workers, and we're probably back at the
+     * point where we use Batch for this.*/
```
jim-sheldon, line 84: I don't know if retrying on errors would provide much benefit, i.e. why wouldn't retrying simply run into the same error. That said, as the data set and user base grow, it might make more sense to use some combination of AMQP/SQS and Batch so we don't overload services.

jim-sheldon:
> There are still open questions on this:
> 
>     * I can't see how I'd add automated tests (I've been testing by making API calls to the curator service, and checking my inbox) and I'd welcome guidance from reviewers.

We could add [localstack](https://github.com/localstack/localstack) to a docker-compose file for this. We'd need to add a script to make the request to the service, and some entries in .env too.
I'd advocate for making that its own ticket/PR.

>     * SES has an attachment limit of 10MB, how do we want to deal with that?
For large files we could upload them to S3 and email the requester a presigned URL
https://docs.aws.amazon.com/AmazonS3/latest/userguide/ShareObjectPreSignedURL.html

>     * Is the approach to error reporting sufficient?
I think so, though maybe better to use `logger` than `console.log`?

>     * How will the extra configuration parameters (SES region and sender address) get populated in k8s?
I think we add them via [kubectl](https://kubernetes.io/docs/concepts/configuration/secret/) and then point to them in aws/curator.yaml.

jim-sheldon (approved):

abhidg (approved):
Looks good to me! +1 on using S3 for larger attachments as @jim-sheldon said.

sergioloporto (approved):

iamleeg:
Thanks everyone! I'll take the approach @jim-sheldon outlined: I'll make config changes ready for this to be available in dev/prod, then merge, then make issues for the remaining tasks.

> I think so, though maybe better to use logger than console.log?

Good observation, thank you, I'll make that change before merging.

PR #1761, opened 2021-04-22
Merge frontend UI work for 1.6 pre-release

abhidg:


iamleeg:
Test failures look to be a problem in rebase (though I don't recall having to manually merge that file at any time), as the tests look like they don't take into account known changes in the prod code.

iamleeg:
Looks like there was one test failing because I'd merged the implementation badly, and one test failing because I'd merged the test badly. So I'm happy to approve the PR when CI succeeds but ideally somebody else (@jim-sheldon, are you available?) should give it a once-over too because I'm marking my own homework here.

iamleeg (approved):

data-serving/data-service/test/util/search.test.ts
```diff
@@ -72,7 +74,7 @@ describe('search query', () => {
                 },
                 {
                     path: '_id',
-                    values: ['abc123'],
+                    values: [new ObjectId('605c8f6a7ee6c2d7fd2670cc')],
```
iamleeg, line 77: This was a change I made to get the tests to pass. The fact that it _wasn't_ already part of the diff makes me suspect this test was just broken.

PR #1764, opened 2021-04-23
Added logging to data service method to list all cases

jim-sheldon:
Will reduce to `debug` once we figure out what the current issue is


abhidg (approved):
LGTM

PR #1773, opened 2021-04-28
1717 geocoding service

iamleeg:
Here's an implementation of the geocoding parts of the data-service, in a standalone Flask app.


geocoding/location-service/run_tests.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+ENV=`mktemp -d`
+python3 -mvenv "${ENV}"
+. "${ENV}/bin/activate"
+pip install -r requirements.txt
+ENABLE_FAKE_GEOCODER=YES pytest
+deactivate
```
jim-sheldon, line 8: Not a blocker but I think it'd be better to use Poetry for better dependency and virtual environment management

geocoding/location-service/src/app/main.py
```diff
@@ -0,0 +1,68 @@
...
+
+@app.route("/geocode/suggest")
+def suggest_geocode():
+    try:
+        return jsonify(suggester.suggest(request.args))
+    except ValueError:
+        return f"Bad limitToResolution value {request.args.get('limitToResolution', str)}", 422
+
+if __name__ == '__main__':
+    app.run(host='0.0.0.0', port=8080)
```
jim-sheldon, line 68: Might be better to make these param values come from environment variables

geocoding/location-service/src/app/main.py
```diff
@@ -0,0 +1,68 @@
...
+    access_token = environ['MAPBOX_TOKEN']
+    mongo_client = None
+    if 'DB_CONNECTION_STRING' in environ:
+        mongo_client = pymongo.MongoClient(environ['DB_CONNECTION_STRING'])
+
+    admins_fetcher = AdminsFetcher(access_token, mongo_client)
+    mapbox_geocoder = Geocoder(access_token, admins_fetcher)
+    geocoders.append(mapbox_geocoder)
+
+suggester = GeocodeSuggester(geocoders)
```
jim-sheldon, line 28: I think it would be better to move this to a combination of [config](https://flask.palletsprojects.com/en/1.1.x/config/#configuration-best-practices) and [blueprints](https://flask.palletsprojects.com/en/1.1.x/blueprints/)

geocoding/location-service/src/app/main.py
```diff
@@ -0,0 +1,68 @@
...
+def seed_fake_geocoder():
+    obj = request.json
+    fake_geocoder.seed(obj['name'], obj)
+    return ''
+
+
+@app.route("/geocode/clear", methods=['POST'])
+def clear_fake_geocoder():
+    fake_geocoder.clear()
+    return ''
```
jim-sheldon, line 57: I don't think we need these endpoints, better to substitute the fake in testing environments.

geocoding/location-service/src/app/main.py
```diff
@@ -0,0 +1,68 @@
...
+        return jsonify({
+            "status": "Unhealthy",
+            "reason": "Mapbox API token is not present"
+        }), 500
+    return jsonify({"status": "Healthy"})
+
+
+@app.route("/geocode")
+def geocode():
+    query = request.args.get('q', type=str)
```
jim-sheldon, line 42: Return a 400 `if not query`?

jim-sheldon (approved):
Looks good, made a couple suggestions

geocoding/location-service/src/app/main.py
iamleeg: Because they're material to how the tests work and I'm trying to avoid changing the test behaviour, I've enabled the endpoints only if `ENABLE_FAKE_GEOCODER` is requested.

geocoding/location-service/src/app/main.py
iamleeg: Good call thanks, done.

geocoding/location-service/src/app/main.py
iamleeg: My argument against is that this service is running in a container anyway, so the operator gets to choose what network and port it's listening on when they expose/map the port on the running container. That, for me, is the place where it should be configurable. I'm open to listening to the counterargument, though.

geocoding/location-service/run_tests.sh
iamleeg: OK, I've changed it to poetry in running the tests, building the docker image, and CI.

geocoding/location-service/src/app/main.py
iamleeg: Thanks! I've turned the fake geocoder into a blueprint that can be registered, that's a tidy way to do it. I haven't changed the config because AIUI you no longer set variables in the environment, you set variables in a file and then tell Flask where that file is through the environment. That would make the deployment for this service differ from the others, there may be benefit to that but at the moment I'm trying to keep this addition as minimally complex as possible.

PR #1813, opened 2021-04-29
1688 Undefined showing as verification status

maciekz1996:
**Changes**
* By default case verification status is shown as unverified in Curator Portal instead of undefined or excluded
* Added Verification status as dropdown in Filters modal


sergioloporto (approved):
works great

Issue #1005, opened 2020-09-14
Sorting in line list view


Mougk:
Would be great to include sorting in line list view.

I downloaded CSV and used sorting function in excel, R which worked fine but is just taking a lot longer. UX may be better having such a simple functionality in the line list view.

allysonjp715:
There is a bug in material-table around sorting with remote data. Hopefully it is fixed soon:
https://github.com/mbrn/material-table/issues/2322
https://github.com/mbrn/material-table/issues/2252

attwad:
We'd also need to add indexes in mongo for ALL kind of sorting we'd want to do (all columns currently shown in the linelist) because we've had OOMs in the past due to missing indexes (could not go to the last page for example as the server had to hold the whole dataset in memory in order to sort).

abhidg:
Seems like we don't have indexes, which is why #1765 happened. This was fixed in https://github.com/globaldothealth/list/commit/71145e387fdbb5f738f05fd66cdf243d12137118 but it's not proper sorting as it doesn't use the entire dataset, only the current view.

What's the best process to add indexes? I suppose it could be a one-time operation on dev and prod, but maybe also from CI? @iamleeg @jim-sheldon @maciekz1996

iamleeg:
Indices are currently defined in the `cases.index.json` and similar files in the backend services, and updated by GitHub workflows.

abhidg:
Correction: it appears that we do have indices, but sorting is still slow before https://github.com/globaldothealth/list/commit/71145e387fdbb5f738f05fd66cdf243d12137118

abhidg:
@maciekz1996 @sergioloporto Sorting works in most cases :tada: but sorting for USA (14.8m cases) or over the entire dataset (29m cases) fails with 504. Sorting for countries with fewer cases works fine.

iamleeg:
504 is gateway timeout so we may be able to reconfigure the ingress controller to wait longer, unless it's something upstream (like maybe mongodb atlas has a load balancer or ingress controller that's timing out waiting for mongodb, but that seems unlikely). That's only a short term fix until "make it faster", but "make it faster" sounds difficult.

abhidg:
Remaining bit in #1875

PR #1818, opened 2021-05-04
#1005 - Sorting in linelist table fix

maciekz1996:
**Changes**
* Modified aggregation pipeline when fetching cases to allow for sorting
* Added missing indexes to DEV database

⚠️ 
This PR was tested on DEV database (17.4M cases) and the sorting was working fine. Some queries were longer than others but the results were always presented. It should be tested on production database as well to see if the amount of data there won't impact the performance to a significant extent.

### DIFF
--- a/.github/workflows/update-dev-data-db-schema.yaml
+++ b/.github/workflows/update-dev-data-db-schema.yaml
@@ -1,0 +1,32 @@
+name: Update dev data-service database schema
+
+on:
+  push:
+    # Run workflow when the schema changes.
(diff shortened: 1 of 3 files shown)


sergioloporto (approved):
works fine o:

DOCUMENTS 28.6m
TOTAL SIZE
19.5GB
AVG. SIZE
730B

abhidg (approved):
Thanks! Looks good to me, will deploy to prod and check once the tests run

Issue #1529, opened 2021-02-06
Unstructured queries on Data do not always show relevant results


joe-brilliant:
E.g when searching for "Turkey" (with no filter paramaters) the search returns 11 results, none of which are relevant to Turkey.

AAU, I would it expect this search to at least return all cases with Turkey in the country field (according to Map: 59,779 cases)

https://dev-curator.ghdsi.org/cases?q=turkey

https://www.dropbox.com/s/vzo84p4ahf7s0fc/Screen%20Shot%202021-02-06%20at%201.57.04%20PM.png?dl=0

elvey:
Umm... this isn't fixed.  E.g when searching for "Turkey" (with no filter paramaters) the search STILL returns 11 results, none of which are relevant to Turkey.  Same at https://data.covid-19.global.health/cases?q=turkey

elvey:
I understand why you opened #1836 (and that the 11 that I do see come up because "Turkey" is within the Notes field).  But is joe-brilliant not right that it should at least return all cases with Turkey in the country field (according to Map: 59,779 cases)?
<edit - yay - fixed>

PR #1821, opened 2021-05-05
#1781 - DataGuideDialog content update

maciekz1996:
#1781 

**Changes**
* Updated the content of SearchGuide in Data Portal
* Renamed SearchGuideDialog component to DataGuideDialog

**Screenshots**
Screen Shot 2021-05-05 at 13 13 07

### DIFF
--- a/verification/curator-service/ui/src/components/DataGuideDialog.test.tsx
+++ b/verification/curator-service/ui/src/components/DataGuideDialog.test.tsx
@@ -1,7 +1,7 @@
 import React from 'react';
 import { fireEvent, render, wait } from '@testing-library/react';
 import { Button } from '@material-ui/core';
-import SearchGuideDialog from './SearchGuideDialog';
+import DataGuideDialog from './DataGuideDialog';
 
 const SearchGuideTestCase = ({ defaultOpen = false }): JSX.Element => {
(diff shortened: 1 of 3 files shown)


sergioloporto (approved):
all good

abhidg:
Thanks! I hadn't reviewed the text before the PR - but we aren't using keywords anymore right? I would reword the first sentence of the last paragraph to: "For full-text search, enter any combination of search terms." I'm not sure what the full text search is doing at the moment (#1529 is still open). @joe-brilliant @Mougk thoughts?

joe-brilliant:
Sounds great to me @abhidg !

maciekz1996:
@abhidg @joe-brilliant Full-test search is currently searching through cases in mongodb based on `text indexes` on several fields in case schema. I've just checked on https://dev-curator.ghdsi.org/cases?q=Turkey and indeed all the results are connected to Turkey. For example other cases that are retrieved from this search query which are not from country Turkey contain Turkey in some fields in case details like for example `Travelled from Turkey` etc.

So in my opinion we can close #1529 as the results are correct, both on production and dev. 

I will also replace the content of data guide as @abhidg suggested

abhidg (changes requested):
Looks good, just the full-text search para needs updating

maciekz1996:
@abhidg I've updated data guide content

PR #1830, opened 2021-05-06
1731 curator eventbridge integration

jim-sheldon:


verification/curator-service/api/src/clients/aws-events-client.ts
```diff
@@ -1,8 +1,12 @@
 import AWS from 'aws-sdk';
-import AwsLambdaClient from './aws-lambda-client';
+import AwsBatchClient from './aws-batch-client';
 import assertString from '../util/assert-string';
 import { logger } from '../util/logger';
 
+const jobRoleArn = "arn:aws:iam::612888738066:role/gdh-ingestion-job-role";
+const jobQueueArn = "arn:aws:batch:us-east-1:612888738066:job-queue/ingestion-queue"
+const eventRoleArn = "arn:aws:iam::612888738066:role/service-role/AWS_Events_Invoke_Batch_Job_Queue_1312384119"
```
iamleeg, line 8: Would these be better as configuration?

jim-sheldon: Good call, will move them.

verification/curator-service/api/src/clients/aws-batch-client.ts
```diff
@@ -0,0 +1,136 @@
...
+    ): Promise<RetrievalPayload> => {
+        try {
+        	const r = await this.batchClient
+                .describeJobDefinitions({ maxResults: 10000 })
+                .promise();
+
+            var jobName = '';
+            jobDefsLoop:
+	          	for (var jd of r.jobDefinitions || []) {
+	        envVarsLoop:
```
iamleeg, line 66: you don't use this label.

verification/curator-service/api/src/clients/aws-batch-client.ts
```diff
@@ -0,0 +1,136 @@
...
+    doRetrieval = async (
+        sourceId: string,
+        parseDateRange?: ParseDateRange,
+    ): Promise<RetrievalPayload> => {
+        try {
+        	const r = await this.batchClient
+                .describeJobDefinitions({ maxResults: 10000 })
+                .promise();
+
+            var jobName = '';
```
iamleeg, line 63: `var` in javascript is [weird](https://www.tutorialsteacher.com/typescript/typescript-variable). I recommend `let jobName = ''` (*) and `const` for the loop variables which never get reassigned.

(*) well, almost. in fact you don't read the initial value except to see if it changed, so you could `let jobName; … if (!jobName)` and typescript will infer that it's `string | undefined` _and_ will infer that it isn't `undefined` after the if condition down there.

iamleeg (approved):
Looks great, made a couple of TS style comments but nothing blocking merge.

verification/curator-service/api/src/clients/aws-batch-client.ts
jim-sheldon: Thanks, removed.

verification/curator-service/api/src/clients/aws-batch-client.ts
```diff
@@ -0,0 +1,135 @@
...
+
+	        if (!jobName) {
+	        	throw Error(
+	        		`Could not find jobDefinition for sourceId ${sourceId}`
+	        	);
+	        }
+
+            const res = await this.batchClient
+        		.submitJob({
+        			jobDefinition: jobName + '-' + this.serviceEnv,
```
abhidg, line 86: Shouldn't this be `jobName`? It's already the job definition name

abhidg (commented):
Looks good! I need a way to get from `japan_ingestor` to `parsing.japan.japan`, since retrieval calls out to the parsing module. Should we add a field? I think we decided that ingestors would be like `japan-japan_ingestor`, in which case I can map it to `parsing.japan.japan` as a 1:1

verification/curator-service/api/src/clients/aws-batch-client.ts
jim-sheldon: Good call, fixed.

verification/curator-service/api/src/clients/aws-batch-client.ts
```diff
@@ -0,0 +1,135 @@
...
+            let ret = {jobName: res.jobName};
+            return ret;
+        } catch (e) {
+            logger.error(e);
+            throw e;
+        }
+    };
+
+    /** Lists the available parsers */
+    listParsers = async (): Promise<BatchJobDefinition[]> => {
```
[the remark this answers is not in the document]
jim-sheldon: Yeah, this gets used by `ui/src/components/ParsersAutocomplete.tsx`.
When it lived in the lambda client it filtered on functions with `FunctionName?.includes('ParsingFunction')`.
Rather than filter on "ingestor", would it be better to filter on "${env}"?

verification/curator-service/api/test/clients/aws-batch-client.test.ts
```diff
@@ -0,0 +1,122 @@
...
+} from '../../src/clients/aws-batch-client';
+
+import AWS from 'aws-sdk';
+import AWSMock from 'aws-sdk-mock';
+
+let client: AwsBatchClient;
+const submitJobSpy = jest.fn();
+const describeJobDefinitionsSpy = jest.fn();
+
+const ingestorName = 'japan_ingestor';
```
abhidg, line 13: Should be `japan-ingestor`

jim-sheldon: Good catch. Fixed.

abhidg (changes requested):
Thanks! Could you also update https://github.com/globaldothealth/list/blob/2a15beb2dd21d147f7f24e9032a2bb8834e3f462/ingestion/functions/retrieval/retrieval.py#L191 ? Currently it expects values of subfolder.name.

jim-sheldon:
> Thanks! Could you also update
> 
> https://github.com/globaldothealth/list/blob/2a15beb2dd21d147f7f24e9032a2bb8834e3f462/ingestion/functions/retrieval/retrieval.py#L191
> ? Currently it expects values of subfolder.name.

For sure! I found and updated it.

iamleeg:
Do love PRs with so much red

il_fullxfull 1237748257_arwo

abhidg (approved):

jim-sheldon:
Moving field renaming to [another issue](https://github.com/globaldothealth/list/issues/1881) until we get migrations. Will leave the database field as "awsLambdaArn" for now.

PR #1835, opened 2021-05-07
1784 Data - Sorting resets to "None" after exiting the window of "View case"

maciej-zarzeczny:
#1784 

**Changes**
* Moved `sortBy` and `sortByOrder` state variables from `LinelistTable` to `App`
* Moved test ensuring that sorting is working from `react-testing-library` to `cypress`


sergioloporto (approved):
it works great

abhidg (approved):
Tested locally, works. Thanks!

Issue #1846, opened 2021-05-11
Fix Lambda timeout on dataset compression in data export


abhidg:
Data export is broken due to `sam deploy` failing

abhidg:
We have now hit AWS Lambda 15 min limit, so have to move this to Batch as well :/

abhidg:
OK, so I've come across a solution, which will address the issue of large data files as well (the current unzipped dataset is 10 GB!). Thoughts @iamleeg @jim-sheldon @Mougk @joe-brilliant?

Previously we had this structure:
```
latestdata.tar.gz/
├── citation_data.rtf
├── data_dictionary.rtf
└── latestdata.csv
```

I'm proposing to move to this:

```
latestdata.tar/  # no gzip, files are compressed inside
├── citation_data.rtf
├── data_dictionary.rtf
└── cases-1-of-N.csv.gz
└── cases-2-of-N.csv.gz
.
.
└── cases-N-of-N.csv.gz
```

This has the advantage of being
* Easy to implement (just modify 03-parse to write a csv.gz, and 04-combine to output a tar)
* Should scale much better due to concurrency
* Individual files are smaller making them easy to open

jim-sheldon:
Short-term, I agree.

Longer-term, moving to Batch and keeping the original format seems preferable; decompressing and opening many files seems onerous for a user.

abhidg:
Agree on this being more onerous, though we would have to split CSVs anyway due to the large file size. I just did a `gzip --best` on a 60 MB cases-X-of-Y CSV which turned it into 2.5 MB, given that the main bottleneck after this change is downloading files to EFS (as we won't be compressing in combine), and 10 GB files took 4 minutes, that would give us 25 GB for 10 minutes (keeping 5 min for tarring it). This corresponds to ~10240 chunks, so more than a billion records!

iamleeg:
The only major question I have about splitting is whether to split by number (easy enough) or some other dimension like country. That could actually end up increasing the running time significantly.

Minor question is whether we can nudge people toward trying to do anything more sensible than opening tens of GBs in Excel. Could we host a mybinder and let people play with the data in pandas/R as an alternative or complement to downloading?

Mougk:
+1 on mybinder. Exploring data in R/pandas in a simple way would be great for users.

splitting by country or splitting by date both good. if so, names of files should either show country name or/and start/end dates.

abhidg:
[...]

PR #1854, opened 2021-05-13
Data export fixes

abhidg:
Fix data export

Data export ParsingFunction seems to have stopped after about 90% of the chunks.


abhidg:
OK, it seems like export service only supports one export per day. Deleted all files in S3 bucket and trying again..

iamleeg (approved):
Looks good, context manager should help prevent it getting into an inconsistent state again (although "NFS" and "consistent state" are not good bedfellows)

jim-sheldon (approved):

abhidg:
As per #1846, we have hit AWS Lambda limits, so this PR has to be reworked for Batch :/

PR #1855, opened 2021-05-14
Add basic data monitoring, fixes #1853

### LINKED ISSUE #1853: Notify when number of cases drop
This could be because of mistaken data deletion for example, or some other bug. We can use a monitoring script on the covid-19-aggregates bucket.

abhidg:


iamleeg (approved):
Did you look at the alert configuration available in mongodb atlas? I can see a chart that shows the prod cluster disk usage dropped by a GB twice in May, and it'd be useful to have these events at real time so we can try to triangulate what caused them. I don't think that obviates having our own monitoring as that can use domain-relevant knowledge, but it might be good in addition.

abhidg:
@iamleeg Will check the atlas logs, thanks! I'll try to get that to post to slack as well

PR #1860, opened 2021-05-17
export: Exclude sources that have excludeFromLineList, fixes #1857

### LINKED ISSUE #1857: Exclude sources in data export

abhidg:


iamleeg (commented):
Have you tested this? One potential issue that I don't know about is that where source IDs are ObjectIDs not strings, you may need to construct ObjectIDs in your query on app.py lines 50-52. But I wouldn't know without testing.

abhidg:
@iamleeg Yes, tested with one excluded source and works, `db.sources.find()` gives ObjectIDs, but `caseReference.sourceId` should be strings according to cases.schema.json

iamleeg (approved):
All good then!

PR #1861, opened 2021-05-17
Run through all case batches before reporting validation error

### LINKED ISSUE #1849: Data service batch upsert should not fail at first error
**Describe the bug**
Currently if there's an error inserting a case from ADI, all cases up to the erroneous one are inserted and no more. It'd be better if we could process all of the cases, and collect all of the errors to report back.

[One potential solution is to replace batch upsert by one-by-one upsert, and let the client collect the errors. This may be intolerable from a running-time perspective.]

iamleeg:
Fixes #1849. There's already a test that the error gets reported, so I know that's still working :)


abhidg (approved):
Looks good! We're not using the counter as a `Counter` so that can be a normal dict.

iamleeg:
I used `collections.defaultdict` because there are places where a key is looked up without being set and this means I can ensure the default value is still 0.

PR #1865, opened 2021-05-18
Parse validation errors to discover which input caused the error and …

### LINKED ISSUE #1850: On a case validation error, report the input that failed validation
**Describe the bug**
Due to batching, indexing and other things, when mongoose validation fails for an inserted/updated case it's not always easy to work out what input in the source data caused the error. It'd be really handy to reproduce the input into the error message so that you can see which document caused the failure.

Definitely P1 because it's blocking "launch" of the VoC data, could be bumped to P0 if others agree

iamleeg:
…attach it to the failure message.

Fixes #1850

### DIFF
--- a/ingestion/functions/common/parsing_lib.py
+++ b/ingestion/functions/common/parsing_lib.py
@@ -165,13 +165,29 @@ def write_to_server(
             counter['numCreated'] += res_json["numCreated"]
             counter['numUpdated'] += res_json["numUpdated"]
             continue
-        elif res.status_code == 207:
+        elif res and res.status_code == 207:
             # 207 encompasses both geocoding and case schema validation errors.
             # We can consider separating geocoding issues, but for now classifying it
             # as a validation problem is pretty reasonable.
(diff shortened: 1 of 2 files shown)


ingestion/functions/common/parsing_lib.py
```diff
@@ -165,13 +165,29 @@ def write_to_server(
...
             encountered_207 = True
-            validation_messages[f"batch_{batch_num}"] = res.text
+            # The errors from the backend tell us which cases failed and for what reason. Make it
+            # easier to diagnose by extracting the failing case and attaching it to the error message.
+            res_json = res.json()
+            if 'errors' in res_json:
+                def add_input_to_error(error):
+                    res = dict(error)
+                    res['input'] = batch[error['index']]
+                    return res
```
jim-sheldon, line 184: Better to `raise` an error message containing the failing case and error message?

abhidg: That's being done by `encountered_207` check in line 205

abhidg (approved):
Looks good!

Issue #1871, opened 2021-05-20
Remove dates from sorting and filtering


marlenachroscicka:

abhidg:
Fixed by #1874, kept filtering by date as that works for most cases.

PR #1874, opened 2021-05-21
1871 - Hide sorting by date

maciekz1996:
#1871 

**Changes**
* Hide sorting by confirmed date until fix is found to the timeout error


sergioloporto (approved):

joe-brilliant:
thanks @abhidg @maciekz1996 @iamleeg @sergioloporto do we have/need another issue to fix the timeout issue for large scale date sorting/filtering?

PR #1882, opened 2021-05-26
data,curator: Fetch latest in mongodb-memory-server

abhidg:
Always fetching the latest mongo fixes the issue where
mongodb-memory-server tries to fetch mongo and fails because
the version does not exist.

Using fix from mongodb-memory-server issue #316


jim-sheldon (approved):

abhidg:
OK, this reproduces the timeout on geocoding.

PR #1884, opened 2021-05-27
1807 show highlighted keywords from fulltextsearch

sergioloporto:
please check PR: https://github.com/globaldothealth/list/pull/1884

image

Also works on text chunks:
image


maciej-zarzeczny (approved):
It works well, good job 👍

maciej-zarzeczny (commented):
Overall it works well, I only think that the alert is not as visible as it should be due to it's background color. Also should this be a dismissible alert or perhaps just simple text? 🤔 

Screen Shot 2021-06-02 at 17 04 22

sergioloporto:
> Overall it works well, I only think that the alert is not as visible as it should be due to it's background color. Also should this be a dismissible alert or perhaps just simple text? 🤔
> 
> Screen Shot 2021-06-02 at 17 04 22

@maciekz1996  
Hmm, what background color do you suggest? It is not vital information, thus I thought for it to be not eye catching.
Dismissable... if somebody wants to dismiss it, the person can. If the person doesn't want to dismiss it, so doesn't click the X.

Please suggest if you think it's better to fix something

maciekz1996:
@sergioloporto 
Maybe something like breadcrumbs / chips (from Material UI) that are visible in LinelistView when some filters are active? 
It would be easier to understand in contrary to just plain text (especially when there will be more active filters)

maciekz1996:
@sergioloporto It looks better now in my opinion 👍  One thing that I noticed is that when the user filters the data insted of using SearchBar and Full text search then it is displayed as `country:India` for example. I think it would be better to parse it to `country - India` the same as on the main page when the filters are applied. Just to make things consistent across pages

Screen Shot 2021-06-04 at 15 41 51

maciej-zarzeczny (changes requested):
It works well for single word values, but falls apart for two worded values like United States for example. (See screenshots below)
Screen Shot 2021-06-07 at 15 59 36

maciekz1996:
@sergioloporto  Also I noticed that when two or more filters are applied there is no margin between chips
Screen Shot 2021-06-07 at 16 15 22

sergioloporto:
@maciekz1996 fixed

verification/curator-service/ui/src/components/App/App.tsx
```diff
@@ -380,9 +387,9 @@ export default function App(): JSX.Element {
     const lastLocation = useLastLocation();
     const history = useHistory();
     const location = useLocation<LocationState>();
-    const [filterBreadcrumbs, setFilterBreadcrumbs] = React.useState<
-        ChipData[]
-    >([]);
+    // const [filterBreadcrumbs, setFilterBreadcrumbs] = React.useState<
```
maciekz1996, line 390: You can delete unused code

verification/curator-service/ui/src/components/App/App.tsx
```diff
@@ -526,22 +533,28 @@ export default function App(): JSX.Element {
         // eslint-disable-next-line
     }, [savedSearchQuery]);
 
+    // const theBreadCrumbs = useSelector(selectFilterBreadcrumbs);
```
maciekz1996, line 536: Delete unused code

verification/curator-service/ui/src/components/Case.tsx
```diff
@@ -56,6 +56,7 @@ export interface Location {
     query?: string;
     // Optional to hint geocoding results.
     limitToResolution?: string;
+    theSearchArray?: any;
```
maciekz1996, line 59: Is this variable being used anywhere? If not then you can delete it

verification/curator-service/ui/src/components/ViewCase.tsx
```diff
@@ -233,6 +251,24 @@ function CaseDetails(props: CaseDetailsProps): JSX.Element {
                     marginLeft: showNavMenu ? '10em' : '0',
                 }}
             >
+                {searchedKeywords && (
+                    <>
+                        <Chip
+                            label="Searched query:"
+                            color="primary"
+                            onClick={() => console.log('clicked')}
```
maciekz1996, line 259: If `onClick` event is unused you can just delete it

maciej-zarzeczny (approved):
I've just checked and it works great, good job 🏅 I left some comments, but nothing major

abhidg (approved):

PR #1901, opened 2021-06-07
Use the same tag name in k8s that we tag images with in github

iamleeg:


abhidg:
I think @jim-sheldon changed it to latest since k8s apparently special cases it to always fetch the latest image rather caching which leads to problems. That will require the GitHub workflows to be changed to update it to latest. 

Or maybe there is a flag to tell k8s to treat main as latest?

abhidg (approved):

PR #1902, opened 2021-06-07
Try some voodoo

iamleeg:
Just pasting interwebs code at this point

### DIFF
--- a/geocoding/location-service/src/app/main.py
+++ b/geocoding/location-service/src/app/main.py
@@ -34,7 +34,10 @@ if 'MAPBOX_TOKEN' in environ:
     access_token = environ['MAPBOX_TOKEN']
     mongo_client = None
     if 'DB_CONNECTION_STRING' in environ:
-        mongo_client = pymongo.MongoClient(environ['DB_CONNECTION_STRING'])
+        # searching for 'pymongo can auth on command line but not in container' leads to suggestions
+        # to add the 'authSource' parameter on connection ¯\_(ツ)_/¯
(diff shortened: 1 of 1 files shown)


jim-sheldon (approved):

geocoding/location-service/src/app/main.py
```diff
@@ -34,7 +34,10 @@ def clear_fake_geocoder():
     access_token = environ['MAPBOX_TOKEN']
     mongo_client = None
     if 'DB_CONNECTION_STRING' in environ:
-        mongo_client = pymongo.MongoClient(environ['DB_CONNECTION_STRING'])
+        # searching for 'pymongo can auth on command line but not in container' leads to suggestions
+        # to add the 'authSource' parameter on connection ¯\_(ツ)_/¯
+        # e.g. https://www.reddit.com/r/learnpython/comments/m8ieht/am_trying_to_use_python_with_mongodb_and_am/
+        mongo_client = pymongo.MongoClient(environ['DB_CONNECTION_STRING'], authSource='admin')
```
abhidg, line 40: Could try this, but admin should be the default source (unless we are specifying defaultauthdb somewhere, which I can't find) according to https://docs.mongodb.com/manual/reference/connection-string/

iamleeg: It should be, I agree. I'm at a loss as to why this might work, but I'm at a loss as to why it doesn't already work: I can run the same code locally as in the cluster using the same connection string and get two different results.

PR #1919, opened 2021-06-09
1876 - Username and password auth

maciekz1996:
Some tests are skipped in UI and API for now. The reason for that is I changed email client from nodemailer to AWS SES and it requires some changes in the tests itself (different way of mocking email client). Sending emails works as before, just some tests broke. We don't want to hold applause testing longer, so we decided to skip those tests and work on fixing them simultaneously with Applause testing. 

**UI**
- [x] Create login form on landing page with error alerts (email, password, sign in button, signup button)
- [x] Create signup form with error alerts (email, password, repeat password, ...)

**API**
- [x] Create login endpoints
- [x] Create sign up endpoints
- [x] Add "Forgot password" functionality

**TESTS**
- [x] Fix failing tests
- [x] Create new tests (integration and E2E) for auth system


verification/curator-service/ui/src/components/landing-page/SignInForm.tsx
```diff
@@ -81,11 +81,14 @@ interface FormValues {
     isNewsletterChecked: boolean;
 }
 
-export default function SignInForm() {
+interface SignInFormProps {
+    setRegistrationScreenOn: any;
```
maciekz1996, line 85: Avoid using any in this case

verification/curator-service/ui/src/components/landing-page/LandingPage.tsx
```diff
@@ -82,6 +84,7 @@ interface CognitoUserExtended extends CognitoUser {
 
 interface LandingPageProps {
     setUser: (user: User | undefined) => void;
+    setRegistrationScreenOn?: () => void;
```
maciekz1996, line 87: You can remove `setRegistrationScreenOn` from `LandingPageProps` as it isn't used anywhere

verification/curator-service/ui/src/components/landing-page/SignInForm.tsx
```diff
@@ -131,11 +134,10 @@ export default function SignInForm() {
             formik.resetForm();
         };
         // eslint-disable-next-line
-    }, [registrationScreenOn]);
+    }, [setRegistrationScreenOn]);
```
maciekz1996, line 137: Why are you passing `setRegistrationScreenOn` as `useEffect` dependency? I think it is unnecessary

verification/curator-service/ui/src/components/landing-page/SignUpForm.tsx
```diff
@@ -0,0 +1,352 @@
...
+    email: string;
+    confirmEmail: string;
+    password: string;
+    passwordConfirmation: string;
+    isAgreementChecked: boolean;
+    isNewsletterChecked: boolean;
+}
+
+interface SignUpFormProps {
+    setRegistrationScreenOn: any;
```
maciekz1996, line 79: The same as above: avoid using any in situations when you know the type

verification/curator-service/ui/src/components/landing-page/SignUpForm.tsx
```diff
@@ -0,0 +1,352 @@
...
+}
+
+export default function SignUpForm({setRegistrationScreenOn}:SignUpFormProps) {
+    const classes = useStyles();
+console.log(setRegistrationScreenOn)
```
maciekz1996, line 84: Delete `console.log`

verification/curator-service/ui/src/components/landing-page/SignUpForm.tsx
```diff
@@ -0,0 +1,352 @@
...
+            console.log(values);
+        },
+    });
+
+    useEffect(() => {
+        return () => {
+            formik.resetForm();
+        };
+        // eslint-disable-next-line
+    }, [setRegistrationScreenOn]);
```
maciekz1996, line 132: The same as above: you can leave dependency of useEffect as empty []

verification/curator-service/ui/src/components/landing-page/SignInForm.tsx
```diff
@@ -81,11 +81,14 @@ interface FormValues {
...
+}
+
+export default function SignInForm({setRegistrationScreenOn}:SignInFormProps) {
     const classes = useStyles();
 
     const [passwordVisible, setPasswordVisible] = useState(false);
-    const [registrationScreenOn, setRgistrationScrenOn] = useState(false);
     const [forgotPasswordScreenOn, setForgotPasswordScreenOn] = useState(false);
 
     const validationSchema = Yup.object().shape({
```
maciekz1996, line 94: in this component you are using only `email`, `password` and `isAgreementChecked` in your form, so you can delete the rest from formik and `validationSchema`

verification/curator-service/ui/src/components/landing-page/SignInForm.tsx
```diff
@@ -298,231 +300,13 @@ export default function SignInForm() {
...
-                        >
-                            {' '}
-                            Sign in!
-                        </span>
-                    </Typography>
-                </form>
-            )}
 
             {/* {forgotPasswordScreenOn &&  
             <form onSubmit={formik.handleSubmit}>
```
maciekz1996, line 312: If this code is unused then delete it instead of commenting out

maciekz1996 (commented):
Apart from the comments in code you should also extract Forgot password Dialog to a separate component. Right now I saw that you are using the same formik for `SignIn` component as well as for `ForgotPassword` Dialog, you should avoid doing that. Extracting Dialog to a separate component will increase code readability

verification/curator-service/ui/src/components/landing-page/ChangePasswordForm.tsx
```diff
@@ -0,0 +1,231 @@
...
+export default function ChangePasswordForm() {
+    const classes = useStyles();
+    const [passwordVisible, setPasswordVisible] = useState(false);
+    const [
+        passwordConfirmationVisible,
+        setPasswordConfirmationVisible,
+    ] = useState(false);
+
+    const validationSchema = Yup.object().shape({
+        email: Yup.string()
```
maciekz1996, line 79: You can delete `email` and `confirmEmail` as you are not using those in this form

verification/curator-service/ui/src/components/landing-page/ForgotPasswordForm.tsx
```diff
@@ -0,0 +1,153 @@
...
+            .required('This field is required'),
+        confirmEmail: Yup.string().test(
+            'emails-match',
+            'Emails must match',
+            function (value) {
+                return this.parent.email === value;
+            },
+        ),
+        password: Yup.string().required('This field is required'),
+        passwordConfirmation: Yup.string().test(
```
maciekz1996, line 80: You can delete `confirmEmail`, `password` and `confirmPassword` as you are not using them in this form

verification/curator-service/ui/src/components/landing-page/LandingPage.tsx
```diff
@@ -92,8 +98,22 @@ export default function LandingPage({
     const smallHeight = useMediaQuery('(max-height:1050px)');
     const classes = useStyles({ smallHeight });
     const lastLocation = useLastLocation();
+    const location = useLocation<LocationState>();
 
     const [registrationScreenOn, setRegistrationScreenOn] = useState(false);
+    const [changePasswordScreenOn, setChangePasswordScreenOn] = useState(false);
+
+    useEffect(() => {
+        // temporary - to be changed while building api.
```
maciekz1996, line 107: Don't do it inside `useEffect` hook. It would be better to setup new route using `react-router`

iamleeg:
The integration test failure on FF in `AutomatedSourceForm` is a timing issue, and we've fixed it on a different PR: if you rebase it'll probably go away. But the other failure in `UsersTest` I haven't seen before so I don't know if that's the same/similar problem.

maciekz1996:
@iamleeg Thank you for your comment, we will take a look at that

verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -158,59 +207,117 @@ export class AuthController {
...
-                    req.body.newsletter
-                ) {
-                    user = await User.findOneAndUpdate(
-                        { email: req.body.email },
-                        { newsletterAccepted: req.body.newsletter },
+                try {
+                    // Check if user with this email address exists
+                    const user = await User.findOne({ email });
+                    if (!user) {
+                        throw new Error('User does not exist');
```
iamleeg, line 222: It's best not to make an error for this case, because it can be used to quickly see if another person has an account with us or to build a database of users at common domains like Gmail or live.com. More info here: https://cheatsheetseries.owasp.org/cheatsheets/Forgot_Password_Cheat_Sheet.html

By all means show an error if something goes wrong, but return 200 whether or not the user exists. The UI message should be something non-committal like "if you have an account with us, check your email for reset instructions."

maciekz1996: Got it, thanks for the clarification, I will change this tomorrow

verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -158,59 +207,119 @@ export class AuthController {
...
+                        token: hash,
+                        createdAt: Date.now(),
+                    }).save();
+
+                    const resetLink = `http://localhost:3002/reset-password/${resetToken}/${user._id}`;
+
+                    await this.emailClient.send(
+                        [email],
+                        'Password reset link',
+                        `To reset your password click <a href="${resetLink}">here</a>`,
```
abhidg, line 246: Won’t resetLink have to be adapted for dev/prod URLs?

verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -212,14 +212,14 @@ export class AuthController {
          */
         this.router.post(
             '/request-password-reset',
-            async (req: Request, res: Response): Promise<void> => {
+            async (req: Request, res: Response): Promise<Response<any>> => {
                 const email = req.body.email as string;
 
                 try {
                     // Check if user with this email address exists
                     const user = await User.findOne({ email });
```
iamleeg, line 220: The behaviour here means that if a user signed up with Gmail, they can still get a password reset link. I think that means that they'll get "converted" to a password user, when they use the link to reset their password. Is that expected? Will Google OAuth2 still work?

iamleeg (changes requested):
I think this looks OK overall, just marking for changes for two reasons:

1. @abhidg's point about the password reset URL
2. my question about Gmail users going through the password reset flow. I expect that if a user has a `googleId` they don't need to be sent a password reset, I'm not sure if that's correct.

iamleeg:
I thought some more about the Gmail users reset issue. Obviously we can't show an error in the browser "sorry you're registered with Gmail" because that leaks information about who is registered. But I think it's fine to send a different email to users with a registered googleId:

> We were asked to reset your password on global.health, but you are registered with your Google account. If you requested the password reset, then please try again, this time using the "Sign in with Google" button on the global.health portal. Otherwise, no further action is needed.

Maybe the wording could be better >_<. We shouldn't generate a token for people with a google ID.

verification/curator-service/api/src/controllers/auth.ts
maciekz1996: You're right, I totally forgot to change that, thank you for the catch!

maciekz1996:
@iamleeg You're right, reset token shouldn't be generated for Gmail users. I think that the approach with sending different email for those users would work well. I will go ahead and implement that

verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -158,59 +208,148 @@ export class AuthController {
...
+                        createdAt: Date.now(),
+                    }).save();
+
+                    let url = '';
+                    switch (env) {
+                        case 'local':
+                            url = 'http://localhost:3002';
+                            break;
+                        case 'dev':
+                            url = 'https://dev-curator.ghdsi.org';
```
abhidg, line 261: should be https://dev-data.covid-19.global.health

maciekz1996: Thanks, already updated

verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -158,59 +208,148 @@ export class AuthController {
...
-                        { newsletterAccepted: req.body.newsletter },
+                try {
+                    // Check if user with this email address exists
+                    const user = await User.findOne({ email });
+                    if (!user) {
+                        return res.sendStatus(200);
+                    }
+
+                    // Check if user is a Gmail user and send appropriate email message in that case
+                    if (user.googleID !== '42') {
```
iamleeg, line 227: The '42' is a hard-coded value from the test suite. I think the correct behaviour should be `if (user.googleID)` (it can't be `if (googleID in user)` because mongoose/mongo defines keys that are in the schema even if the document doesn't have a value).

verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -127,6 +133,50 @@ export class AuthController {
             },
         );
 
+        this.router.post(
```
iamleeg, line 136: Signing up as an email user fails for me when I try locally, and the error message isn't very helpful: it just says "Rejected". If I inspect the network request in the browser I see that posting to this endpoint got a 504 Gateway timeout error: what is it trying to talk to upstream when a user signs up?

verification/curator-service/api/src/controllers/auth.ts
maciekz1996: Fixed, thank you

verification/curator-service/api/src/controllers/auth.ts
maciekz1996: I couldn't reproduce this error. Both sign in and sign up work as expected on my local machine. Were you testing that while running the app with `dev/run_stack.sh` script?

iamleeg: No, neither can I now, might have been a configuration problem on my side. OK, this is good to go!

iamleeg (approved):
OK, looks great now! One tiny nit is that the terms and conditions button still says "By creating an account, I accept the Global.health Terms of Use and Privacy Policy *" even when you're signing in with an existing email account. Might need changing to "By using this service…" or something. Great work, thank you.

abhidg (approved):

Issue #1893, opened 2021-05-31
Disable the adding colons to searchBar by copy/paste


sergioloporto:
Users can now add colons to searchbar by copy/paste. This should be disallowed.

PR #1931, opened 2021-06-11
disallowed colon

sergioloporto:
Issie: https://github.com/globaldothealth/list/issues/1893


maciekz1996 (changes requested):
After pasting `country:india` into the search bar filters modal and error message appear, however after using some filters and searching the error message doesn't disappear (even though the search bar is already empty)

Screen Shot 2021-06-17 at 08 57 57

maciekz1996 (approved):
Works great! good job 💯

PR #1934, opened 2021-06-11
Ingestion dockerfile uses poetry

jim-sheldon:


ingestion/functions/Dockerfile
```diff
@@ -1,13 +1,72 @@
-FROM python:3.8  
+# `python-base` sets up all our shared environment variables
+FROM python:3.8.1-slim as python-base
```
abhidg, line 2: Is there a 3.8-slim? Might as well get the point updates automatically

jim-sheldon: Good point. Updated.

abhidg (approved):

PR #1936, opened 2021-06-12
Remove mapbox dependency from data service

abhidg:
Geocoding has been moved to its own service, so we don't need
mapbox dependency in data service.

Fixes: CVE-2021-33502


iamleeg (approved):
@abhidg as with the other PR I had to remove @types/mongoose to fix the build, then patch up some type errors (perhaps the API version in the @types definitions is different from the mongoose package we actually installed?). Check that you're happy with the changes I made before merging.

abhidg:
Looks good, thanks!

PR #1942, opened 2021-06-14
parsing: Fix crash with missing country code in Argentina

abhidg:
Argentina parser was crashing with missing 'corea republicana'.
Fix by adding entry to country code list, and fixes the error class
of missing country codes to log an error instead of crashing.

Fixes: #1922


jim-sheldon (approved):

iamleeg (approved):
Would be nice to add a test capturing this condition, looks good otherwise

fojackson8 (approved):
Looks good, thanks @abhidg

Issue #1921, opened 2021-06-09
Support data deletion when ingesting non-UUID sources


abhidg:
Previously, we were using an arbitrary cutoff (ingest only data 3 days prior) for non-UUID sources. We should ideally delete and re-ingest the data to avoid duplicates.

This will require support for the `batchDel` curator API endpoint to retrieval code.

Mougk:
Thanks.

Please see this doc for ADI setups for different dataset categories: https://docs.google.com/document/d/1VxLNFf_iMjwGd7cMNQ3blZlxnbeRgcRyiHxVIn6yquI/edit

iamleeg:
Use this on sources where we don't have `sourceEntryId`, e.g. Peru and USA.

iamleeg:
Don't let people specify a date range for these parsers, or run backfill. In fact, this points to a reorganisation of the source ID and scheduling: don't put the schedule information in the UI at all, except for a "schedule updates for this source" checkbox. You can only check that once the parser for the source has been supplied, and it runs the retrieval/parser to work out what category the source is in and sets the schedule accordingly.

Bonus points: instead of "delete cases on reimport", consider "diff the source and apply the diffs as actions to the parser", so removed lines get deleted, and added lines get imported.

iamleeg:
Short term: check box for "this source doesn't have UUIDs", which stops people running filtered backfills. Long term: the thing I described above. I will make that a separate issue.

PR #1946, opened 2021-06-15
1921 delete and reimport sources without stable ids

iamleeg:
This is the backend work for #1921, the parser lib work will come in another PR (assuming we don't decide to move away from the data-service before that happens!).


abhidg:
Looks good to me! Maybe we should have non-stable UUID as default, as that seems to be the majority of sources.

jim-sheldon (approved):

PR #1948, opened 2021-06-17
curator/api: Security update

abhidg:


iamleeg:
The main problem in the curator-service build is that the newer version of mongoose includes its own type definitions which conflict with those in `@types/mongoose`. Removing that leaves one error: I'll look into it and push fixes to the PR branch.

iamleeg (approved):
I'm happy with your changes @abhidg, you may wish to go over mine before merging.

PR #1950, opened 2021-06-18
Check in parser if the source has stable IDs and if not, delete the existing cases

iamleeg:
Mostly putting this up as draft now as it's the easiest way to get the dockerised tests to run, which are the only way to test the function I need to change.

#1921

### DIFF
--- a/verification/curator-service/ui/package-lock.json
+++ b/verification/curator-service/ui/package-lock.json
@@ -13,6 +13,7 @@
         "@material-ui/lab": "^4.0.0-alpha.56",
         "@material-ui/pickers": "^3.2.10",
         "@mongodb-js/charts-embed-dom": "^1.1.3",
+        "@reduxjs/toolkit": "^1.5.1",
         "@testing-library/jest-dom": "^4.2.4",
         "@testing-library/react": "^9.5.0",
         "@testing-library/user-event": "^7.2.1",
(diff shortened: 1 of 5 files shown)


abhidg:
Looks good, other than the default for hasStableIdentifiers which should be False as that's the case for majority. If a source actually has stable identifiers with hasStableIdentifiers, we are going to ingest the entire dataset needlessly, but it won't have incorrect data.

verification/curator-service/ui/package-lock.json
```diff
@@ -41094,6 +40963,20 @@
         "whatwg-fetch": "^3.0.0"
       }
     },
+    "react-beautiful-dnd": {
```
abhidg, line 40966: Is this part of another PR? Seems to be from the highlighting search terms work.

iamleeg: You know what? I don't know. I didn't explicitly put this here, I ran `npm i` to ensure the deps were up to date and I committed the change to the lock file. I can't find `react-beautiful-dnd` in the package.json so I can revert this file if you think it's wrong.

abhidg: Right, so the highlighting work seems to be already in main, which would explain this. I'm fine with keeping this change.
```
verification/curator-service/ui/src/components/ViewCase.tsx
30:import Highlighter from 'react-highlight-words'
```

abhidg (approved):

PR #1955, opened 2021-06-18
fix ingestion

abhidg:
- parsing: Default to has_stable_ids = True
- parsing: Pass headers and cookies to fix 403


iamleeg:
Didn't we want unstable IDs to be the default?

iamleeg:
Ah OK I read the comments.

Issue #1954, opened 2021-06-18
Production ingestion should use images tagged from -stable


abhidg:
Ingestion stable should track 1.X-stable.

- [x] Create a new stable tag for ingestion
- [x] Switch job definitions for prod to use :stable

PR #1960, opened 2021-06-28
Build ingestion image from stable branch #1954

abhidg:
The latest stable branch is used to build a :stable image
for gdh-ingestor. Also updates the job definition section
in aws.py to use stable images from prod environment.

### DIFF
--- a/.github/workflows/ingestion-functions-deploy.yml
+++ b/.github/workflows/ingestion-functions-deploy.yml
@@ -2,7 +2,7 @@ name: Ingestion functions deploy
 
 on:
   push:
-    branches: [main]
+    branches: [main, 1.6-stable]
     paths:
     - '.github/workflows/ingestion-functions-python.yml'
     - 'ingestion/functions/**'
@@ -25,6 +25,7 @@ jobs:
       uses: aws-actions/amazon-ecr-login@v1
 
     - name: Build, tag, and push image to Amazon ECR
+      if: ${{ github.ref == 'refs/heads/main' }}
       working-directory: ingestion/functions
       env:
         ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ingestion-functions-deploy.yml
```diff
@@ -2,7 +2,7 @@ name: Ingestion functions deploy
 
 on:
   push:
-    branches: [main]
+    branches: [main, 1.6-stable]
```
iamleeg, line 5: I think `*-stable` would be better, as we'll probably change the version at some point :)

.github/workflows/ingestion-functions-deploy.yml
```diff
@@ -34,3 +35,15 @@ jobs:
         docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG -t $ECR_REGISTRY/$ECR_REPOSITORY .
         docker push $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG
         docker push $ECR_REGISTRY/$ECR_REPOSITORY:latest
+
+    - name: Build, tag, and push image to Amazon ECR
+      if: ${{ github.ref == 'refs/heads/1.6-stable' }}
```
iamleeg, line 40: This either wants to be fuzzy matched too, or we need a variable somewhere that says what our current stable version is. Or we need a long-lived stable branch. Whichever way, we don't want to fix "Batch is on main, prod is on stable" by introducing "Batch is on old stable, prod is on new stable".

ingestion/functions/aws.py
```diff
@@ -11,7 +11,10 @@
 import common.common_lib as common_lib
 
 AWS_REGION = "us-east-1"
-AWS_IMAGE = "612888738066.dkr.ecr.us-east-1.amazonaws.com/gdh-ingestor:latest"
+AWS_IMAGE = {
+    'dev': "612888738066.dkr.ecr.us-east-1.amazonaws.com/gdh-ingestor:latest",
+    'prod': "612888738066.dkr.ecr.us-east-1.amazonaws.com/gdh-ingestor:stable"
+}
```
iamleeg, line 17: Does this change update the already-deployed batch jobs, or is there a manual intervention needed? If so, can we create a plan for doing that?

abhidg: Yes, there's a task in the issue to do that.

PR #1962, opened 2021-06-29
Upgrade docker publish action to v2

iamleeg:
Followed migration instructions at https://github.com/docker/build-push-action/blob/master/UPGRADE.md

I didn't touch the ingestion functions deployment as that's using a different workflow. Not sure how to test this without tagging or merging though.

### DIFF
--- a/.github/workflows/curator-service-package.yml
+++ b/.github/workflows/curator-service-package.yml
@@ -17,15 +17,34 @@ jobs:
 
     steps:
       - uses: actions/checkout@v2
-      - uses: docker/build-push-action@v1
+      -
+        name: Prepare
+        id: prep
+        run: |
+          DOCKER_IMAGE=${{github.repository}}/curatorservice
+          VERSION=edge
+          if [[ $GITHUB_REF == refs/tags/* ]]; then
+            VERSION=${GITHUB_REF#refs/tags/}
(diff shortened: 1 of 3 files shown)


.github/workflows/curator-service-package.yml
```diff
@@ -17,15 +17,34 @@ jobs:
...
+        name: Prepare
+        id: prep
+        run: |
+          DOCKER_IMAGE=${{github.repository}}/curatorservice
+          VERSION=edge
+          if [[ $GITHUB_REF == refs/tags/* ]]; then
+            VERSION=${GITHUB_REF#refs/tags/}
+          elif [[ $GITHUB_REF == refs/heads/* ]]; then
+            VERSION=$(echo ${GITHUB_REF#refs/heads/} | sed -r 's#/+#-#g')
+          elif [[ $GITHUB_REF == refs/pull/* ]]; then
```
abhidg, line 30: This won't be called as we are not building on PR, but okay to leave it I guess

abhidg (approved):
There should be a  github actions to test github actions :) Looks good

.github/workflows/curator-service-package.yml
iamleeg: Yup, I pasted that from the example docs and just deleted the outputs that I don't think we need, missed that bit.

Issue #1933, opened 2021-06-11
Swap out info@global.health for ghdsi.info@gmail.com across site / code base / social accounts


joe-brilliant:
Running list of instances:

- [x] Contact us footer
- [x] FAQ
- [x] TOU / Privacy Policy
- [x] Get Involved webform submission
- [x] User registration / 
- [x] Social media accounts
- [x] GitHub

rkassa:
@joe-brilliant 
Everything that's checked off on that list is done on dev.  The remaining items are ones I don't either have access to, or would concede to @maciekz1996 to implement (the user registration emails)

joe-brilliant:
@rkassa is this hooked up to Get Involved yet?

rkassa:
@joe-brilliant it is, and we've confirmed, right?

joe-brilliant:
I'll check these and all the dev ready issues, thanks @rkassa

joe-brilliant:
@jim-sheldon suggest we move to Proxy if this is pending filtered Data DL via email

jim-sheldon:
We do not use emails for filtered downloads anymore. I think we can close this.

PR #1974, opened 2021-07-01
Backport pendingRemoval and updated ingestion code to 1.6

abhidg:
- #1921 add flag to source documents indicating whether they have stable identifiers
- Add missing field for excluding source from line list to new source form
- Use material design for switch and add test for presence
- Add a switch for the stable identifiers setting.
- Add field to table for stable identifiers, and improve presentation of booleans in table.
- Only show sources with stable identifiers for backfill.
- Update email to info@global.health #1933
- Data service endpoints for pending removal and deletion
- parsing: Fix crash with missing country code in Argentina
- parsing_lib: Add safe_int()
- parsing: Skip negative or unconfirmed cases for Mexico
- Add curator service endpoints for pending case removals
- Remove unused import #1921
- add prop value to useMemo dependencies
- Add case revision metadata when marking and clearing pending deletion
- Create revision metadata when cases are deleted after pending
- Pass the curator email from curator service to data service
- Check in parser if the source has stable IDs and if not, delete the existing cases
- Move source information retrieval out of try: block
- patch up existing e2e tests
- Test unstable identifiers path
- parsing: Pass headers and cookies to fix 403
- curator/api: Add optional-require to fix test


iamleeg (approved):
I think all these changes were approved when they went into main, weren't they? No need to review again.

PR #1977, opened 2021-07-02
Enable tests for case deletion in non-UUID sources

abhidg:
Tests for non-UUID sources were not running because of this error:
 Returning a Promise from "describe" is not supported. Tests must be defined synchronously.
 Returning a value from "describe" will fail the test in a future version of Jest.


iamleeg (approved):
Oh whoops!

Issue #1976, opened 2021-07-02
dev-data does not load on Safari


abhidg:
dev-data does not load on Safari (macOS, iOS)

Problem appeared after a restart on 2021-07-01

iamleeg:
Found out the Safari doesn't support lookahead/lookbehind in regex: https://caniuse.com/js-regexp-lookbehind

PR #1980, opened 2021-07-05
Remove regex features that Safari does not support #1976

iamleeg:
Safari doesn't support lookahead/lookbehind in regex, so the syntax highlighting had broken Safari support. I've changed the regex so it doesn't use them, which obviously means it doesn't match the same things any more so I've basically broken syntax highlighting. @sergioloporto hoping your regex skills can help design a highlight implementation that will work, any ideas?

Here's the compatibility matrix: https://caniuse.com/js-regexp-lookbehind

### DIFF
--- a/verification/curator-service/ui/src/components/ViewCase.tsx
+++ b/verification/curator-service/ui/src/components/ViewCase.tsx
@@ -48,12 +48,6 @@ interface Props {
     onModalClose: () => void;
 }
 
-interface State {
-    case?: Case;
-    errorMessage?: string;
-    loading: boolean;
-}
-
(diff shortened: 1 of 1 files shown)


iamleeg:
Great, thanks @sergioloporto!

sergioloporto:
> Great, thanks @sergioloporto!

no worries, please also double check it on your side on Safari and Chrome... a double check is better than one

iamleeg:
Yes, works great, thank you! I'm not allowed to approve as it's "my" PR but thumbs up here :)

abhidg (approved):
Can't test on Safari, but looks good otherwise

Issue #1943, opened 2021-06-14
Address / resolve vulnerability / security / dependabot alerts 


joe-brilliant:
Placeholder issue so we can track / triage

https://github.com/globaldothealth/list/security/dependabot

PR #1982, opened 2021-07-05
Upgrade react-scripts #1943

iamleeg:
Also requires a lot of lint changes.

This is as much as I can do on the security report for the frontend, because the remaining issues are in dependencies on libraries that we're already up to date on. Tagging everyone so that you all see this change go through; updating `react-scripts` added a lot of new lint rules to the build so there is a massive change here!


abhidg:
Any way of telling the linter to not touch *.md?

iamleeg:
I abandoned this because the work involved in rebasing is great, and the way dependabot reports vulnerabilities appears to have become less noisy so this isn't generating alerts any more.

PR #1984, opened 2021-07-06
curator/api: Encode urls POSTed to data service

abhidg:


iamleeg (approved):
Nice, I would have wrapped `encodeURIComponent(user.email)` but this works equally well.

Issue #1991, opened 2021-07-08
Ingestion testing should use poetry


iamleeg:
Following on from #1918, the instructions to parser writers to test their parsers still tell them to use venv, even though the Docker infrastructure all uses poetry. This means potential for the two to get out of sync, and two sets of dependencies to monitor for security fixes. Let's go all-in on poetry.

iamleeg:
It does use poetry; it just wasn't documented.

PR #1992, opened 2021-07-08
1943 more security fixes

iamleeg:
This fixes the two vulnerabilities currently highlighted on the project home page.


abhidg:
@iamleeg requirements.txt is no longer used in the main branch after the move to poetry

iamleeg:
There's one place it is still used @abhidg: it's recommended for parser writers to write their unit tests. We would have to port all the test environments over to poetry (one for each parser), and the documentation, and help the parser authors learn the new tools, before we can remove it. I noted that as #1991 when I did this upgrade, because it wasn't obvious to me either that we still needed this file.

abhidg (approved):

PR #1998, opened 2021-07-12
download: Change from address to info@global.health #1933

abhidg:
This might need a corresponding change in AWS SES


jim-sheldon (approved):

jim-sheldon:
Confirmed `info@global.health` registered and verified in SES

PR #2001, opened 2021-07-14
Added localstack batch and locale2e

jim-sheldon:
- Create S3 buckets
- Create Batch Job Queue
- Add email address to SES
- Curator AWS clients use localstack when using locale2e docker-compose stack
- Curator async download uses localstack when using locale2e docker-compose stack

### DIFF
--- a/dev/run_full_stack.sh
+++ b/dev/run_full_stack.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+set -e
+# Store current directory.
+# We have to run docker compose from this directory for it to pick up the .env file.
+pushd `dirname "$0"`
+docker compose -f docker-compose.yml -f docker-compose.dev.full.yml up --build --force-recreate --renew-anon-volumes
+# Restore directory.
+popd
--- a/dev/setup_localstack.py
+++ b/dev/setup_localstack.py
@@ -1,0 +1,192 @@
+from os import environ
+from time import sleep
+
+import boto3
+import requests
+
+
+LOCALSTACK_URL = environ.get("AWS_ENDPOINT", "http://localstack:4566")
+BATCH_READY = '"batch": "running"'
+
+IAM_PROFILE_NAME = "ecsInstanceRole"
+IAM_PROFILE_ARN = "".join(["arn:aws:iam::000000000000:instance-profile/", "ecsInstanceRole"])
+BATCH_SERVICE_ROLE_NAME = "foo"
+ECS_INSTANCE_ROLE_NAME = "ecsInstanceRole"
+IAM_ROLE_POLICY_DOCUMENT = "file://./iam_role.json"
+EC2_SECURITY_GROUP_DESCRIPTION = "bar"
+EC2_SECURITY_GROUP_NAME = "sg-1234abcd"
+BATCH_COMPUTE_ENVIRONMENT_NAME = "M4Spot"
+BATCH_COMPUTE_ENVIRONMENT_ARN = "".join(["arn:aws:batch:us-east-1:000000000000:compute-environment/", BATCH_COMPUTE_ENVIRONMENT_NAME])
+BATCH_COMPUTE_ENVIRONMENT_TYPE = "MANAGED"
+
+BATCH_COMPUTE_RESOURCES = {
+    "type": "EC2",
+    "minvCpus": 1,
+    "maxvCpus": 10,
+    "instanceTypes": [
+        "m4.large",
+    ],
+    "imageId": "string",
+    "subnets": [],
+    "securityGroupIds": [],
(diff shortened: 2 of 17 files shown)


dev/run_full_stack.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+set -e
+# Store current directory.
+pushd `pwd`
+# We have to run docker compose from this directory for it to pick up the .env file.
+cd `dirname "$0"`
```
abhidg, line 6: can be pushd `dirname "$0"` as pushd also changes directory

dev/setup_localstack.py
```diff
@@ -0,0 +1,192 @@
...
+        try:
+            response = requests.get(healthcheck_url)
+            if BATCH_READY in response.text:
+                return
+        except requests.exceptions.ConnectionError:
+            pass
+        counter += 1
+        print("Waiting for localstack")
+        sleep(5)
+    raise Exception("Localstack not available")
```
iamleeg, line 172: When I try running `dev/run_full_stack.sh` locally I hit this exception, is there any setup I need to do? I've set `LOCALSTACK_API_KEY` in `dev/.env`.

dev/setup_localstack.py
```diff
@@ -0,0 +1,192 @@
...
+        self.ses_client.verify_email_identity(
+            EmailAddress=email_address
+        )
+        print("Verified email address")
+
+
+def wait_for_localstack():
+    healthcheck_url = "".join([LOCALSTACK_URL, "/health"])
+    counter = 0
+    while True and counter < 15:
```
abhidg, line 162: same as `while counter < 15`

abhidg (commented):
Couple of small changes, looks good otherwise!

dev/setup_localstack.py
jim-sheldon: whoops, forgot to remove that during cleanup. good catch.

dev/run_full_stack.sh
jim-sheldon: oh, did not know that. updated.

dev/setup_localstack.py
jim-sheldon: I suspect a timing issue. I increased the number of attempts to 20.

abhidg: Can't reproduce this, but I got this error message:

```Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /tmp/localstack```

Creating /tmp/localstack fixed this

Comment (iamleeg) [reply dev/setup_localstack.py:172]: > Creating /tmp/localstack fixed this

Should that ideally go in the host's `/tmp` or the docker image's `/tmp`? Either way, it needs to be scripted.

Comment (jim-sheldon) [reply dev/setup_localstack.py:172]: `/tmp` exists in the docker image, but I'm not using it for anything--the volume declared in the compose file was a default--so I removed it.

Comment (iamleeg) [reply dev/setup_localstack.py:172]: I still see that the timeout exception is raised before the localstack has launched:

```
setup-localstack_1  | Waiting for localstack
setup-localstack_1  | Traceback (most recent call last):
setup-localstack_1  |   File "./setup_localstack.py", line 177, in <module>
setup-localstack_1  |     wait_for_localstack()
setup-localstack_1  |   File "./setup_localstack.py", line 172, in wait_for_localstack
setup-localstack_1  |     raise Exception("Localstack not available")
setup-localstack_1  | Exception: Localstack not available
setup-localstack_1 exited with code 1
localstack_1        | Waiting for all LocalStack services to be ready
localstack_1        | 2021-07-16T08:25:19:INFO:botocore.credentials: Found credentials in environment variables.
localstack_1        | Waiting for all LocalStack services to be ready
localstack_1        | Ready.
localstack_1        | 2021-07-16 08:25:24,808:API: 127.0.0.1 - - [16/Jul/2021 08:25:24] "GET / HTTP/1.1" 200 -
localstack_1        | 2021-07-16T08:25:24:INFO:localstack.utils.analytics.profiler: Execution of "start_api_services" took 12558.80ms
```

Could the setup image use [`wait-for-it.sh`](https://github.com/vishnubob/wait-for-it/blob/master/wait-for-it.sh) and run when the localstack service is finally accepting connections?

abhidg (approved):

iamleeg:
I see the same test failure in FF on a different PR. I'll try to fix and get a commit for you to cherry-pick.

iamleeg:
I'm not proud but https://github.com/globaldothealth/list/pull/2003/commits/df56a02323a6d04372d94caddde09308b79ff3a9 works.

iamleeg (approved):

Issue #2002, opened 2021-07-15
Migrate to asynchronous deletion for non-UUID sources


abhidg:
- [x] Rename `pendingRemoval` to `list`
- [x] Modify export scripts to only export cases with list = true
- [x] Update get/list endpoints to only return cases with list = true
- [x] Update curator UI to insert new cases with list = true
- [x] Remove pendingRemoval code.
- [x] Write periodic script that sets everything other than the last successful upload to list = false (for non-UUID sources). For UUID sources, set everything to list = true. Delete everything with list = false. Run in batch or Gh Actions.
- [x] non-UUID sources were previously treated the same as UUID sources, i.e. one successful upload did not correspond to the entire dataset but a part. Set epoch date as an optional environment variable, after which successful uploads will be counted.
- [x] One-off after merge: Set `list` to true on dev and production database (in dev, delete GISAID data)
- [x] Add GitHub actions/Batch job to actually periodically prune cases

PR #2004, opened 2021-07-16
2002 async deletion

abhidg:
This doesn't have the workflow/Batch action to actually periodically run, will add it once this is merged and list is set to True on prod/dev.

- Remove markPendingRemoval endpoints and rename pendingRemoval->list
- Only export cases with list = True
- data: Update endpoints to only return cases with list=True #2002
- Add script to mark uploads periodically #2002


data-serving/data-service/test/controllers/case.test.ts
```diff
@@ -1759,114 +1759,3 @@ describe('DELETE', () => {
...
-
-        await request(app)
-            .post(`/api/cases/removePendingCases?sourceId=${sourceId1}`)
-            .expect(204);
-
-        const allCases = await Case.find({});
-        expect(allCases.length).toEqual(1);
-        expect(allCases[0].caseReference.sourceEntryId).toEqual(entryId2);
-    });
-});
```
iamleeg, line 1872: would be beneficial to prove that `list:false` cases are skipped, so that future devs understand that to be a requirement

data-serving/scripts/export-data/functions/02-export/app.py
```diff
@@ -47,8 +49,8 @@ def export_chunk(skip, limit, num_cases, num_chunk, num_chunks, field_names, exc
             f"--limit={limit}"
     ]
     if exclude_sources:
-        query = json.dumps({"caseReference.sourceId": {"$nin": exclude_sources}})
-        mongoexport.append(f"--query='{query}'")
+        query['caseReference.sourceId'] = {"$nin": exclude_sources}
```
iamleeg, line 52: Is `exclude_sources` here used for GISAID? If so this can now be removed.

data-serving/scripts/prune-uploads/requirements.txt
```diff
@@ -0,0 +1 @@
+pymongo
```
iamleeg, line 1: don't we want to use `poetry` everywhere now?

iamleeg: (besides which, it'd be good to pin the version)

iamleeg:
It doesn't look like there's any default value for `list` supplied anywhere in the API layer. Should there be? I'm guessing it should be `false` (or the type should be `bool?`). Then no cases are displayed until the periodic script has fixed up the values by connecting them to the correct uploads. And when we land this, some manual step will need to `list:true` all the existing cases.

Is that correct?

data-serving/scripts/prune-uploads/prune_uploads.py
```diff
@@ -0,0 +1,105 @@
+# prune_uploads.py -- Mark new uploads as ingested and delete failed uploads
+
+# This is a script that should be run periodically that sets every upload
+# older than the last successful upload to list = false (for non-UUID
+# sources). For UUID sources, set everything to list = true. After marking,
+# this deletes everything with list = false.
+
+import os
+import sys
+import pymongo
```
jim-sheldon, line 10: nitpick and not at all a blocker, but `import pymongo` should go in its own group below the other import statements, separated by a blank line https://www.python.org/dev/peps/pep-0008/#imports

data-serving/scripts/prune-uploads/prune_uploads.py
```diff
@@ -0,0 +1,105 @@
...
+    db.cases.update_many(
+        {"caseReference.sourceId": source_id}, {"$set": {"list": True}}
+    )
+
+
+if __name__ == "__main__":
+
+    # CONN is https://docs.mongodb.com/manual/reference/connection-string/
+    client = pymongo.MongoClient(os.environ.get("CONN"))
+    db = client.covid19
```
jim-sheldon, line 68: better to add some error checking around connecting to the database?

data-serving/scripts/prune-uploads/requirements.txt
abhidg: I prefer requirements.txt for simple one liner scripts with few dependencies, can change to poetry if we want to maintain consistency. Advantage of requirements.txt is the setup code is one-liner unlike poetry which uses a `curl | python` style of installation.

abhidg:
> It doesn't look like there's any default value for `list` supplied anywhere in the API layer. Should there be? I'm guessing it should be `false` (or the type should be `bool?`).

Currently there is no default value. Only `list=true` is considered true; this avoids changing a lot of code which has no knowledge of the `list` parameter; with the idea being that prune_uploads orchestrates the line list marking by itself. The curator UI has been changed to use `list=true` when adding or uploading cases. I couldn't see any optional type marker for the current fields, it seems that is set using `required: true` in the schema.

> Then no cases are displayed until the periodic script has fixed up the values by connecting them to the correct uploads. And when we land this, some manual step will need to `list:true` all the existing cases.

Yes, tracking this in #2002

data-serving/scripts/export-data/functions/02-export/app.py
abhidg: Good catch, removed

data-serving/data-service/test/controllers/case.test.ts
abhidg: Added a test case for this

jim-sheldon (approved):

PR #2014, opened 2021-07-22
2009 restrict notes

### LINKED ISSUE #2009: Hide notes from Brazil data sources
Oftentimes the case notes we import have surprisingly detailed, potentially even identifying, information about the affected cases. To avoid leaking those sensitive details, we want to hide the case notes until we've got better access control and can be confident of who is working with the notes.

The proposed solution is:

 - [x] add an optional field `restrictedNotes` to the case schema. This field won't get reflected in the curator portal or the data downloads.
 - [x] as a one-off action, move the `notes` from any Brazil case (national SRAG or state-level data) to `restrictedNotes`
 - [ ] update the Brazil parsers so all notes are stored in `restrictedNotes`
 - [x] write the one-off action to re-migrate the notes back from `restrictedNotes` to `notes`, for later use.

iamleeg:
Fixes #2009.


data-serving/scripts/convert-data/move_notes.py
```diff
@@ -0,0 +1,53 @@
...
+def database(connection_string, db_name):
+    """Connect to mongodb."""
+    client = pymongo.MongoClient(connection_string)
+    return client[db_name]
+
+
+def move_notes(db, collection, source_id, source, dest, dry_run):
+    """Move the values in the 'source' key to the 'dest' key for cases in the 'collection'
+    collection in the 'db' where their caseReference.source_id matches 'source_id'.
+    If 'remove_source' is set, then additionally remove values from the source key.
```
abhidg, line 16: remove_source isn't in params

abhidg (commented):
There's a missing parameter, looks good otherwise

jim-sheldon (approved):

data-serving/scripts/convert-data/move_notes.py
iamleeg: Good catch, it was originally but I realised it isn't needed, we never want to copy the notes into both fields. I'll remove it from the docstring!

PR #2015, opened 2021-07-22
prune-uploads: Use epoch and remove ERROR uploads

abhidg:
Epoch date specified using PRUNE_EPOCH is the date after which
successful uploads are considered for non-UUID sources. This is
to facilitate transition from the previous approach where non-UUID
sources were ingested over multiple, rather than single, uploads.


iamleeg (approved):

jim-sheldon (approved):
I might use the term `date` instead of `epoch` but LGTM

Issue #2020, opened 2021-07-26
Fix randomly failing Cypress tests


maciekz1996:
Some of the Cypress tests fail during CI on Github, most of the time there are timeout errors. The behaviour seems random as different tests fail each time.

abhidg:
Fixed in a8478f8

PR #2032, opened 2021-07-28
Update cypress version #2020

iamleeg:


iamleeg:
I'm not getting the chance to make good progress on this, I'll abandon it so others don't think it's already in progress. Please do take up the mantle of upgrading cypress!

PR #2038, opened 2021-07-28
prune-uploads: Add CLI arguments and mark accepted uploads

abhidg:
- prune-uploads: Add command line arguments
- curator/api: Remove pendingRemoval endpoints
- curator: Add accepted field to uploads
- prune-uploads: Mark accepted uploads in sources collection


data-serving/scripts/prune-uploads/prune_uploads.py
```diff
@@ -6,19 +6,44 @@
...
 
 
 def _ids(xs):
     return [str(x["_id"]) for x in xs]
 
 
-def accept_upload(upload: Dict[str, Any], threshold: float) -> bool:
-    """Whether to accept upload
+def notify(string, webhook_url):
+    import requests
```
jim-sheldon, line 30: Move this to the import block?

Issue #2042, opened 2021-07-29
Track data downloads


jim-sheldon:
We should keep a count of filtered downloads, and track them by user.
This will help us with analytics and setting up a cache.

joe-brilliant:
Thanks for getting this started @jim-sheldon ! cc @iamleeg @abhidg @Mougk

joe-brilliant:
@jim-sheldon @abhidg @iamleeg @Mougk I moved this into Strongcat for now just to ensure we have this tracking up when we launch and promote the feature. Happy to discuss further

jim-sheldon:
More precisely, we should add tracking of queries used in filtered downloads, e.g. `{country:France}`.
We already keep count of downloads by each user.

jim-sheldon:
Filtered data downloads now stored in the database

PR #2082, opened 2021-08-09
1949 streaming filtered download

jim-sheldon:
Data service streams filtered download instead of loading into memory
Curator service streams response from data service to user
Added skeleton for preliminary e2e test cases -- signup, login, download cases
Stubbed out SendWelcomeEmail lambda invocation
Added logging, including event to monitor for tracking filtered download requests (#2042)

Going to clean up logging, add a call to `setup_db.sh` in `run_full_stack.sh`, and add more cases, assertions and file cleanup in `run_session.py` before merging


data-serving/data-service/src/controllers/case.ts
```diff
@@ -14,6 +14,9 @@ import { logger } from '../util/logger';
 import stringify from 'csv-stringify/lib/sync';
 import _ from 'lodash';
 
+// TODO: de-jank
+const CaseFieldTextFile = 'https://raw.githubusercontent.com/globaldothealth/list/main/data-serving/scripts/export-data/functions/01-split/fields.txt';
```
abhidg, line 18: This can be loaded from the filesystem, right? data service image would need this present

data-serving/data-service/src/controllers/case.ts
```diff
@@ -61,113 +64,160 @@ export class CasesController {
...
         }
 
+        logger.info(`Streaming case data in format ${req.body.format} matching query ${req.body.query} for correlation ID ${req.body.correlationId}`);
+
         const queryLimit = Number(req.body.limit);
 
+        logger.info(`Request body: ${req.body}`);
+
         // Goofy Mongoose types require this.
         // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
abhidg, line 74: can be removed

data-serving/data-service/src/controllers/case.ts
```diff
@@ -61,113 +64,160 @@ export class CasesController {
...
             } else {
-                casesQuery = Case.find({ list: true }).lean();
-            }
-
-            // Limit number of results if present
-            // eslint-disable-next-line
-            if (queryLimit) {
-                casesQuery = casesQuery.limit(queryLimit);
+                logger.info('Request body with no query, limit, or case IDs');
+                cursor = Case.find({ list: true })
```
abhidg, line 122: Fine with this in data service, but curator service shouldn't hit this and redirect to the full dataset download

data-serving/data-service/src/controllers/case.ts
```diff
@@ -61,113 +64,160 @@ export class CasesController {
...
-                        .then((txtRes) => {
-                            const columns = txtRes.data.split('\n');
-                            const parsedCases = _.map(
-                                matchingCases,
-                                parseDownloadedCase,
-                            );
-                            const stringifiedCases = stringify(parsedCases, {
-                                header: true,
-                                columns: columns,
+                    await axios.get<string>(CaseFieldTextFile)
```
abhidg, line 168: csv and tsv can be combined into one block

data-serving/data-service/src/controllers/case.ts
```diff
@@ -61,113 +64,160 @@ export class CasesController {
...
+                    } catch(err) {
+                        logger.error(`Error streaming case data: ${err}`);
+                        res.status(500).json('A server error occurred while streaming case data');
+                        return;
+                    }
+                    res.end();
+                    
                     break;
 
                 case 'json':
```
abhidg, line 199: Looks good, maybe we should think about writing JSONL, since that's more natural for lists of records https://jsonlines.org/. Again, without user surveys, not sure which would be more useful. JSONL does make command line parsing easier (every record on its own line). Disadvantages: less well-known than JSON.

verification/curator-service/api/src/controllers/cases.ts
```diff
@@ -39,8 +39,10 @@ export default class CasesController {
     };
 
     /** Download forwards the request to the data service and streams the
-     * streamed response as a csv attachment. */
+     * streamed response as an attachment. */
     download = async (req: Request, res: Response): Promise<void> => {
+        const correlationId = crypto.randomBytes(16).toString('hex');
+        req.body.correlationId = correlationId;
```
abhidg, line 45: Won't this override the correlationId set by downloadAsync? unless that's the intended behaviour

abhidg (commented):
Looks good! Few nits, nothing major. I'd copy the fields.txt into the data service in the Docker image

data-serving/data-service/src/controllers/case.ts
jim-sheldon: For requests for .csv format? Or including .tsv and .json?

verification/curator-service/api/src/controllers/cases.ts
jim-sheldon: I don't think so; it's a separate endpoint from `downloadAsync` and one does not redirect to the other

data-serving/data-service/src/controllers/case.ts
jim-sheldon: Oh yeah

data-serving/data-service/src/controllers/case.ts
jim-sheldon: Done

data-serving/data-service/src/controllers/case.ts
jim-sheldon: Good call. Done.

data-serving/data-service/src/controllers/case.ts
jim-sheldon: Maybe. Gonna hold off for now and hoping command line users know `jq`.

data-serving/data-service/src/controllers/case.ts
abhidg: Yes, for csv. I don't think we should allow full dataset downloads except the standard csv since it'll put unnecessary load on data service

data-serving/data-service/src/controllers/case.ts
```diff
@@ -129,108 +134,54 @@ export class CasesController {
...
-                    .then((txtRes) => {
-                        columnsList = txtRes.data.split('\n');
-                        const columnsString = columnsList.join('\t');
-                        res.write(columnsString);
-                        res.write('\n');
+            let doc: CaseDocument;
+
+            if (req.body.format == 'csv' || req.body.format == 'tsv') {
+                res.setHeader('Content-Type', `text/${req.body.format}`);
+                let delimiter: string = (req.body.format == 'tsv') '\t' : ',';
```
abhidg, line 141: ternary op missing ?

abhidg (commented):
looks good other than a typo syntax error

data-serving/data-service/src/controllers/case.ts
jim-sheldon: yep, not sure how that happened, but caught it during testing

abhidg (approved):

dev/run_session.py
```diff
@@ -1,65 +1,160 @@
...
+                f.write(chunk)
+
+
+def load_case_fields() -> list[str]:
+    with open(FIELDS_FILE) as f:
+        contents = f.read()
+        return contents.splitlines()
+
+
+def check_files(fields: list[str]) -> None:
```
abhidg, line 74: list[str] syntax was introduced in Python 3.9, for 3.8 and below, it's List[str] with a from typing import List

jim-sheldon: Yep, I run 3.9 locally and intend to put this and any other e2e scripts in a Docker container running 3.9.

jim-sheldon: Ugh. I forgot we use 3.8 everywhere. Is this a vestige of using AWS Lambda, since that service does not support a python runtime more current than 3.8?
3.10 will release in October, and it would be good to stay current.

abhidg: Yes, Lambda doesn't support 3.9 https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html

abhidg (approved):

abhidg (approved):

joe-brilliant:
Hi @abhidg @jim-sheldon awesome to see the progress here! Which issue(s) are we using to track filtered data DL / emails on Strongcat? Or should we just add this issue to Strongcat? Just want to make sure they are captured so we can review / update / prioritize next week. thanks for keeping things moving!

jim-sheldon:
Hey @joe-brilliant I just added this issue to Strongcat. Didn't realize it hadn't been tagged before.
For tracking filtered downloads, we got https://github.com/globaldothealth/list/issues/2042.

PR #2085, opened 2021-08-10
prune-uploads: Prune older accepted uploads

abhidg:
Previously older, accepted uploads were not pruned. This was
incorrect behaviour for non-UUID sources (UUID sources don't get
pruned) as only the latest successful upload should be present
in the database, with older accepted uploads being deleted as well.

### DIFF
--- a/data-serving/scripts/prune-uploads/prune_uploads.py
+++ b/data-serving/scripts/prune-uploads/prune_uploads.py
@@ -78,14 +78,15 @@ def find_acceptable_upload(
       (last acceptable upload, list of uploads to mark for deletion).
       If no acceptable upload is found, returns None
     """
-    uploads = source.get("uploads", [])
-    # skip accepted or rejected uploads
-    uploads = [u for u in uploads if "accepted" not in u]
-    if not uploads:
+    if not (uploads := source.get("uploads", [])):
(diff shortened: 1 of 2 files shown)


data-serving/scripts/prune-uploads/prune_uploads.py
```diff
@@ -78,14 +78,15 @@ def find_acceptable_upload(
...
-    uploads = [u for u in uploads if "accepted" not in u]
-    if not uploads:
+    if not (uploads := source.get("uploads", [])):
         return None
 
     if source.get("hasStableIdentifiers", False):
-        return _ids(u for u in uploads if u['status'] != "IN_PROGRESS"), []
+        return _ids(
+            u for u in uploads
+            if u['status'] != "IN_PROGRESS"
```
jim-sheldon, line 87: I find `dict.get('key', <default>)` better than `dict['key']`

jim-sheldon (approved):

data-serving/scripts/prune-uploads/prune_uploads.py
abhidg: Normally I'd use that, but status is a required field in the schema so always present

PR #2086, opened 2021-08-10
parsing(srag): Add missing entries to country ISO2 map

abhidg:
Some countries were es_ES version, use pt_BR instead


AnyaLindstromBattle (approved):
Nice one.

PR #2091, opened 2021-08-12
retrieval: Add ::daysbefore=N suffix support to source URLs

abhidg:
Using the ::daysbefore=N suffix will offset the current date
by N days in the past before substitution in the source URL. This
is useful for sources that publish regularly, but whose filename
is offset by a fixed number of days.

### DIFF
--- a/ingestion/functions/retrieval/retrieval_test.py
+++ b/ingestion/functions/retrieval/retrieval_test.py
@@ -114,6 +114,14 @@ def test_format_url(mock_today):
     assert retrieval.format_source_url(
         url) == "http://foo.bar/2020-06-08/6/8.json"
 
+@patch('retrieval.retrieval.get_today')
+def test_format_url_daysbefore(mock_today):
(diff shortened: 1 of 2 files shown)


ingestion/functions/retrieval/retrieval_test.py
```diff
@@ -114,6 +114,14 @@ def test_format_url(mock_today):
     assert retrieval.format_source_url(
         url) == "http://foo.bar/2020-06-08/6/8.json"
 
+@patch('retrieval.retrieval.get_today')
```
jim-sheldon, line 117: you can also use `freezegun` for this https://github.com/spulec/freezegun

jim-sheldon (approved):

ingestion/functions/retrieval/retrieval_test.py
abhidg: Nice! didn't know about it :)

PR #2108, opened 2021-08-17
Restoring the filtered download button

jim-sheldon:


data-serving/data-service/src/controllers/case.ts
```diff
@@ -139,10 +139,25 @@ export class CasesController {
...
+
+            //Get current date
+            const dateObj = new Date();
+
+            // adjust 0 before single digit date
+            const day = ('0' + dateObj.getDate()).slice(-2);
+            const month = ('0' + (dateObj.getMonth() + 1)).slice(-2);
+            const year = dateObj.getFullYear();
+
+            const filename = `gh_${year}-${month}-${day}`;
```
abhidg, line 151: `(new Date()).toISOString().slice(0, 10)` is a one liner

abhidg (approved):

iamleeg (approved):

data-serving/data-service/src/controllers/case.ts
jim-sheldon: Nice, using that instead.

PR #2109, opened 2021-08-17
parsing: Retry on connection error to data service

abhidg:
Data service is occasionally unreachable, retry on receiving
a ConnectionError. This is different from the exponential backoff
that handles the case of data service returning a 5xx error.


ingestion/functions/common/parsing_lib.py
```diff
@@ -155,12 +158,21 @@ def write_to_server(
...
+            total_wait <= (MAX_WAIT_TIME if env in ["dev", "prod"] else 0)
+            and connection_tries < MAX_CONN_TRIES
+        ):
+            try:
+                res = requests.post(put_api_url, json={"cases": batch},
+                                    headers=headers, cookies=cookies)
+            except requests.exceptions.ConnectionError:
+                print(f"Failed to connect to data service, waiting {CONN_WAIT}s")
+                connection_tries += 1
+                time.sleep(CONN_WAIT)
```
iamleeg, line 175: if this fails all MAX_CONN_TRIES times we probably want an alarm because the data service has disappeared.

ingestion/functions/common/parsing_lib.py
```diff
@@ -25,6 +25,9 @@
 AUTH_FIELD = "auth"
 
 MAX_WAIT_TIME = 600  # 10 minutes maximum wait for response
+MAX_CONN_TRIES = 3  # Number of connection attempts to data service
+CONN_WAIT = 600  # Wait time between data service connection attempts
```
iamleeg, line 29: is this a bit pessimistic? The situation is "the data service appears to sometimes not be responding so we'll try a few times", is ten minutes too long to wait? How about exponential backoff from a couple of seconds up to a couple of minutes? Or do we know it takes 10 minutes for the data service to restart itself/start accepting connections again?

abhidg: Added another exponential backoff counting up to 15 minutes

ingestion/functions/common/parsing_lib.py
```diff
@@ -154,13 +157,28 @@ def write_to_server(
...
+        while (
+            total_wait <= (MAX_WAIT_TIME if env in ["dev", "prod"] else 0)
+            and total_conn_wait < MAX_CONN_WAIT_TIME
+        ):
+            try:
+                res = requests.post(put_api_url, json={"cases": batch},
+                                    headers=headers, cookies=cookies)
+            except requests.exceptions.ConnectionError:
+                print(f"Failed to connect to data service, waiting {conn_wait}s")
+                time.sleep(conn_wait)
```
jim-sheldon, line 178: Could we use the `timeout` parameter of `.post()` rather than `time.sleep()`?

jim-sheldon:
Do we know what causes the data service to be unreachable?

jim-sheldon (approved):

ingestion/functions/common/parsing_lib.py
abhidg: Didn't know about that! With no timeout, requests will wait for the data service which I think is the correct approach when inserting cases as we don't know how long that'll take.

PR #2115, opened 2021-08-19
Added null option to dateRange on event parse

jim-sheldon:


jim-sheldon:
I also pinned the mongo version to what we use in prod

data-serving/data-service/src/util/case.ts
```diff
@@ -12,15 +12,16 @@ export const parseCaseEvents = (
...
 } =>
     events.reduce(
-        (agg, { name, value, dateRange }: EventDocument) => ({
+        (agg, { name, value, dateRange }: EventDocument) => (
+        {
             ...agg,
             [name]: {
                 value: value ?? '',
-                date: new Date(dateRange.start).toISOString().split('T')[0], // dateRange.start and dateRange.end have always the same values
+                date: (dateRange) ? new Date(dateRange.start).toISOString().split('T')[0] : null, // dateRange.start and dateRange.end have always the same values
```
jim-sheldon, line 24: Do dateRange.start and dateRange.end truly always have the same values?

abhidg (approved):

jim-sheldon:
Updated UI to give feedback on download in progress instead of silently firing request and leaving modal open.

abhidg (approved):

PR #2122, opened 2021-08-23
Use hardcoded locations for some parsers

abhidg:
We rarely have location data below admin2 level, and some parsers have a small set which we can manually geocode and save; this should speed up parsing as it won't need to refer to the location service.


ingestion/functions/common/parsing_lib.py
```diff
@@ -36,6 +36,10 @@
 # usage on the server-side and is known to cause OOMs so increase with caution.
 CASES_BATCH_SIZE = 250
 
+with (Path(__file__).parent / 'geocoding_countries.json').open() as g:
+    GEOCODING_COUNTRIES = json.load(g)
+    COUNTRY_ISO2 = sorted(GEOCODING_COUNTRIES.keys())
+
```
jim-sheldon, line 42: I think adding a `try/except` block here and logging any error would be good.

ingestion/functions/common/parsing_lib.py
```diff
@@ -54,6 +58,20 @@ def safe_int(x):
...
+    try:
+        lat, lon, country = GEOCODING_COUNTRIES[two_letter_iso_code]
+        return {
+            "country": country,
+            "geoResolution": "Country",
+            "name": country,
+            "geometry": {"latitude": lat, "longitude": lon}
+        }
+    except KeyError:
+        return None
```
jim-sheldon, line 72: I'd log the exception here too. Are there cases in which we expect the `two_letter_iso_code` to not be in `GEOCODING_COUNTRIES`?

ingestion/functions/parsing/cuba/cuba.py
```diff
@@ -16,6 +20,59 @@
...
+    newloc['geoResolution'] = "Admin1"
+    return newloc
+
+# Geocoding data is from Wikipedia:
+# Municipalities of Cuba,
+# https://en.wikipedia.org/w/index.php?title=Municipalities_of_Cuba&oldid=1010874027
+# (version as of 21 August 2021)
+
+with (Path(__file__).parent / 'geocodes.csv').open() as geof:
+    reader = csv.DictReader(geof)
```
jim-sheldon, line 42: I'd also wrap this in a `try/except` and log any errors

jim-sheldon (commented):
I'd suggest adding error handling and logging anywhere there exists code like:
`with (Path(__file__).parent / 'geocodes.json').open() as geof:
    _GEOCODES = json.load(geof)`
otherwise LGTM

ingestion/functions/common/parsing_lib.py
iamleeg: Could happen, we've seen for example that map box and google disagree on whether to only use two-letter country codes or also two-letter region codes for autonomous regions (e.g. `gi` for Gibraltar).

ingestion/functions/parsing/cuba/cuba.py
abhidg: Geocoding can't fail without the rest of it working, so I'd rather have it crash immediately. AWS Cloudwatch captures exception tracebacks.

jim-sheldon (approved):

ingestion/functions/parsing/new_zealand/sample_data.csv
```diff
@@ -1,11 +1,11 @@
...
+2021-01-11,Confirmed,Male,40 to 49,Managed Isolation & Quarantine,Yes,
+2021-01-11,Confirmed,Female,40 to 49,Managed Isolation & Quarantine,Yes,
+2021-01-11,Confirmed,Male,20 to 29,Managed Isolation & Quarantine,Yes,
+2021-01-11,Confirmed,Female,10 to 19,Managed Isolation & Quarantine,Yes,
+2021-01-10,Confirmed,Male,30 to 39,Managed Isolation & Quarantine,Yes,
+2021-01-10,Confirmed,Male,0 to 9,Managed Isolation & Quarantine,Yes,
+2021-01-10,Confirmed,Male,30 to 39,Managed Isolation & Quarantine,Yes,
+2021-01-10,Confirmed,Male,20 to 29,Managed Isolation & Quarantine,Yes,
+2021-05-05,Confirmed,Female,90+,Bay of Plenty,Yes,
+2021-05-06,Confirmed,Female,20 to 29,West Coast,No,
```
iamleeg, line 11: Are these date format changes reflected in the upstream data?

abhidg: Yes, upstream changed date format.

PR #2133, opened 2021-08-31
parsing(new_zealand): Fix error when age = NA

abhidg:
Don't set demographics when age = NA and gender = Unknown

### DIFF
--- a/ingestion/functions/parsing/new_zealand/new_zealand.py
+++ b/ingestion/functions/parsing/new_zealand/new_zealand.py
@@ -62,13 +62,13 @@ def convert_demographics(entry):
     If age is listed as 90+, setting age range as between 90 and 120.
     '''
     demo = {}
-    if age := entry[_AGE]:
+    if (age := entry[_AGE]) != "NA":
         if '90+' in age:
             demo["ageRange"] = {"start": 90, "end": 120}
         else:
             start, end = list(map(int, age.split(' to ')))
             demo["ageRange"] = {"start": start, "end": end}
-    if gender := entry[_GENDER]:
+    if (gender := entry[_GENDER]) != "Unknown":
         demo["gender"] = gender if gender in ['Male', 'Female'] else None
 
     return demo or None
@@ -117,6 +117,8 @@ def parse_cases(raw_data_file, source_id, source_url):
                         },
                     ]
                 }
+                if case["demographics"] is None:
+                    del case["demographics"]
                 if entry[_TRAVEL] == 'Yes':
                     case["travelHistory"] = {
                         "traveledPrior30Days": True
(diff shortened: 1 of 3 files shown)


jim-sheldon (approved):

ingestion/functions/parsing/new_zealand/new_zealand_test.py
```diff
@@ -47,20 +47,23 @@
 
 
 def _c(location, gender, age_start, age_end, confirmed_date, travel=False, notes=None):
+    demographics = {}
+    if age_start is not None and age_end is not None:
+        demographics["ageRange"] = {"start": age_start, "end": age_end}
```
iamleeg, line 52: is there value in keeping partial information? Can we create a range if we only have the start age, for example? (Also: does that ever happen?)

ingestion/functions/parsing/new_zealand/new_zealand_test.py
```diff
@@ -47,20 +47,23 @@
...
-        },
         "events": [
             {
                 "name": "confirmed",
                 "dateRange": {"start": confirmed_date, "end": confirmed_date},
             }
-        ]
+        ],
     }
+    if demographics:
```
iamleeg, line 65: this test is redundant, as demographics is definitely not None (it might be empty though)

abhidg: that checks for demographics != {}, as we don't want to add an empty demographics

ingestion/functions/parsing/new_zealand/new_zealand.py
```diff
@@ -62,13 +62,13 @@ def convert_demographics(entry):
     If age is listed as 90+, setting age range as between 90 and 120.
     '''
     demo = {}
-    if age := entry[_AGE]:
+    if (age := entry[_AGE]) != "NA":
```
iamleeg, line 65: we now enter the true branch if age is None, and (still) raise `KeyError` if `_AGE not in entry`, are these both intended?

ingestion/functions/parsing/new_zealand/new_zealand_test.py
abhidg: it's always 90+ or X to Y, also this is for the test runner which should always have both limits specified

ingestion/functions/parsing/new_zealand/new_zealand.py
abhidg: _AGE should always be in entry, it can be NA or a valid value

ingestion/functions/parsing/new_zealand/new_zealand_test.py
iamleeg: ah you're right, it does! I thought empty but existing collection would be truthy.

iamleeg (approved):

Issue #2120, opened 2021-08-23
Add Vaccination Status to Schema


Mougk:
As we enter a new phase of the pandemic in spring 2021; namely that people have been vaccinated, we would like to add vaccination status to the schema.

**Proposal**: 

Add a new section to the schema called: "Vaccination"

1. Vaccination dose 1
Answers: yes/no

2. Vaccine Name dose 1 
Answer: list of WHO licensed vaccines (https://www.who.int/news-room/q-a-detail/coronavirus-disease-(covid-19)-vaccines?topicsurvey=v8kj13)&gclid=Cj0KCQjwjo2JBhCRARIsAFG667VxncQUvTueaWqoZkFz6KAyRL9juTKY4H83r61p8x6ygq-gYt1k_WEaAk99EALw_wcB) or other (free text)

3. Vaccination dose 1 date
Answer: 'date'

4. Side effects dose 1
Answer: Symptom ontology

5. Vaccination dose 2
Answers: yes/no

6. Vaccine Name dose 2
Answer: list of WHO licensed vaccines (https://www.who.int/news-room/q-a-detail/coronavirus-disease-(covid-19)-vaccines?topicsurvey=v8kj13)&gclid=Cj0KCQjwjo2JBhCRARIsAFG667VxncQUvTueaWqoZkFz6KAyRL9juTKY4H83r61p8x6ygq-gYt1k_WEaAk99EALw_wcB) or other (free text)

7. Vaccination dose 2 date
Answer: 'date'

8. Side effects dose 2
Answer: Symptom ontology

9. Vaccination dose booster
Answers: yes/no

10. Vaccine Name dose 3
Answer: list of WHO licensed vaccines (https://www.who.int/news-room/q-a-detail/coronavirus-disease-(covid-19)-vaccines?topicsurvey=v8kj13)&gclid=Cj0KCQjwjo2JBhCRARIsAFG667VxncQUvTueaWqoZkFz6KAyRL9juTKY4H83r61p8x6ygq-gYt1k_WEaAk99EALw_wcB) or other (free text)

11. Vaccination dose 3 date
Answer: 'date'

12. Side effects dose 3
Answer: Symptom ontology

13. Previous COVID-19 infection
Answer:  yes, no, NA

14.  Previous COVID-19 infection detection method
Answer: Clinical symptoms, antibody test, PCR test, Antigen test

abhidg:
[...]

Issue #2103, opened 2021-08-17
Fields.txt used in three places


jim-sheldon:
The `fields.txt` file gets consumed by the data service and two helper scripts. All of these have access to the database.
We should add the contents of the file to the database and use that as the single, shared source of truth.

PR #2134, opened 2021-08-31
#2120 vaccine status

iamleeg:
I've made the schema changes and added the fields to the data downloads, but not made any changes to the line list or curator form. This at least allows us to quickly get the data in where ADI sources contain it, and we can decide later how the UI needs to support the fields.


data-serving/data-service/src/model/vaccine.ts
```diff
@@ -0,0 +1,23 @@
...
+    },
+    { _id: false },
+);
+
+export type VaccineDocument = mongoose.Document & {
+    name: string;
+    batch: string;
+    date: Date;
+    sideEffects: [SymptomsDocument];
+    previousInfection: 'yes' | 'no' | 'NA';
```
abhidg, line 21: Could this be boolean? That way presence of previousInfection would have true / false and absence indicates no previous infection or data not known.

abhidg (commented):
Looks good - one field could be boolean. We should come up with a way to address #2103, but that can be done later.

jim-sheldon:
IMHO the case fields workflow could get consolidated/simplified--
We have the`export` script which pulls field names from the database into a text file, now a `prepare` script to turn the text file into a JSON file, which then gets consumed by `controllers/case`.
Could we simply pull the fields from the database, then feed them into the case controller constructor?

data-serving/data-service/src/model/vaccine.ts
iamleeg: The request at #2120 was for yes/no/NA. I guess we have to ask @mougk and other experts what the expectation is.

jim-sheldon (approved):

iamleeg:
@jim-sheldon Oh is the export script the canonical source of fields.txt (from the database)? I didn't know that! I don't know how that will work with this change: there's a list of vaccine documents for each case document, but we specifically want the CSV export to have space for exactly three vaccines per case.

iamleeg:
The database shouldn't ever be the canonical source of what fields get included in the export, because we have fields that are present like `restrictedNotes` that should not be exported.

jim-sheldon:
It looks like it will work just fine once the schema is applied.
Agreed the database shouldn't be the unqualified source...I'd like there to exist a `hidden` or `private` field in the [jsonSchema spec](https://docs.mongodb.com/v4.2/reference/operator/query/jsonSchema/#json-schema) but no dice...in the absence of that, we do `delete doc.restrictedNotes;` and supporting tests for that requirement exist.

abhidg (approved):

Issue #2098, opened 2021-08-16
Add ability to sort by date added/confirmed to Data application


joe-brilliant:
(Placeholder issue)

abhidg:
@joe-brilliant Suggest moving this and #1875 to Lewis. The work is done, but a migration is running to denormalise confirmed date which isn't going to be completed by 30 Nov.

PR #2139, opened 2021-09-02
1880 mongo migrations

iamleeg:
I'm quite worried about this PR because it has a very deep effect on the infra, but I've tested all the local dev use cases and don't appear to have broken anything! I haven't tested the changes to the GitHub workflows, of course.

I've updated all the documentation, so rather than try to explain how things work here I'll let that stand for itself: if it doesn't, please amend/request clarification and we'll get to the point where the docs carry all the necessary information.

One remaining point that doesn't fit in the docs: I removed the bit where the setup script would automatically drop all cases that were imported from the spreadsheet. I don't see the point of that, and the migrations can be designed so that those cases are valid both before and after the migration. Also I know it's caused problems before where those cases have gone missing and we didn't understand why. Let me know what you think: my preference is to leave that behaviour out, and if we need to drop those cases on a future migration then the migration itself can do that.


jim-sheldon (approved):
Awesome

dev/setup_db.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-DELETE_ALL_DOCUMENTS=true \
-npm --prefix=`dirname "$0"`/../data-serving/scripts/setup-db/ run-script setup-cases
-npm --prefix=`dirname "$0"`/../data-serving/scripts/setup-db/ run-script setup-restricted-cases
-npm --prefix=`dirname "$0"`/../data-serving/data-service/ run-script import-data
\ No newline at end of file
+CONN="${CONN:-mongodb://localhost/covid19}"
```
abhidg, line 3: Is this only for local testing? Then I would set CONN directly to prevent errors if CONN is somehow set manually and dev/setup_db.sh is run by mistake. Alternatively disallowing mongodb+srv in CONN could help prevent this as well.

data-serving/scripts/setup-db/schemas/sources.schema.json
```diff
@@ -28,10 +28,10 @@
                 "bsonType": "string"
             },
             "excludeFromLineList": {
-                "bsonType": "boolean"
+                "bsonType": "bool"
```
abhidg, line 31: Is this due to migrate-mongo?

data-serving/scripts/setup-db/package.json
```diff
@@ -2,14 +2,12 @@
...
-    "setup-cases": "npm install && tsc && mongo --eval \"let args={databaseName: 'covid19', collectionName: 'cases', schemaPath: './../../data-service/schemas/cases.schema.json', indexesPath: './../../data-service/schemas/cases.indexes.json', deleteAllDocuments: '$DELETE_ALL_DOCUMENTS'}\" dist/index.js",
-    "setup-restricted-cases": "npm install && tsc && mongo --eval \"let args={databaseName: 'covid19', collectionName: 'restrictedcases', schemaPath: './../../data-service/schemas/cases.schema.json', indexesPath: './../../data-service/schemas/cases.indexes.json', deleteAllDocuments: '$DELETE_ALL_DOCUMENTS'}\" dist/index.js",
-    "setup-sources": "npm install && tsc && mongo --eval \"let args={databaseName: 'covid19', collectionName: 'sources', schemaPath: './../../../verification/curator-service/api/src/schemas/sources.schema.json', indexesPath: './../../../verification/curator-service/api/src/schemas/sources.schema.json', deleteAllDocuments: '$DELETE_ALL_DOCUMENTS'}\" dist/index.js"
+    "drop-cases-and-import-sample-data": "mongoimport --drop --uri=$CONN --db covid19 --collection cases --jsonArray --file ../../samples/cases.json",
```
abhidg, line 9: Can use $DB variable

abhidg (commented):
Looks good! A few minor comments.

Can migrations also add/modify data? I'm thinking of #2098 which would require a new "confirmedDate" field, copying over existing information from events[0].

data-serving/scripts/setup-db/schemas/sources.schema.json
iamleeg: I'm not sure why this worked before if it did work before: boolean isn't a valid bson type alias.

dev/setup_db.sh
iamleeg: OK, that's a good point.

iamleeg:
> Can migrations also add/modify data?

Yes! You can run anything you like in the `up()` function, so you can modify data if needed.

abhidg (approved):

PR #2148, opened 2021-09-13
2083 country drop down

sergioloporto:


verification/curator-service/ui/src/redux/store.ts
```diff
@@ -2,17 +2,20 @@ import { configureStore, combineReducers } from '@reduxjs/toolkit';
 
 import appReducer from './app/slice';
 import authReducer from './auth/slice';
+import countryListReducer from './filters/slice';
```
maciekz1996, line 5: To make things consistent you can change the name here from `countryListReducer` to `filtersReducer`

verification/curator-service/ui/src/redux/filters/thunk.ts
```diff
@@ -0,0 +1,30 @@
...
+
+const countryListJsonFile =
+    'https://covid-19-aggregates.s3.amazonaws.com/country/latest.json';
+
+type countryObject = {
+    _id: string
+}
+
+export const fetchCountries = createAsyncThunk(
+    'countryList/fetchCountries',
```
maciekz1996, line 12: Here you can change `countryList/fetchCountries` to `filters/fetchCountries` to make things consistent

verification/curator-service/ui/src/redux/filters/slice.ts
```diff
@@ -0,0 +1,29 @@
...
+    isLoading: false
+};
+
+interface initialStateTypes {
+    countryList:string[],
+    isLoading: boolean
+  }
+
+const slice = createSlice({
+    name: 'countryList',
```
maciekz1996, line 15: You can change the name to `filters`

verification/curator-service/ui/src/redux/@types/react-redux.d.ts
```diff
@@ -0,0 +1,7 @@
+import 'react-redux';
```
maciekz1996, line 1: What was the purpose of adding those custom types?

maciekz1996 (changes requested):

verification/curator-service/ui/src/redux/@types/react-redux.d.ts
sergioloporto: I was getting ts errors, and I found that it is a good practice to add that custom type

maciekz1996 (changes requested):

verification/curator-service/ui/cypress/integration/components/LinelistTableTest.spec.ts
```diff
@@ -454,7 +468,7 @@ describe('Linelist table', function () {
         cy.contains('7 rows selected').should('not.exist');
     });
 
-    it('Can delete all cases across rows for a search result', function () {
+    it.only('Can delete all cases across rows for a search result', function () {
```
maciekz1996, line 471: This `only` is preventing other tests from running in this test file

maciekz1996 (changes requested):

maciekz1996 (approved):

iamleeg (approved):
Looks great, nice one. Some slight data concerns (Gibraltar is listed as a country, officially it isn't, for example; Haiti and Honduras are missing) but we already have a couple of different places where Mapbox and other sources don't agree on this. I think there's no value and a lot of time spent in trying to build a "canonical" list, so let's leave it as it is and if it turns out that someone can't do something they want we'll address it then.

sergioloporto:
@iamleeg agreed, thank you

Issue #1875, opened 2021-05-24
Large scale date sorting / filtering


maciekz1996:
Try to improve performance of sorting filtering big data queries

abhidg:
Likely an indexing issue since at the moment confirmed date is located within the events array. Should be fixed by adding a `confirmedDate` field and index with just the date

iamleeg:
I don't know whether there are specific expectations of what to speed up here, but I see a small number of queries in prod that are still slow:

 - a couple times someone searched for `verification status is EXCLUDED and sourceId is {{Taiwan source}} and confirmation date between 1 dec 2019 and now`, seems a bit weirdly specific but may be an important case
 - someone was searching for cases by one curator (maybe themselves?) and their query came out as a text search for `curatoremail:{{the email address}}`, so this field wasn't parsed by the back end. Maybe it should be, and should be indexed? They were also searching for `country:Deutschland` which is not going to match but will be fixed by having the country drop-down in the search form.
 - the aggregations for generating the map data do a whole collection scan. They search for `list:true`, then e.g. one of them projects country fields, then buckets by country. Maybe the bucketing could be done before the projection to speed it up.
 - all of the mongoexport commands that rely on `list:true` are all slow. Probably we need to look for another way to deal with that parser limitation.

iamleeg:
> a couple times someone searched for verification status is EXCLUDED and sourceId is {{Taiwan source}} and confirmation date between 1 dec 2019 and now, seems a bit weirdly specific but may be an important case

This just came up again as the slowest query on the cluster in the graph window, so someone is doing it reasonably frequently.

iamleeg:
A couple of indexes recommended by the Atlas automatic tool:

caseReference.sourceId: 1
caseReference.verificationStatus: 1

and

list: 1
revisionMetadata.creationMetadata.date: -1

PR #2157, opened 2021-09-22
New indexes suggested by the Atlas performance review #1875

iamleeg:
The dashboard suggested these as high-impact indexes to speed up some common queries.


iamleeg:
There were two indexes on the recommendation yesterday, and there are four today! Between them, these should speed up the very slow "Taiwan and excluded" query that runs every few hours, and the batch tasks @abhidg wrote to fake atomic updating of cases without stable identifiers.

The remaining very-slow activity is preparing the daily download, because mongoexport pages through every case in the collection. There's nothing we can do about that: it needs to page through every case in the collection.

jim-sheldon (approved):

PR #2158, opened 2021-09-23
2068 data deep dive sort by option not necessary filter country

sergioloporto:


verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -441,11 +442,14 @@ export function SortSelect({
...
 
+    const filterBreadCrumbs = useSelector(selectFilterBreadcrumbs);
+    let filteredKeys = filterBreadCrumbs.map(({ key }) => key);  
+
     const sortKeywords = [
         { name: 'None', value: SortBy.Default },
         // Commenting out until fix is found for big queries sorting in mongodb
         // { name: 'Confirmed date', value: SortBy.ConfirmedDate },
-        { name: 'Country', value: SortBy.Country },
+        {...filteredKeys[0] !== 'country' && { name: 'Country', value: SortBy.Country }},
```
maciekz1996, line 452: You should use `filteredKeys.includes('country')` here, because country filter won't always be first

maciekz1996 (changes requested):
Apart from the comments that I left above this feature is not working correctly. When `country` filter is applied then in the `Sort By` dropdown there is a blank item, after selecting this blank item an error appears. It should be hidden completely if the `country` filter is applied. Also when country filter is not applied as the first filter then sorting by country is still allowed. 

Screen Shot 2021-09-23 at 15 46 33
Screen Shot 2021-09-23 at 15 46 39

verification/curator-service/ui/src/components/LinelistTable.tsx
```diff
@@ -441,17 +443,22 @@ export function SortSelect({
...
         // Commenting out until fix is found for big queries sorting in mongodb
         // { name: 'Confirmed date', value: SortBy.ConfirmedDate },
-        { name: 'Country', value: SortBy.Country },
         { name: 'Location admin 1', value: SortBy.Admin1 },
         { name: 'Location admin 2', value: SortBy.Admin2 },
         { name: 'Location admin 3', value: SortBy.Admin3 },
         { name: 'Age', value: SortBy.Age },
     ];
 
+    !filteredKeys.includes('country') && sortKeywords.push({ name: 'Country', value: SortBy.Country });
```
maciekz1996, line 459: You can use `splice` method here to still keep `country` as the first filter in the dropdown. Like for example: 
`sortKeywords.splice(1, 0, { name: 'Country', value: SortBy.Country })`

maciekz1996 (approved):
Everything looks fine, I added one comment with a suggestion

iamleeg (approved):
Looks good, thank you

PR #2163, opened 2021-09-24
Switching findOneAndUpdate with findByIdAndUpdate

jim-sheldon:
And adding logging on errors, and database entries for download formats and queries.

### DIFF
--- a/verification/curator-service/api/src/controllers/cases.ts
+++ b/verification/curator-service/api/src/controllers/cases.ts
@@ -1,5 +1,4 @@
 import { Request, Response } from 'express';
-import { Worker } from 'worker_threads';
 import { User, UserDocument } from '../model/user';
 import axios from 'axios';
 import { logger } from '../util/logger';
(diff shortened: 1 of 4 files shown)


verification/curator-service/api/src/controllers/cases.ts
```diff
@@ -136,11 +167,20 @@ export default class CasesController {
...
-                {
-                    _id: user._id,
-                },
-                { $push: { downloads: { timestamp: new Date() } } },
+            await User.findByIdAndUpdate(
+                user._id,
+                { $push: { downloads: {
+                    timestamp: new Date(),
+                    format: req.body.format,
+                    query: req.body.query,
```
jim-sheldon, line 175: Should I leave `format` and `query` out here? Or make them `csv` and `{}`?

iamleeg: I would say pass on what's in the req, but check in swagger if they're required here and supply defaults if not.

iamleeg (approved):
assuming it passes tests now, let's get it out there and try it

verification/curator-service/api/src/controllers/cases.ts
jim-sheldon: They're optional fields for backwards compatibility. Omitting them here.

PR #2171, opened 2021-09-28
1939 adding parsing to local e2e

jim-sheldon:


jim-sheldon:
TODO:
- figure out why the USA parser tries to parse an empty file
- cleanup `print` statements
- condense environment variables
- try fixes to localstack
- add function to reupload `sample_data` files to S3
- add function to rebuild a parser image
- figure out why batchUpserted cases don't appear in the curator UI
- rebase and squash commits
- update README

dev/Dockerfile-ingestion
```diff
@@ -0,0 +1,7 @@
+FROM python:3.8
+
+RUN pip install boto3==1.18.38 requests==2.26.0 docker==5.0.2
```
iamleeg, line 3: is there value to using a separately-defined environment (e.g. a poetry specification) here? Might I want to develop/test these scripts outside the docker environment, or is that too much hassle to support?

dev/docker-compose.dev.full.yml
```diff
@@ -41,17 +41,18 @@ services:
     ports:
       - "53:53"
       - "443:443"
+      - "4510:4510"
```
abhidg, line 44: Would be good to have docs on which port is for what and linking to localstack docs

dev/mock_source_data.py
```diff
@@ -0,0 +1,165 @@
...
+	"automation": {
+		"parser": {
+			"awsLambdaArn": "",
+		},
+	},
+	"dateFilter": "",
+	"hasStableIdentifiers": False
+}
+
+SOURCE_NAMES_TO_PARSERS = {
```
abhidg, line 29: OK for now, but ideally this should be taken automatically from the parsing/ directory.

dev/mock_source_data.py
```diff
@@ -0,0 +1,165 @@
...
+	"taiwan": "taiwan-taiwan-ingestor-locale2e",
+	"thailand": "thailand-thailand-ingestor-locale2e",
+	"usa": "USA-USA-ingestor-locale2e",
+	"variants": "variants-variants-ingestor-locale2e",
+}
+
+SOURCE_IDS_TO_NAMES = {}
+
+UPLOAD_RESPONSE = {
+	"_id": "".join("0" for i in range(24))
```
abhidg, line 72: “0” * 24 works as well :)

dev/mock_source_data.py
```diff
@@ -0,0 +1,165 @@
...
+
+		response = self.s3_client.list_objects(
+			Bucket=BUCKET_NAME,
+			Prefix=source_name
+		)
+		objects = response.get("Contents", [])
+		if not objects:
+			raise Exception(f"No sample data files present for {source_name}")
+		obj = objects[0]
+		obj_key = obj.get("Key")
```
abhidg, line 97: Just a nit - you can use walrus := (assignment) operator to collapse assignment and value checking https://docs.python.org/3/reference/expressions.html#assignment-expressions

ingestion/functions/common/parsing_lib.py
```diff
@@ -427,14 +436,18 @@ def run(
...
 
         case_data = parsing_function(
             local_data_file_name, source_id,
             source_url)
+        print(f"Got case data: {case_data}")
         excluded_case_ids = retrieve_excluded_case_ids(source_id, date_filter, date_range, env,
                                                        headers=api_creds, cookies=cookies)
+        print(f"Excluded case ids: {excluded_case_ids}")
         final_cases = prepare_cases(case_data, upload_id, excluded_case_ids)
+        print(f"Final cases: {final_cases}")
```
abhidg, line 450: This is a generator, so won’t be much use in printing

dev/Dockerfile-ingestion
abhidg: Do we plan to support other languages than Python? Otherwise we could just use a venv/poetry here.

abhidg (commented):
Looks good, other than a few comments!

dev/mock_source_data.py
jim-sheldon: IMHO we should refactor some of our Python parser/ingestor code and make it into a library, then we can simply import that library and call a function to give us that.
Relying on relative pathing and breaking dependency inversion is bad, and so is hardcoding this dict.
In the setup_ingestion service I mounted "../ingestion/functions:/ingestion/functions" as a volume...so importing a function from a library and calling it will shrink LOC and our Docker images, and make them less brittle (and probably yield other benefits).

dev/mock_source_data.py
jim-sheldon: It does seem a bit suspect in a dynamically-typed language. The [Zen of Python](https://www.python.org/dev/peps/pep-0020/#id2) says to favor explicit over implicit, so I think I slightly prefer `range`.

dev/mock_source_data.py
jim-sheldon: I need to use the walrus more. In this block of code, I think it's more readable without them though:
```
		if objects := response.get("Contents", []):
                    obj = objects[0]
		    if obj_key := obj.get("Key"):
                        return obj_key
                    else:
                        raise Exception(f"No object key present for {source_name}")
		else:
		    raise Exception(f"No sample data files present for {source_name}")
		
```
Going forward I will try to use them more when the code follows a `if x exists then do something` pattern.

jim-sheldon:
Any thoughts on a better name for `batch.py`? I didn't know what to call it and don't really like it.

iamleeg:
Maybe `job_queue` or `ingestor_queue`? BTW you might want to amend the last couple of commit messages :)

abhidg:
@iamleeg yes, without `list:true` cases are invisible to the curator portal, for UUID sources, it's safe to add list:true at ingestion time, as that's what prune-uploads does. For non-UUID sources, something similar to prune will have to be setup.

Issue #2138, opened 2021-09-01
Variance in Scotland data from official stats


abhidg:
Last ingestion reported more than 800,000 cases, but according to
https://www.gov.scot/publications/coronavirus-covid-19-daily-data-for-scotland/:

a total of 2,770,674 people in Scotland have been tested at least once. Of these people:
**436,688** have tested positive
2,333,986 were confirmed negative

This probably means data for the same person is being repeated; in which case 800,000 is the number of positive tests, or it could be a bug in the parser.

Source URL for Scotland is https://www.opendata.nhs.scot/dataset/b318bddf-a4dc-4262-971f-0ba329e09b87/resource/9393bd66-5012-4f01-9bc5-e7a10accacf4/download/trend_agesex_YYYYMMDD.csv

jim-sheldon:
Once I finish https://github.com/globaldothealth/list/issues/1939 this will become easier to test.

joe-brilliant:
thanks @jim-sheldon !

joe-brilliant:
@jim looks like #1939 is in Proxy so might move this there, unless there's any reason to keep in Strongcat?

iamleeg:
This is straightforward double-counting. The source data includes counts for age ranges 0-14, 15-29, ..., 85+ but it _also_ includes counts for 0-59, 60+. If you look on any day, the sum of the counts for _each_ of those groups of age range adds up to the total cases for that day. We're ingesting both, and getting two lines for every case (i.e. a 0-14 year old appears in the line list as one case age 0-14 and one case age 0-59).

@Mougk guessing the preference is to change the parser to only count the more specific age ranges and to drop existing cases for 0-59 and 60+ age brackets, is that correct?

PR #2173, opened 2021-10-01
Exclude overlapping age ranges from scotland data #2138

iamleeg:
I never got an answer to my question on #2138 about whether this is the correct way to fix the double-counting on the Scotland data, but I'm going to assume it is and go ahead :). This is only a partial fix: we also need to drop the incorrect cases that have already been imported. If you think I should attach that as a migration to this PR, please do reject this change :).


AnyaLindstromBattle:
Sorry about that @iamlee, @fojackson8 wrote the Scotland parser so may be best placed to review this?

abhidg:
No need for migration, new parser will drop existing cases :)

abhidg (approved):

PR #2214, opened 2021-10-29
export tsv json

abhidg:
- export: Drop .value fields for some event categories
- export: Add support for TSV,JSON,JSONL exports

### DIFF
--- a/data-serving/scripts/export-data/country_export.sh
+++ b/data-serving/scripts/export-data/country_export.sh
@@ -2,8 +2,13 @@
 
 set -euo pipefail
 
+FORMAT="${FORMAT:-csv,tsv,json}"
 SLUG=$(echo "$COUNTRY" | sed "s/ /_/g;s/[.,']//g" | tr '[:upper:]' '[:lower:]')
 QUERY="{\"list\": true, \"location.country\": \"$COUNTRY\"}"
 mongoexport --query="$QUERY" --uri="$CONN" --collection=cases \
-    --fieldFile=fields.txt --type=csv | python3 transform.py | gzip > "$SLUG.csv.gz"
-aws s3 cp "$SLUG.csv.gz" s3://covid-19-country-export/csv/
+    --fieldFile=fields.txt --type=csv | python3 transform.py -f "$FORMAT" "$SLUG"
+
+for fmt in ${FORMAT//,/ }
+do
+    aws s3 cp "$SLUG.$fmt.gz" s3://covid-19-country-export/$fmt/
+done
--- a/data-serving/scripts/export-data/transform.py
+++ b/data-serving/scripts/export-data/transform.py
@@ -7,12 +7,16 @@
 import sys
 import csv
 import json
+import gzip
 import logging
+import argparse
 from functools import reduce
+from contextlib import contextmanager
 from typing import Any, Optional
 
 logging.basicConfig(filename="transform.log", level=logging.INFO)
 
+VALID_FORMATS = ["csv", "tsv", "json"]
 
 __ARRAYS = [
     "caseReference.uploadIds",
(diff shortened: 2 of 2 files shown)


data-serving/scripts/export-data/country_export.sh
```diff
@@ -2,8 +2,13 @@
```
jim-sheldon, line 2: Does this need to be a shell script? It would be easier to test, maintain, and modify as a Python script.
Especially if it's a shell script that calls a Python script.

data-serving/scripts/export-data/transform.py
```diff
@@ -84,10 +86,12 @@ def convert_event(event: dict[str, Any]) -> dict[str, Any]:
     suffix = event["name"]
     col_name = f"events.{suffix}"
 
-    return {
-        f"{col_name}.value": event.get("value"),
+    o = {
```
jim-sheldon, line 89: OK to add meaningful variable names? It makes code easier to read and work with.

data-serving/scripts/export-data/transform.py
```diff
@@ -220,13 +224,88 @@ def convert_row(row: dict[str, Any]) -> Optional[dict[str, Any]]:
     return row
 
 
+class Writer:
+    "Generic writer class for different formats"
+
+    VALID_FORMATS = ["csv", "tsv", "json", "jsonl"]
```
jim-sheldon, line 230: Not sure `jsonl` is needed. AFAIK it's only referred to in this file.

data-serving/scripts/export-data/transform.py
```diff
@@ -220,13 +224,88 @@ def convert_row(row: dict[str, Any]) -> Optional[dict[str, Any]]:
...
+
+    def __init__(self, fmt, output, fields):
+        self.fmt = fmt
+        if fmt not in self.VALID_FORMATS:
```
jim-sheldon, line 234: Can use [choices parameter](https://docs.python.org/3/library/argparse.html#choices) instead.
We also might want to consider a better argument parsing library like [click](https://click.palletsprojects.com/en/8.0.x/) at some point.

data-serving/scripts/export-data/transform.py
```diff
@@ -220,13 +224,88 @@ def convert_row(row: dict[str, Any]) -> Optional[dict[str, Any]]:
...
+        if fmt == "csv":
+            self.writer = csv.DictWriter(
+                self.file, fieldnames=fields, extrasaction="ignore"
+            )
+        elif fmt == "tsv":
+            self.writer = csv.DictWriter(
+                self.file, fieldnames=fields, extrasaction="ignore", delimiter="\t"
+            )
+        else:
+            self.writer = None
```
jim-sheldon, line 246: Better to use a context manager (i.e. `with open()...`) and error handling. I also think this code becomes cleaner _without_ the use of classes.

data-serving/scripts/export-data/transform.py
```diff
@@ -220,13 +224,88 @@ def convert_row(row: dict[str, Any]) -> Optional[dict[str, Any]]:
...
+        """
+        if self.fmt == "jsonl":
+            self.writerow(row)
+        elif self.fmt in ["csv", "tsv"]:
+            self.writer.writeheader()
+            self.writer.writerow(row)
+        elif self.fmt == "json":
+            self.file.write("[\n " + json.dumps(row, sort_keys=True) + "\n")
+        else:
+            pass
```
jim-sheldon, line 263: `else: pass` can be deleted and have the same effect.

jim-sheldon:
This is a great use case for end-to-end testing :)
Arrange: create S3 bucket, stand up stack
Act: ingest data and run export script
Assert: S3 bucket contains exported files containing valid data

data-serving/scripts/export-data/transform.py
iamleeg: Does seem this is a candidate for Strategy pattern, as this `if fmt==…` gets replicated in multiple methods.

data-serving/scripts/export-data/country_export.sh
abhidg: The previous implementation used mongoexport as a subprocess -- I have used a bash script to keep the dependencies in Python to zero, and as aws-cli has a simple interface that I'm planning on using in the full-export script, it keeps them similar.

data-serving/scripts/export-data/transform.py
abhidg: Thanks for the contextmanager suggestion, and reworking it without classes has made the code clearer!

data-serving/scripts/export-data/country_export.sh
jim-sheldon: That makes sense for Lambda, to limit the layers, and I think for Batch moving away from bash makes more sense. Not a blocker by any means.

jim-sheldon (approved):

jim-sheldon:
Well done!

Issue #2191, opened 2021-10-14
Data export broken


iamleeg:
This is related to daily downloads. The export job is timing out because of the amount of data we request.

PR #2220, opened 2021-11-03
2191 full export

abhidg:
- export: Add full export scripts #2191
- ci: Build full-exporter images
- curator/api: Use latestdata-csv as we support multiple formats

### DIFF
--- a/data-serving/scripts/export-data/common.sh
+++ b/data-serving/scripts/export-data/common.sh
@@ -1,0 +1,3 @@
+ECR=612888738066.dkr.ecr.us-east-1.amazonaws.com
+# ingestion role contains necessary permissions to access S3 buckets
+JOB_ROLE_ARN="arn:aws:iam::612888738066:role/gdh-ingestion-job-role"
--- a/data-serving/scripts/export-data/full_export.sh
+++ b/data-serving/scripts/export-data/full_export.sh
@@ -1,0 +1,34 @@
+#!/bin/bash
+# Full data export
+# Depends: aws-cli
+
+set -xeuo pipefail
+
+# https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html#cli-usage-retries-configure
+export AWS_MAX_ATTEMPTS=10
+
+COUNTRY_EXPORT_BUCKET="s3://covid-19-country-export"
+FULL_EXPORT_BUCKET="s3://covid-19-data-export"
+
+# Unlike country_export.sh, here FORMAT only takes one value
+# Multiple formats are exported using separate AWS Batch job definitions
+FORMAT="${FORMAT:-csv}"
+FILE="latestdata-$FORMAT.tar"
+FILE_DATE="$(date '+%Y-%m-%d')-$FORMAT.tar"
+
+DIR="$(mktemp -d)"
+echo "Starting full export for format: $FORMAT"
+echo "Made temporary directory $DIR"
+trap "rm -rf $DIR" EXIT  # Cleanup before exit
+
+cp data_dictionary.txt citation.txt "$DIR"
+
+mkdir "$DIR/country"
+echo "Downloading files from $COUNTRY_EXPORT_BUCKET..."
+aws s3 sync $COUNTRY_EXPORT_BUCKET/$FORMAT/ "$DIR/country"
+echo "Preparing tarball..."
+tar cf $FILE -C "$DIR" .
+echo "Uploading to $FULL_EXPORT_BUCKET..."
+aws s3 cp $FILE $FULL_EXPORT_BUCKET/latest/$FILE
+aws s3 cp $FULL_EXPORT_BUCKET/latest/$FILE $FULL_EXPORT_BUCKET/archive/$FILE_DATE
+echo "Done!"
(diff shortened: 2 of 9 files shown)


data-serving/scripts/export-data/full_export.sh
```diff
@@ -0,0 +1,32 @@
...
+cp citation.txt "$DIR"
+
+mkdir "$DIR/country"
+echo "Downloading files from $COUNTRY_EXPORT_BUCKET..."
+aws s3 sync $COUNTRY_EXPORT_BUCKET/$FORMAT/ "$DIR/country"
+echo "Preparing tarball..."
+tar cf $FILE -C "$DIR" .
+echo "Uploading to $FULL_EXPORT_BUCKET..."
+aws s3 cp $FILE $FULL_EXPORT_BUCKET/latest/$FILE
+aws s3 cp $FILE $FULL_EXPORT_BUCKET/archive/$FILE_DATE
```
jim-sheldon, line 31: Could this be a Python script instead? These longer-running `aws` operations could benefit from error handling (including a message about what went wrong) and retry logic in case of network issues.

data-serving/scripts/export-data/setup_full_export.sh
```diff
@@ -0,0 +1,24 @@
...
+    P+=", \"environment\": [{\"name\": \"FORMAT\", \"value\": \"$1\" }]"
+    P+='}'
+    echo "$P"
+}
+
+for fmt in tsv csv json; do
+    aws batch register-job-definition --job-definition-name "full-exporter-$fmt" \
+        --container-properties "$(containerprops "$fmt")" \
+        --timeout "attemptDurationSeconds=3600" --type container
+done
```
jim-sheldon, line 24: For the same reasons as above I think this script would also be better as Python than Bash.

jim-sheldon:
Good stuff! I just have a couple of concerns around unhappy paths.

data-serving/scripts/export-data/common.sh
```diff
@@ -0,0 +1,3 @@
+ECR=612888738066.dkr.ecr.us-east-1.amazonaws.com
+# ingestion role contains necessary permissions to access S3 buckets
+JOB_ROLE_ARN="arn:aws:iam::612888738066:role/gdh-ingestion-job-role"
```
iamleeg, line 3: Is there a different approach we can use for configuration management? I would like changing the variables to be an explicit decision rather than something that happens when a PR gets merged or new software gets deployed, as different parts get deployed through different means.

data-serving/scripts/export-data/full_export.sh
abhidg: If it is about retry logic (which I should use, thanks!) then AWS CLI has retry logic built in which we can use https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-retries.html#cli-usage-retries-configure

All errors are logged to CloudWatch.

If it is about rewriting it in Python, I would prefer not to for this case, as I would have to duplicate the S3 sync functionality in Python, for unclear advantages.

jim-sheldon: It is about retry logic and error handling, to be sure, and also about maintainability and extensibility.
Great that `aws-cli` has retry logic built in; of course we should use that.
OK that errors are logged to CloudWatch. The benefit of doing our own exception handling is that we can add searchable messages, metrics, and provide context to otherwise ambiguous error messages. Unfortunately some error messages don't actually provide much useful diagnostic information.
For this script, adding an `x` to `set -eou` would add at least context around which command failed.

Maintainability and extensibility may seem like unclear advantages, as it's impossible to predict what requirements will change. However, requirements absolutely will change, so using tools that make it easy to adjust clearly wins over using tools that don't. This is what [SOLID programming](https://en.wikipedia.org/wiki/SOLID) is all about, and Bash doesn't really allow for that.

Imagine, further, that a bug is discovered around some behavior managed by Bash. Ideally we'd fix the bug and add an accompanying test to prove the bug is fixed and prevent back-sliding. Bash doesn't really help us in that regard. I like Bash as an adhesive between different system components (e.g. with Docker), but only because I don't know what else to use, and in general think it should not do any kind of "heavy lifting"--some scripting and/or systems language should do that, backed by the confidence given by a suite of automated tests.

Not a blocker, but this kind of approach (untested/untestable system-critical code) does feel a bit like an [unhedged bet](https://www.castsoftware.com/blog/bad-code-isnt-technical-debt-its-an-unhedged-call-option).

abhidg: Agreed with all your points, especially relating to larger pieces of code, for which Bash is unsuitable. I disagree that we have to move to Python now; as this in Batch, it is easier to swap out when requirements change. As it stands, this a short script which is basically a sync tar and copy. We can discuss this in call, if we want to not use Bash going forward in the project due to maintainability and extensibility concerns.

iamleeg (approved):

abhidg:
Added AWS_MAX_ATTEMPTS which should activate retrying logic in aws-cli, and copying YYYY-MM-DD file from S3->S3 rather than local -> S3 as it's much faster.

data-serving/scripts/export-data/full_export.sh
jim-sheldon: Apologies if I implied that we need to move to Python now -- we don't, and it's worth considering and discussing.
And if we're using Batch, which uses Docker, it makes sense to me to simply default to using Python containers. I understand why it was used in Lambda (to limit layers) but I don't think it makes sense going forward. 
For sure, let's talk about it tomorrow.

jim-sheldon (approved):

PR #2224, opened 2021-11-05
add country to sources

abhidg:
- curator: Add iso-3166-1 to dependencies
- schemas: Add countryCodes to sources
- curator: Add countryCodes field


jim-sheldon (approved):

iamleeg (changes requested):
Don't modify sources.schema.json, add a migration. Otherwise the historical information that the countryCodes field was added is missing from history, and it's likely that production and local databases would get out of sync as the prod db already thinks it's migrated past this level.

jim-sheldon (approved):

iamleeg (approved):
thank you!

PR #2225, opened 2021-11-07
Prune triggers country export

abhidg:
Adds a hook to prune-uploads to trigger appropriate exporter job definitions.

### DIFF
--- a/.github/workflows/prune-uploads-dev.yml
+++ b/.github/workflows/prune-uploads-dev.yml
@@ -12,10 +12,10 @@ jobs:
         working-directory: data-serving/scripts/prune-uploads
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Python 3.8
+    - name: Set up Python 3.9
       uses: actions/setup-python@v2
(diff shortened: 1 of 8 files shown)


.github/workflows/prune-uploads-dev.yml
```diff
@@ -12,10 +12,10 @@ jobs:
         working-directory: data-serving/scripts/prune-uploads
     steps:
     - uses: actions/checkout@v2
-    - name: Set up Python 3.8
+    - name: Set up Python 3.9
```
jim-sheldon, line 15: Better to go straight to 3.10?

jim-sheldon (approved):

.github/workflows/prune-uploads-dev.yml
iamleeg: 3.9 is not EOL until [2025](https://devguide.python.org/#status-of-python-branches) so no rush on this.

iamleeg (approved):

PR #2228, opened 2021-11-08
Add hooks in prune, support multiple environments in export

abhidg:
We don't need to run export on dev, so disable hooks by default.


jim-sheldon (approved):

iamleeg:
We don't need to, but should we? Would we get earlier indication of failures, or the ability to test features that depend on export?

abhidg:
@iamleeg Good point, I’ll need to make some changes to the export scripts to accommodate multiple environments

abhidg:
@iamleeg updated to support multiple environments

iamleeg (approved):

PR #2235, opened 2021-11-12
1875 denormalise confirmed date

iamleeg:
I've added an index on confirmed date. To do that I've added a migration to extract the date on the confirmed event and store it in a top-level field, and I've added mongoose middleware to keep that in sync for new additions/changes. This means that there are no changes to ingestion scripts or the APIs, the app works as previously but the data-service does the heavy lifting of keeping this extra field in sync.

The reason I've gone for mongoose middleware here is simply that mongo itself doesn't have database triggers (mongodb atlas does, via realm, but then we couldn't actually synchronise those triggers with our existing migration tools). I could've written express middleware to rewrite the requests in the same way that our creation metadata works, but then there's more work remembering that any new method needs to borrow that middleware. To be clear I don't think the existing approach to revision metadata is wrong, because it needs access to the user property on the request which wouldn't be available in mongoose middleware.

@maciekz1996 let me know if this unblocks you on sorting by date, or if you need something more.

### DIFF
--- a/data-serving/data-service/src/controllers/case.ts
+++ b/data-serving/data-service/src/controllers/case.ts
@@ -1,4 +1,4 @@
-import { Case, CaseDocument, RestrictedCase } from '../model/case';
+import { Case, CaseDocument, caseWithDenormalisedConfirmationDate, RestrictedCase } from '../model/case';
 import { EventDocument } from '../model/event';
 import caseFields from '../model/fields.json';
 import { Source } from '../model/source';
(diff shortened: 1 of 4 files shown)


abhidg (approved):
:tada: For this, might be better to run the migration manually, as Actions will probably timeout on 60m cases.

iamleeg:
Thanks! I'll do what I did last time: I'll merge now, and I'll watch whether the action fails on main. If it does, I'll run the migration manually on dev.

PR #2238, opened 2021-11-15
Removing string template

jim-sheldon:
Filtered downloads (country only) throw `There was an error while downloading data, please try again later. TypeError: o.headers['content-disposition'] is undefined`. Full data set downloads and non-country-only filtered downloads (e.g. country + gender) work as expected.


iamleeg (approved):
I don't see how the change proposed solves the problem described, but if it works then go for it :)

jim-sheldon:
> I don't see how the change proposed solves the problem described, but if it works then go for it :)

Same, but it was the only difference I could see from looking at the code and S3 objects. Bad idea but easy to test.

PR #2241, opened 2021-11-15
export: Fix first line being dropped in output

abhidg:
First line was being dropped in the output due to two
csv.DictReader() calls, the first to scan the headers,
the second to actually read through the file. Header scanning
read the first row, which was then skipped in the main loop.

* Change first DictReader() call to readline()
* Add option -i (--input) to specify input file instead of stdin
* Add option to use stdout as output (specify -)
* Add tests for empty and non-empty mongoexport output


data-serving/scripts/export-data/transform.py
```diff
@@ -271,30 +274,54 @@ def open_writers(formats: list[str], fields: list[str], output: str):
         print("Error occurred in open_writers():")
         print(e)
     finally:
+        if output == '-':
+            return
         for fmt in files:
             if fmt == "json":
                 files[fmt].write("]\n")
             files[fmt].close()
```
jim-sheldon, line 282: Should this be `for fmt in formats` and `if files[fmt] == "json"`?

data-serving/scripts/export-data/transform.py
```diff
@@ -271,30 +274,54 @@ def open_writers(formats: list[str], fields: list[str], output: str):
...
+        yield f
+    except Exception as e:
+        print("Error occurred in open_input():")
+        print(e)
+    finally:
+        f.close()
+
+
+def transform(output: str, formats: list[str], input: str = None):
+    with open_input(input) as inputfile:
```
jim-sheldon, line 298: Simpler as just `with open`? Then you get the call to `close` for free.

data-serving/scripts/export-data/transform.py
```diff
@@ -271,30 +274,54 @@ def open_writers(formats: list[str], fields: list[str], output: str):
...
+        reader = csv.DictReader(inputfile, fieldnames=headers)
+        hasrows = False
+        with open_writers(formats, fields, output) as writers:
+            for i, row in enumerate(map(convert_row, reader)):
+                hasrows = True
+                writerow(formats, writers, row, i)
+        if output != '-' and not hasrows:  # cleanup empty files
+            for fmt in formats:
+                with suppress(FileNotFoundError):
+                    os.remove(f"{output}.{fmt}.gz")
```
jim-sheldon, line 309: Simpler as `if exists then remove`?

data-serving/scripts/export-data/transform.py
abhidg: Thanks, didn't realise sys.stdin could be used as a contextmanager.

data-serving/scripts/export-data/transform.py
abhidg: files[fmt] refers to file objects

data-serving/scripts/export-data/transform.py
abhidg: Path().unlink(missing_ok=True) is even better :)

jim-sheldon (approved):

iamleeg (approved):

PR #2242, opened 2021-11-16
2012 aggregate script to batch

iamleeg:
Got this script working correctly against S3 in localstack, so putting up the PR so you can comment while I'm still working out how to define the job in batch for it. AFAICT I need to create a Docker container for the script, a GitHub action to push that image to Amazon ECR, and two batch job definitions with a container spec that references the image and contains the environment variables for dev and prod respectively.


data-serving/scripts/aggregate/aggregate/app.py
```diff
@@ -354,40 +358,43 @@ def generate_total_json():
...
+    # S3 endpoint is allowed to be None (i.e. connect to default S3 endpoint),
+    # it's also allowed to be not-None (to use localstack or another test double).
+    s3_endpoint = os.environ.get("S3_ENDPOINT", None)
+    print("Logging into MongoDB...")
+    client = pymongo.MongoClient(connection_string)
+    db = client.covid19
+    cases = db.cases
+    print("And we're in!")
+
+    date_string = datetime.datetime.now().strftime("%m-%d-%Y")
```
iamleeg, line 390: I calculate this once in case the script ever runs over midnight so it uses a consistent date for one invocation.

data-serving/scripts/aggregate/aggregate/app.py
```diff
@@ -354,40 +358,43 @@ def generate_total_json():
...
-          https://docs.aws.amazon.com/lambda/latest/dg/python-context-object.html
-    """
+if __name__ == '__main__':
+    connection_string = os.environ.get("CONN")
+    line_list_url = os.environ.get("LINE_LIST_URL")
+    map_url = os.environ.get("MAP_URL")
+    s3_bucket = os.environ.get("S3_BUCKET")
+    # S3 endpoint is allowed to be None (i.e. connect to default S3 endpoint),
+    # it's also allowed to be not-None (to use localstack or another test double).
+    s3_endpoint = os.environ.get("S3_ENDPOINT", None)
```
jim-sheldon, line 383: Really only a style thing: these could be constants, and the default argument for `get` is `None`.

jim-sheldon (approved):

data-serving/scripts/aggregate/aggregate/app.py
iamleeg: Ah, yes it is! I was expecting it to be `KeyError`. OK, I'll change this to explicitly test for the ones that must be set.

iamleeg:
I have a docker container and a workflow to publish it. I _think_ I have to do the rest from the AWS console once the image is published, and there isn't an infrastructure-as-code way to set up the batch job within this repo, is that correct?

abhidg (approved):

abhidg:
There isn’t an automated system in place yet. There are some shell scripts in export which sets up the batch job definitions for that case. In practice, most of the AWS console web stuff can be done through aws-cli.

PR #2243, opened 2021-11-19
Fix bug in generating API keys that made them impossible to find on use

iamleeg:
The API keys are supposed to be prefixed by the user ID so that we can easily find the correct user to test the key against, but this was not being formatted correctly by the auth controller.

### DIFF
--- a/verification/curator-service/api/src/controllers/auth.ts
+++ b/verification/curator-service/api/src/controllers/auth.ts
@@ -23,11 +23,12 @@ let isNewsletterAccepted: boolean;
 
 async function findUserByAPIKey(apiKey?: string): Promise<Express.User> {
     if (!apiKey) {
-        throw new Error("No API key");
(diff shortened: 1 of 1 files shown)


verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -306,25 +306,25 @@ export class AuthController {
                     }
                     // prefix the API key with the user ID to make it easier to find users by API key in auth
                     const randomPart = await getRandomString(32);
-                    currentUser.apiKey = `${theUser.id}${randomPart}`;
+                    currentUser.apiKey = `${theUser.id.toString()}${randomPart}`;
```
iamleeg, line 309: This is the one meaningful change in this commit. The rest was done by the husky pre-commit hook :-/

abhidg (approved):

PR #2244, opened 2021-11-19
migrations: Add suggested indexes for cases and users

abhidg:


iamleeg (approved):
Nice, the atlas profiler also shows that there are 16 indexes we have but aren't using, taking up multiple GBs of space. We should drop those (not saying as part of this PR, just it reminded me)

PR #2257, opened 2021-11-29
Add scripts to access API

abhidg:
Draft version, comments welcome! I haven't put up versioned dependencies as the libraries are fairly stable, and people might be stuck on older versions; also not sure what the standard is in R.﻿

### DIFF
--- a/.github/workflows/api-python-tests.yml
+++ b/.github/workflows/api-python-tests.yml
@@ -1,0 +1,31 @@
+name: API tests (Python)
+
+on:
+  push:
+    branches: [main]
+    paths:
+    - '.github/workflows/api-python-tests.yml'
+    - 'api/python/**'
+  pull_request:
+    paths:
+    - '.github/workflows/api-python-tests.yml'
+    - 'api/python/**'
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: api/python
+    steps:
+    - uses: actions/checkout@v2
+    - name: Set up Python 3.9
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.9
+    - name: Install dependencies
+      run: |
+        pip install -r requirements.txt
+        pip install pytest
+    - name: Run tests
+      run: pytest
--- a/api/python/gdh.py
+++ b/api/python/gdh.py
@@ -1,0 +1,100 @@
+# Depends: requests pandas
+# You may already have these installed, otherwise install using pip:
+#     pip install requests pandas
+
+# Example invocations:
+# >>> from gdh import get_cases
+# >>> key = "API KEY HERE"
+# >>> c1 = get_cases(key, country="New Zealand")
+# >>> c2 = get_cases(key, country="New Zealand", admin1="Auckland")
+#
+# Use get_cached_cases() to cache the cases locally. This is useful for
+# rerunning the script which will then use the cached version.
+
+import os
+import io
+import json
+import hashlib
+import datetime
+
+import requests
+import pandas as pd
+
+GDH_URL_DEFAULT = "https://data.covid-19.global.health"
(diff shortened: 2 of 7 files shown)


iamleeg (approved):
Looks good, bear in mind I don't understand R very much! Here's the next challenge: is there a way to build an integration test to ensure we don't regress the API?

jim-sheldon (approved):

jim-sheldon:
I also can't offer much around R...yet :)
Once the end-to-end testing environment is done I can give a demo and show how to add new tests.

If it's helpful I put a couple assert-like functions in [this file](https://github.com/globaldothealth/list/blob/main/dev/run_session.py) to do things like check the file for valid structure and contents.

api/python/gdh.py
```diff
@@ -0,0 +1,60 @@
...
+
+import os
+import io
+import json
+import hashlib
+
+import requests
+import pandas as pd
+
+GH = "https://data.covid-19.global.health"
```
jim-sheldon, line 22: Could this be an environment variable, with the current value as the default?
That way we could reuse it for other environments easily.

api/python/gdh.py
```diff
@@ -0,0 +1,60 @@
...
+downloadAsync = "/api/cases/downloadAsync"
+
+def cases(query, apikey, server=GH):
+    res = requests.post(
+        f"{server}{downloadAsync}",
+        data=json.dumps({"format": "csv", "query": query}),
```
jim-sheldon, line 28: "csv" might be better as a default `kwarg` value to extend this function to other formats

api/python/gdh.py
```diff
@@ -0,0 +1,60 @@
...
+        headers={"Content-Type": "application/json", "X-API-Key": apikey},
+    )
+    if res.status_code != 200:
+        print(res.text)
+        return False
```
jim-sheldon, line 33: I think a `raise` might be better here, so the return value of the function is of a single type

api/python/gdh.py
abhidg: `cases()` is about providing the case data as a dataframe without regard to the underlying format, I'll add a function to download in the specific format requested separately.

api/python/gdh.py
```diff
@@ -0,0 +1,101 @@
...
+    "admin1",
+    "admin2",
+    "admin3",
+    "variant",
+    "dateconfirmedafter",
+    "dateconfirmedbefore",
+}
+
+
+def qs(**kwargs):
```
jim-sheldon, line 49: Does this function turn filters into a string? If so, maybe a better name would be something like `stringify_filters`?

api/python/gdh.py
```diff
@@ -0,0 +1,101 @@
...
+
+def cases_cachefile(query, server=GDH_URL, folder="cache"):
+    to_hash = f"{query.strip()}|{server}"
+    sha256 = hashlib.sha256(to_hash.encode("utf-8")).hexdigest()
+    return f"{folder}/{sha256}.csv"
+
+
+# cases_cached returns a cached copy of cases if it exists,
+# otherwise saves to cache
+def cases_cached(query, apikey, server=GDH_URL, refresh=False, folder="cache"):
```
jim-sheldon, line 92: In general I find functions named as verbs and objects named as nouns to be a better pattern to follow...i.e. a function named something like `get_cached_cases` will do just that.

.github/workflows/api-python-tests.yml
```diff
@@ -0,0 +1,31 @@
...
+    paths:
+    - '.github/workflows/api-python-tests.yml'
+    - 'api/python/**'
+
+jobs:
+  tests:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: data-serving/scripts/prune-uploads
```
iamleeg, line 19: Surely a mistake?

PR #2277, opened 2021-11-30
Omicron dropdown

iamleeg:
When we inevitably decide that we want to track the omicron variant, merge this.

### DIFF
--- a/verification/curator-service/ui/.gitignore
+++ b/verification/curator-service/ui/.gitignore
@@ -5,6 +5,8 @@
 # Don't include CSVs, except for Cypress fixtures.
 *.csv
 !cypress/fixtures/*.csv
+# and the variants definitions
+!variantsofconcern.csv
 
 #amplify
 amplify/\#current-cloud-backend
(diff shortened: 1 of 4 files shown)


jim-sheldon (approved):

iamleeg:
Will wait for the tests to re-run: saw a failure on this branch earlier with no code changes and I'd like to see why. If it's the thing with Italy going missing I'll just merge it.

joe-brilliant:
Thanks for jumping on this so quickly all, let's go ahead and to production as soon as possible. Relatedly, we may want consider adding in all official WHO VoCs so we're fully up to date: https://www.who.int/en/activities/tracking-SARS-CoV-2-variants/

Would it be possible / make sense to add the greek names in parentheses or vice versa with the pango lineage?

iamleeg:
I actually did add the other VoCs, did you want the VoIs too? Can add the name, but I'd prefer to do it as a separate field so we can track unnamed variants if desired.

joe-brilliant:
I think just all VOCs are fine for now. Agreed "Pango Lineage" and "WHO Label" should be separate collumns but we can add that later. cc: @Mougk for his input. Would be great to get this change in to prod first thanks all!

PR #2285, opened 2021-12-02
#2266 - Filtering by gender: Not Provided

maciekz1996:
**Changes**
- Added the ability to filter cases that doesn't have `gender` provided
- Added test case for that


sergioloporto (approved):
excellent

abhidg (approved):

PR #2312, opened 2021-12-08
prune-uploads: Add aggregate hook to prune-uploads

abhidg:
Now that aggregate is fixed, we can hook it up to prune-uploads, which will avoid us having to manage another EventBridge rule, as well as making sure it runs *after* prune.


iamleeg (approved):
Looks good, I'd encourage adding logging around success as well as failure so we can distinguish "worked but something downstream failed" from "doesn't even run" :)

jim-sheldon:
Any downside to combining the scripts?

abhidg:
I think it's best to keep them separate for now, if we have to move away from prune (there's a support question with MongoDB on this atm)

Issue #1877, opened 2021-05-24
Location autocomplete doesn't work in case details


marlenachroscicka:
Autocomplete feature in data portal when adding new line list case (Location section) is not working. In the console I see that each request ends up with 500 status, it seems that the error cause is located in geocoding service and more precisely `list/geocoding/location-service/src/app/geocoder_suggester.py`. The request that is being sent looks like this: `GET http://localhost:3002/api/geocode/suggest?q=united` for example. Below is the stack trace containing error:

<img width="1285" alt="Screen Shot 2021-05-27 at 11 29 57" src="https://user-images.githubusercontent.com/44740625/122431498-cf012400-cf94-11eb-9f78-1eb32b7bd8fa.png">

iamleeg:
This is working now, closing the issue
<img width="792" alt="Screenshot 2021-12-16 at 16 04 14" src="https://user-images.githubusercontent.com/237254/146406355-8ff8b837-b334-47e3-bc4f-c0769a81608b.png">
.

PR #2322, opened 2021-12-10
geocoding: Return None if tile ID not found in admins collection

abhidg:
This is a possible fix for #1877

### DIFF
--- a/geocoding/location-service/src/app/admins_fetcher.py
+++ b/geocoding/location-service/src/app/admins_fetcher.py
@@ -1,9 +1,10 @@
 import json
+import logging
+
 import ratelimiter
 from lru import LRU
 
 from src.integration.mapbox_client import mapbox_tile_query
-
 
 class AdminsFetcher:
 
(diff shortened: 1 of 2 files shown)


abhidg:
Ok, so this fixes the 500 error, but the resulting geocodes miss certain fields. As an example Meath, Ireland geocodes at resolution Admin1, but there is no administrativeAreaLevel1 field in the geocode (since it's missing from the admins collection). The latitude and longitude seem to be OK.

jim-sheldon (approved):

geocoding/location-service/src/app/admins_fetcher.py
```diff
@@ -47,6 +47,9 @@ def getKey(self, layer):
         }[layer]
 
     def getName(self, id):
-        return self.admins.find_one({
-            'id': id
-        })['name']
+        admin = self.admins.find_one({'id': id})
+        if not admin:
+            print(f"Warning: ID {id} not found in admins collection, returning None")
```
jim-sheldon, line 52: Looks quite like a `logging.warning()` :)

iamleeg (approved):
With @jim-sheldon's comment about using the logger, this is fine.
There's more that can be done here: there's a script to populate the tile IDs in `verification/scripts` so it's worth getting the boundary JSON files from Mapbox and re-running the import (locally, to start with): it may be that we're using out of date data.

PR #2327, opened 2021-12-11
Stringifying JSON response data

jim-sheldon:


abhidg (approved):
Looks good -- does this affect the streaming? i.e. will line 585 be only executed once the entire Blob or encodeURIComponent() has finished executing?

jim-sheldon:
It does not affect streaming; the browser will block until the response completes. This is observable currently in production.

PR #2332, opened 2021-12-13
Checking for keys before writing CSV

jim-sheldon:
Last run failed due to a KeyError. Also added some more logging.


data-serving/scripts/aggregate/aggregate/app.py
```diff
@@ -110,19 +110,22 @@ def get_jhu_counts():
...
+    retries = 5
+    current_attempt = 0
+    while req.status_code != 200 and current_attempt < retries:
         logging.info("Got status " + str(req.status_code) + " for '" + url + "'")
         date = date - datetime.timedelta(days=1)
         now = date.strftime("%m-%d-%Y")
         logging.info(f"Checking for JHU data on {now}")
         url = f"https://raw.githubusercontent.com/CSSEGISandData/COVID-19/master/csse_covid_19_data/csse_covid_19_daily_reports/{now}.csv"
         req = requests.head(url, timeout=10)
+        current_attempt += 1
```
iamleeg, line 128: Two things here, one is definitely a problem and the other I don't understand.

1. This loop doesn't handle the failure condition at all. If the repo renames master to main, or deletes this file, or GitHub goes down, this will just receive an error five times and then carry on as normal. There should definitely be an early exit here if we don't actually get the data. Looking at the rest of the code it would be possible to return an empty Pandas series and then not have JHU data downstream, but that might not be desirable.
2. I don't see the value in testing availability with a HEAD request if we are going to unconditionally GET the data anyway, which is the thing I'm confused by. I would use HEAD if we expected a large response and wanted to test whether we could use cached data, or needed to pre-allocate resources before the download. As it is, this is testing that we _could_ download the data, then making a separate request to download it (which could still fail). I _think_ it's better to just try grabbing the data with a GET in this loop and using it if successful: unless there's another reason to prefer HEAD that I missed.

data-serving/scripts/aggregate/aggregate/app.py
```diff
@@ -141,13 +144,15 @@ def get_jhu_counts():
     )
     jhu_counts = jhu_counts.set_index("Country_Region")
     jhu_counts = pd.Series(jhu_counts.values.flatten(), index=jhu_counts.index)
+    logging.debug("Got aggregated counts from JHU")
```
iamleeg, line 147: This line seems like it's at the same level of importance as the two above, which have `info` severity.

data-serving/scripts/aggregate/aggregate/app.py
```diff
@@ -210,10 +217,11 @@ def generate_country_json(cases, s3_endpoint, bucket, date, line_list_url, map_u
             record["lat"] = country_codes[country]['latitude']
             record["long"] = country_codes[country]['longitude']
         except KeyError:
-            raise ValueError(f"Could not find country in list of country codes: {country}")
+            logging.critical(f"Could not find country in list of country codes: {country}")
```
iamleeg, line 220: Is this critical? The script ignores the error, maybe it's just a warning (or the script should do more)?

iamleeg (changes requested):

data-serving/scripts/aggregate/aggregate/app.py
jim-sheldon: Before it was failing on countries with relatively low numbers of reported cases, so it seemed like the right thing to do was to let the script continue.

data-serving/scripts/aggregate/aggregate/app.py
jim-sheldon: Updated.

data-serving/scripts/aggregate/aggregate/app.py
jim-sheldon: Yes, and I added an early exit. New logic is: check for today's file. If it does not exist, check for yesterday's file. If that does not exist, then exit. Good point about the HEAD request being redundant. I agree, the file is only ~500 kb. I removed those and use GETs now.

iamleeg (approved):

PR #2338, opened 2021-12-15
Update insecure dependencies in ingestion/functions

iamleeg:


iamleeg:
BTW somehow the chardet package was missing from the pyproject which broke the tests (maybe it was transiently included before?) so I added that too.

jim-sheldon (approved):

PR #2342, opened 2021-12-17
Logging non-default line list filters those causing 500+ errors

jim-sheldon:
Not sure if this is as detailed as we'd like, but it's a start.

### DIFF
--- a/verification/curator-service/api/src/controllers/cases.ts
+++ b/verification/curator-service/api/src/controllers/cases.ts
@@ -10,6 +10,10 @@ import crypto from 'crypto';
 //   https://github.com/globaldothealth/list/issues/961.
 axios.defaults.timeout = 0;
 
+// The query (URL) when the line list first loads
+const defaultInputQuery = '/cases';
+const defaultOutputQuery = '/cases/?limit=50&page=1&count_limit=10000&sort_by=default&order=ascending';
+
 /**
  * CasesController mostly forwards case-related requests to the data service.
  * It handles CRUD operations from curators.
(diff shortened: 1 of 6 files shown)


verification/curator-service/api/src/controllers/cases.ts
```diff
@@ -10,6 +10,9 @@ import crypto from 'crypto';
 //   https://github.com/globaldothealth/list/issues/961.
 axios.defaults.timeout = 0;
 
+// The query (URL) when the line list first loads
+const defaultQuery = '/cases/?limit=50&page=1&count_limit=10000&sort_by=default&order=ascending';
+
```
iamleeg, line 15: this seems brittle as the UI and API need to agree on this query, I can imagine us changing it in the UI and forgetting to update this line. Perhaps the UI can send a sentinel query, e.g `?limit=default&page=default&…` and the API tests for whether the qs contain any non-default fields?

Alternatively, the UI _doesn't_ request any query parameters by default, and the default value in the API is the one the UI expects. API customers can still request other limits/pages/counts etc by explicitly setting them, at which point we log their query.

verification/curator-service/api/src/controllers/cases.ts
```diff
@@ -22,10 +25,16 @@ export default class CasesController {
...
+        if (req.url !== defaultQuery) {
+            logger.info(`Filter applied: ${req.url}`);
+        }
         try {
             const response = await axios.get(
                 this.dataServerURL + '/api' + req.url,
             );
+            if (response.status >= 500) {
+                logger.error(`A server error occurred when trying to list data using URL: ${req.url}. Response status code: ${response.status}`);
+            }
```
iamleeg, line 37: maybe log anything that isn't a 2xx?

jim-sheldon: I defaulted to 5XX because I wasn't sure whether we wanted to log user errors, but I think 4XX and above makes sense as the web UI really is the user in this case.
I also think ignoring 3XX makes sense, unless we care about redirects or cached responses?

verification/curator-service/api/src/controllers/cases.ts
jim-sheldon: I agree. Thanks for providing those suggestions; I think the second one makes the most sense as it keeps the responsibility of the definition of "default" internal. Going to pursue that.

verification/curator-service/api/src/controllers/cases.ts
iamleeg: ah yes good catch, we probably don't need to log 3xx

iamleeg: but 400 is good as we've seen those when we have mismatched deployments

iamleeg (approved):

PR #2343, opened 2021-12-17
Replacing data script print statements with logging

jim-sheldon:
Let me know if any of the `setup_logger` functions should change.

### DIFF
--- a/data-serving/scripts/convert-data/convert_data.py
+++ b/data-serving/scripts/convert-data/convert_data.py
@@ -1,7 +1,7 @@
-'''
-Script to convert CSV line-list data into json compliant with the MongoDB
-schema.
-'''
+#!/usr/bin/env python3
+
+#Script to convert CSV line-list data into json compliant with the MongoDB schema.
+
 
 import argparse
 import csv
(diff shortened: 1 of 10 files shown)


data-serving/scripts/convert-data/convert_data.py
```diff
@@ -1,7 +1,7 @@
-'''
```
iamleeg, line 1: AFAIK this script is only run manually so this wasn't needed, but I suppose consistency will act as a signal to us to remember to do logging this way in future.

data-serving/scripts/prune-uploads/hooks/country_export.py
```diff
@@ -1,15 +1,16 @@
 # Exports country specific data by submitting
 # corresponding exporter_* job definitions
 
+from functools import cache
```
iamleeg, line 4: I don't see what motivated this line or line 8 additions, was that your IDE?

iamleeg:
I don't think all of these scripts have automatic tests, could you describe your test plan for the ones not covered by CI please?

data-serving/scripts/prune-uploads/hooks/country_export.py
jim-sheldon: In these scripts imports were not separated by group (standard library, third party, local library) per https://www.python.org/dev/peps/pep-0008/#imports so I changed those, and in the process reorganized groups in alphabetical order (PEP8 is silent on this, but it makes it easier for me to read).

jim-sheldon:
> I don't think all of these scripts have automatic tests, could you describe your test plan for the ones not covered by CI please?

Good point and question. I think the right way to handle this is for me to open an issue for each script to refactor, add testing for the script and make tests run automatically. The ones that do have testing could benefit from this as well.

Further, CI does not report on neither coverage or line hits/misses, and does not fail on PEP8 or CCN violations. I would like to add those to the pipeline as well, as as we have at least coverage and linting gating builds for TS. I would treat this work as a separate issue.

iamleeg (approved):

Issue #2349, opened 2021-12-21
Events outcome.value field not parsing correctly for Colombia or Argentina


fojackson8:
Colombia and Argentina are both rich datasets that provide case events + timestamp in good detail. We want to capture case outcome and the date that outcome was confirmed. For both countries, **case outcome information is provided in the original datasets, is extracted in the parser, but then is missing in the final G.h database.**

For all parsers, case outcome {death; recovered} is parsed as the value of the event 'outcome', and is provided alongside the relevant date, eg for Colombia:

```
 if entry["fecha_fallecimiento"]:
                    case["events"].append({
                        "name": "outcome",
                        "value": "Death",
                        "dateRange": {
                            "start": convert_date(entry['fecha_fallecimiento']),
                            "end": convert_date(entry['fecha_fallecimiento']),
                        }
                    })

```
For both Argentina and Colombia, these parsers seem to capture this information, but the information is not available in the G.h data download / API download.

Below is a screenshot of when I run the Colombia parser locally - it shows the first couple JSON dicts that are yielded from the parser. As you'll see, the events outcome value field is succesfully populated, which shows the parser is working as expected (at least locally). This leads me to suspect this information is being lost somewhere after parsing but before entering the final database.

![image](https://user-images.githubusercontent.com/23453141/146920259-1aa97757-0d12-4981-8d3c-118e8716ee1a.png)

PR #2350, opened 2021-12-21
export: Fix bug where events.*.value is always null

abhidg:
In transform.py, both the input and output variables were named
event, which caused all event.get("value") to be None. Renamed
output variable to fix this. Also removed unused imports and
added a test.

Fixes: #2349


iamleeg (approved):

jim-sheldon (approved):
Not in scope for this bugfix, but adding a test for JSON input and output would be good.

Issue #2354, opened 2021-12-21
Downloaded JSON files should not include an 'events' list


jim-sheldon:
Downloading a country-specific data set returns a flattened list of cases, but also contains an 'events' list whose data is already provided in `events.EVENT_NAME` fields.

PR #2355, opened 2021-12-23
2318 denormalize schema on filtered downloads

jim-sheldon:
This PR removes the 'events' column from CSV and TSV files and replaces it with a set of flattened events date and value columns.
Most of the code here flattens the JSON file so that it matches the JSON file created by the export script, with one exception, which I made this issue for: https://github.com/globaldothealth/list/issues/2354


jim-sheldon:
After these changes, some discrepancies exist between JSON made by the service and JSON made by the export script:

Keys in JSON from export script not in JSON from service: ['events', 'genomeSequences']

Keys in JSON from service not in JSON from export script: ['variantOfConcern', 'events.firstClinicalConsultation.date', 'events.icuAdmission.date', 'events.outcome.value', 'travelHistory.travel.location.country', 'travelHistory.travel.location.geometry.coordinates', 'events.firstClinicalConsultation.value', 'travelHistory.travel.location.administrativeAreaLevel2', 'travelHistory.travel.location.administrativeAreaLevel1', 'events.outcome.date', 'events.onsetSymptoms.value', 'events.selfIsolation.date', 'events.icuAdmission.value', 'travelHistory.travel.location.administrativeAreaLevel3', 'events.selfIsolation.value']

The 'events' fields are a bug in the data export script, and can be dealt with separately.

I could use some guidance on the next two:
The 'genomeSequences' field (really a document) is not used anywhere, and I can't set it to an empty string due to strong typing. If I set it to undefined it does not write the key with an empty value to the JSON file. OK to simply remove it from the fields file?

The SGTF field (a number) is absent from both the JSON from the export script and the JSON from the service, but it is in the fields file. I also can't set it to an empty string due to strong typing, and setting it to undefined results in it missing from the JSON file. Should I change its type to `Number | String`? That feels wrong; is there a better way?

The travelHistory fields do not exist in the fields file; I will need to remove those from the code that flattens the fields.

I would like to make https://github.com/globaldothealth/list/issues/2103 a higher priority. Not only does the fields file represent a schema for exported data (and as such belongs in the database), it lacks features normal database use provides, leaving us to write, maintain, and troubleshoot those features ourselves.

iamleeg:
1. @Mougk says we'll be working more with genomic data this year. On that basis I say yes to your suggestion: remove the unused `geneSequences` field. When we actually know what we're doing, we can add back something better.
2. The SGTF field should be an optional number (0 or 1 are the only valid values), not a string. Looks like I forgot to add it to the OpenAPI schema, sorry.

Mougk:
Perhaps add an NA to SGTF (UK line list for example has 0,1, and NA).

data-serving/data-service/test/util/case.test.ts
```diff
@@ -63,4 +83,673 @@ describe('Case', () => {
...
+            'revisionMetadata.creationMetadata.curator','revisionMetadata.creationMetadata.date',
+            'revisionMetadata.creationMetadata.notes','revisionMetadata.editMetadata.curator',
+            'revisionMetadata.editMetadata.date','revisionMetadata.editMetadata.notes','revisionMetadata.revisionNumber',
+            'SGTF', 'symptoms.status','symptoms.values','transmission.linkedCaseIds','transmission.places',
+            'transmission.routes','travelHistory.travel.dateRange.end','travelHistory.travel.dateRange.start',
+            'travelHistory.travel.location.name','travelHistory.travel.methods','travelHistory.travel.purpose',
+            'travelHistory.traveledPrior30Days','vaccines.0.name','vaccines.0.batch','vaccines.0.date',
+            'vaccines.0.sideEffects','vaccines.1.name','vaccines.1.batch','vaccines.1.date','vaccines.1.sideEffects',
+            'vaccines.2.name','vaccines.2.batch','vaccines.2.date','vaccines.2.sideEffects','vaccines.3.name',
+            'vaccines.3.batch','vaccines.3.date','vaccines.3.sideEffects',''];
```
abhidg, line 104: What's the last blank header for?

abhidg:
Looks good to me, pending #2356 :tada:

jim-sheldon:
I removed the `genomeSequences` field,  removed the extra travelHistory fields, added SGTF to the OpenAPI docs, and added 'NA' as the default SGTF value when the case does not contain that field.

data-serving/data-service/src/model/case.ts
```diff
@@ -60,7 +60,7 @@ export const caseSchema = new mongoose.Schema(
         restrictedNotes: String,
         pathogens: [pathogenSchema],
         list: Boolean,
-        SGTF: Number,
+        SGTF: mongoose.Schema.Types.Mixed,
```
iamleeg, line 63: What do you think about an enum of ['0', '1', 'NA'] and a type (below) of '0'|'1'|'NA'? On the downside the "numbers" 0 and 1 are represented as strings, on the upside the value is limited to the specific values supported.

jim-sheldon: I like it. I made the changes and added a migration to switch SGTF from int to string.

iamleeg (approved):

Issue #2223, opened 2021-11-04
Transform country data in line list to use ISO 3166-1 country codes


abhidg:
**Is your feature request related to a problem? Please describe.**
Currently country data is free text. This leads to issues like alternate names for countries
such as US, and bugs like #2221

**Describe the solution you'd like**
Transition to https://en.wikipedia.org/wiki/ISO_3166-1 two letter codes which
will be consistent and unique.

PR #2368, opened 2022-01-12
2223 iso country codes

iamleeg:
For #2223, use ISO-3166-1 alpha-2 country codes everywhere.

~Draft for two reasons, putting review up to~ Please can we discuss reason 2 and get feedback on the parts that are done:

1. ~Still need to change ingestion to send country codes instead of country names.~
2. The migration for this change is one of those big, "read every case and change its content" migrations that we know went really badly the last time we tried. I don't yet know what to do about that.

The _workaround_ last time was that I wrote a script that ran in mongosh and used its batch update facility. That vastly sped up the migration. I think it's possible to do that within the migration script, though I couldn't work out how due to the interactions between the batches and the `async` functions here, so if anyone is able to fix that up I'd welcome it. Otherwise we can go down the same route of a script that is applied using mongosh: we need to work out where to commit that, how to document its purpose (someone has to run it manually in dev and prod)…we never did that for the confirmation date migration so there's currently a gap between the recorded migration path in the repo and reality.

Or maybe there's some other approach someone can recommend that will work well!

### DIFF
--- a/data-serving/data-service/api/openapi.yaml
+++ b/data-serving/data-service/api/openapi.yaml
@@ -75,12 +75,12 @@ paths:
                           value: 'this -butnotthis'
                           summary: Full text search with items that must or must not be present.
                       keywords:
-                          value: 'curator:foo@bar.com,baz@meh.com country:france'
+                          value: 'curator:foo@bar.com,baz@meh.com country:fr'
                           summary: >
                               values are OR'ed for the same keyword and all keywords are AND'ed.
                               Keyword values can be quoted for multi-words matches and concatenated
                               with a comma to union them. Bare words are compared for equality, e.g.
-                              country:Germany. The special character '*' matches any value, e.g. 
+                              country:de. The special character '*' matches any value, e.g. 
                               variant:* returns any case where the variant is not empty.
                               Supported keywords are: curator, gender, nationality, occupation,
                               country, outcome, caseid, uploadid, sourceid, sourceurl, verificationstatus,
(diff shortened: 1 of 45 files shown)


jim-sheldon (approved):

iamleeg:
The failure in ingestion CI is an unrelated bug: #2369.

abhidg:
Looks good! Do the country names from iso3166 match what we have in the db? I am thinking of a hypothetical scenario where we migrate up from United States -> US and migration down gives US -> United States of America.

iamleeg:
They may not—they're the official country names. In general as I note in the comment we can't guarantee that the migration is genuinely reversible; in addition to well-known aliases (e.g. United States for the United States of America) there may be multiple values (e.g. USA, US, America) or even bad data (e.g. Algiers). I've made a hopefully-acceptable compromise that the result of the "down" migration is that countries are stored by name.

In fact due to the performance issues with running the migration noted above, I think that I won't change the data in the migration at all, but will have a separate script that gets run manually in mongoose, as I ended up with for confirmation dates. That means there's no need to specify what "down" looks like at all at the moment, other than changing the validation rules.

abhidg (approved):

iamleeg:
Of course I don't know how to write the separate helper script either :(. In mongosh we don't have access to the project's `npm_modules` so I can't use `i18n-iso-countries` from the helper. Gah.

iamleeg:
oh! Apparently [I can use node_modules](https://docs.mongodb.com/mongodb-shell/write-scripts/require-external-modules/) \o/

iamleeg:
Re-requesting review due to changes to migration approach.

jim-sheldon (approved):

Issue #2369, opened 2022-01-12
Ages calculated incorrectly in Republic of Korea parser


iamleeg:
**Describe the bug**
The Korea parser calculates the ages of cases by doing `year now - birth year`. This is incorrect, because it means that the ages go up every year even for historically reported cases. It also introduces test failures that break CI, because the expected parsed case list in the test was created in 2021 and is now wrong.

PR #2371, opened 2022-01-13
parsing(republic_of_korea): Use confirmed date for age

abhidg:
Previously republic_of_korea was using the current year at the
time the parser was run, which led to the age changing in
successive years. This changes the age calculation to use
confirmed date.

Fixes: #2369


ingestion/functions/parsing/republic_of_korea/republic_of_korea.py
```diff
@@ -54,13 +54,14 @@ def convert_location(entry):
     return {"query": ", ".join(query_terms)}
 
 
-def convert_demographics(entry, current_year):
-    ''' Calculating age by subtracting birth year field from current_year global variable.'''
+def convert_demographics(entry):
+    ''' Calculating age by subtracting birth year field from confirmed date aear global variable.'''
```
iamleeg, line 58: typo: confirmed date year (and remove 'global')

iamleeg (approved):

PR #2378, opened 2022-01-24
Add freshness scripts, closes #2366

### LINKED ISSUE #2366: Calculate freshness
Write a script that for each country, will output the last date when the country data was ingested.

abhidg:


ingestion/monitoring/freshness.py
```diff
@@ -0,0 +1,80 @@
...
+def fetch_sources(api_key, instance):
+    logging.info(f"Fetching sources from {instance}")
+    res = requests.get(
+        f"{instance}/api/sources?limit=100",
+        headers={
+            "Content-Type": "application/json",
+            "X-API-Key": api_key,
+        },
+    )
+    return res.json() if res.status_code == 200 else None
```
jim-sheldon, line 24: If 4XX or 5XX log the response?

ingestion/monitoring/freshness.py
```diff
@@ -0,0 +1,80 @@
...
+
+def upload(data, bucket, s3_endpoint=None):
+    logging.info(f"Uploading data to {bucket}, with s3_endpoint={s3_endpoint}")
+    s3 = boto3.client("s3", endpoint_url=s3_endpoint)
+    s3.put_object(
+        ACL="public-read",
+        Body=json.dumps(data, indent=2, sort_keys=True),
+        Bucket=bucket,
+        Key="country/freshness.json",
+    )
```
jim-sheldon, line 61: put this in a try/except block?

ingestion/monitoring/freshness.py
iamleeg: Agreed

iamleeg: Also consider what to do if there are more than 100 sources, I know it looks like that won't ever be a problem with the covid-19 data sources but we should at least error out if we're not going to handle it, for when it eventually happens on this/another deployment.

abhidg: thanks, added paging support

ingestion/monitoring/freshness.py
```diff
@@ -0,0 +1,99 @@
...
+    h = logging.StreamHandler(sys.stdout)
+    rootLogger = logging.getLogger()
+    rootLogger.addHandler(h)
+    rootLogger.setLevel(logging.INFO)
+
+
+if __name__ == "__main__":
+    setup_logger()
+    if not (api_key := os.getenv("GDH_API_KEY")):
+        print("Set GDH_API_KEY to your Global.health API key")
```
jim-sheldon, line 93: `api_key` is an unused variable. `print` should be `logging` and `sys.exit(1)` can be `raise`.

iamleeg: `api_key` is used on line 98 :)

jim-sheldon: Ah my bad, did not realize walrus assignment extended beyond the `if` block.

jim-sheldon (approved):
Looks great!

PR #2399, opened 2022-01-25
Correct the country coding script to update location documents, not replace them

iamleeg:
This fixes a bug I found when running the migration in dev, that locations were ending up looking like `{country: "BE"}` without all the other fields like lat, long, query, resolution etc.

Because the script errored out when it detected the validation failures with the first batch, a maximum of 10k cases in dev are affected. I will just drop those, then run the fixed migration. I do not know why mongodb was willing to commit the changed versions despite the validation failures.


iamleeg:
> I will just drop those

Now done.

abhidg (approved):

PR #2406, opened 2022-01-28
Add moar logging

iamleeg:
Introduce middleware that will log every request/response, and any errors, in the data-service. It filters out potential PII.

### DIFF
--- a/data-serving/data-service/src/index.ts
+++ b/data-serving/data-service/src/index.ts
@@ -25,6 +25,8 @@ import dotenv from 'dotenv';
 import express from 'express';
 import mongoose from 'mongoose';
 import swaggerUi from 'swagger-ui-express';
+import winston from 'winston';
+import expressWinston from 'express-winston';
 import validateEnv from './util/validate-env';
 import { logger } from './util/logger';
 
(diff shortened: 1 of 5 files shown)


data-serving/data-service/src/index.ts
```diff
@@ -40,6 +42,20 @@ if (!deployment_envs.includes(env.SERVICE_ENV)) {
...
+// log all non-200 responses: this needs to come _before_ any middleware or routers
+app.use(expressWinston.logger({
+    transports: [
+        new winston.transports.Console(),
+    ],
+    format: winston.format.json(),
+    /* don't log user info. We don't get user cookies or passwords in this service, so it's just
+     * belt-and-braces to ensure we don't log the API key if it was forwarded from the curator service.
+     */
+    headerBlacklist: ['X-Api-Key'],
```
abhidg, line 55: this is case-insensitive, right?

iamleeg: [yes](https://github.com/bithavoc/express-winston/blob/main/index.js#L120) :)

abhidg (approved):

Issue #2410, opened 2022-01-29
batchValidate passes but batchUpsert fails validation at DB


abhidg:
**Describe the bug**
batchValidate passes but batchUpsert fails with BulkWriteError for SRAG data after adding vaccine fields

**To Reproduce**
Re-run SRAG parser.

Previous log at https://console.aws.amazon.com/cloudwatch/home?region=us-east-1#logsV2:log-groups/log-group/$252Faws$252Fbatch$252Fjob/log-events/brazil_srag-srag-ingestor-prod$252Fdefault$252Fd3457dba353f43f2a3db4f8123bd56c4

Tried re-inserting document in MongoDB 5.0 (which shows validation errors), and the vaccines field is correctly inserted.

iamleeg:
And I tried manually inserting a case from the AWS logs you linked into dev (4.4.12) and it failed validation. Here's what I'd try next: turn off validation (in dev or locally!), insert the documents, turn validation back on, [validate the collection](https://docs.mongodb.com/manual/reference/method/db.collection.validate/).

PR #2417, opened 2022-02-02
data-service/model: Fix mismatch in vaccine sideEffects schema

abhidg:
Vaccine side effects are an array in the Mongoose schema, but
of the form {values: [string], status: string} in the MongoDB
schema, which led to validation passing but raised an error
on insert to MongoDB.

Fixes: #2410


iamleeg (approved):
Looks good, maybe always set the status to "Symptomatic" for side effects.

iamleeg (approved):
🎆

PR #2418, opened 2022-02-02
parsing: Fix TypeError when travelHistory is None

abhidg:
When travelHistory is None, the first condition succeeds but
the second raises a TypeError as NoneType is not iterable.


ingestion/functions/common/parsing_lib.py
```diff
@@ -171,12 +171,12 @@ def prepare_cases(cases: Generator[Dict, None, None], upload_id: str, excluded_c
     """
     for case in cases:
         case["caseReference"]["uploadIds"] = [upload_id]
-        if "location" in case and "country" in case["location"]:
+        if case.get("location") and case["location"].get("country"):
```
jim-sheldon, line 174: Could do `if country := case.get("location", {}).get("country"):`

PR #2419, opened 2022-02-03
parsing(srag): Disallow both vaccine dates being the same

abhidg:


iamleeg (commented):
Is this a problem with the source data? And do we know that it's a mistake even if it does occur? (I can't imagine that people would be getting two doses on the same day, but I couldn't imagine many things that have happened)

abhidg:
Not sure about whether it is possible to have both vaccines be on the same day. At the very least, we have to exclude cases where both the date and the batch are the same, as the MongoDB schema requires vaccines array to be unique. I've started a discussion in Slack.

Mougk:
even if given on the same date it does not make sense from an immunisation persepctive. ignore these please.

PR #2426, opened 2022-02-03
iso3166 keyerror

abhidg:
- parsing: Make geocode_country() return ISO codes
- parsing: Add fixes for missing country to ISO codes


iamleeg (approved):
Thanks, Abhishek!

PR #2429, opened 2022-02-04
Map country codes to names in generating aggregate data

iamleeg:


jim-sheldon (approved):

data-serving/scripts/aggregate/aggregate/app.py
```diff
@@ -318,14 +321,14 @@ def generate_region_json(cases, s3_endpoint, bucket, date):
...
-                id = record["country"]
+                id = country_name(record["country"])
                 if id is None:
                     continue
                 search_term = "country"
             new_record = {
                 "_id": id,
                 "casecount": record["casecount"],
-                "country": record["country"],
+                "country": id,
```
abhidg, line 331: Add a country_code field so that we don't have to lookups elsewhere in case we need it?

PR #2500, opened 2022-02-08
1826 create QA environment

jim-sheldon:
Added QA cluster
Added references to QA environment
Removed aws/k8s folder


data-serving/data-service/src/index.ts
```diff
@@ -35,7 +35,7 @@ const app = express();
 dotenv.config();
 const env = validateEnv();
 
-const deployment_envs = ['prod', 'dev'];
+const deployment_envs = ['prod', 'dev', 'qa'];
```
iamleeg, line 38: proper nitpicking, but I'd put qa in between dev and prod just as a reminder of the deployment order.

ingestion/functions/aws.py
```diff
@@ -12,8 +12,9 @@
 
 AWS_REGION = "us-east-1"
 AWS_IMAGE = {
-    'dev': "612888738066.dkr.ecr.us-east-1.amazonaws.com/gdh-ingestor:latest",
-    'prod': "612888738066.dkr.ecr.us-east-1.amazonaws.com/gdh-ingestor:stable"
+    "dev": "612888738066.dkr.ecr.us-east-1.amazonaws.com/gdh-ingestor:latest",
+    "qa": "612888738066.dkr.ecr.us-east-1.amazonaws.com/gdh-ingestor:latest",
```
iamleeg, line 16: should this actually be on `stable`, so that qa is a prod-like environment? Or do we need an `rc` tag so that qa is what prod would become if we were happy with the behaviour in qa? To discuss, I think.

k8s/qa/curator.yaml
```diff
@@ -0,0 +1,115 @@
...
+          ports:
+            - containerPort: 3001
+          livenessProbe:
+            httpGet:
+              path: /health
+              port: 3001
+            initialDelaySeconds: 10
+            periodSeconds: 10
+          env:
+            # TODO: Remove this pending https://github.com/globaldothealth/list/issues/961.
```
iamleeg, line 34: sorry, but you drew the short straw: #961 has been closed for a while, this TODO can be DID. Unless we have other reasons for wanting this output (I think it gets in the way of grokking application logs, personally)

k8s/qa/data.yaml
```diff
@@ -0,0 +1,67 @@
...
+          ports:
+            - containerPort: 3000
+          livenessProbe:
+            httpGet:
+              path: /health
+              port: 3000
+            initialDelaySeconds: 10
+            periodSeconds: 10
+          env:
+            # TODO: Remove this pending https://github.com/globaldothealth/list/issues/961.
```
iamleeg, line 34: See above.

k8s/qa/curator.yaml
```diff
@@ -0,0 +1,115 @@
...
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        environment: qa
+        app: curator
+    spec:
+      containers:
+        - name: curator
+          image: ghcr.io/globaldothealth/list/curatorservice:main
```
iamleeg, line 23: again, questioning whether we want this to be dev-like or prod-like.

iamleeg (changes requested):
Good work, and nice job on claiming the round-number PR. Just marked for changes as I think there are some things here that we need to agree as a team regarding what this qa environment represents, but there's nothing wrong with the work.

ingestion/functions/aws.py
abhidg: +1 for qa -> stable

data-serving/data-service/src/index.ts
jim-sheldon: > Looks good @jim-sheldon! I think qa should generally be prod like, with deployments from the tip of the latest stable branch.
> 
> I'd rename prod to prod-dev or sth similar, as right now prod/ contains both prod and dev deployments

Good call, I will rename the folder.
There are a few branching and deployment strategies we could adopt; we should discuss this at greater length on a call or in a meeting.

iamleeg (approved):

PR #2506, opened 2022-02-10
2217 delete old ADI source data

iamleeg:
Draft because I still need to integrate this into the LocalStack e2e test runner (and into ops too), but asking for feedback on the script and the approach. Also, how _will_ this get run? Is there somewhere I can run this on a schedule that already has access to a service account for our AWS?

### DIFF
--- a/ingestion/functions/aws_access/EventBridgeClient.py
+++ b/ingestion/functions/aws_access/EventBridgeClient.py
@@ -1,0 +1,58 @@
+import boto3
+from uuid import uuid4
+
+
+"""
+Discover AWS EventBridge rules
+"""
+class EventBridgeClient:
+    def __init__(self, region: str):
+        self.aws_client = boto3.client("events", region)
+
+    def get_rule_descriptions(self):
+        return self.aws_client.list_rules().get("Rules")
+
+    def get_rule_targets(self, rule_name: str):
+        batch_targets = {}
+        targets = self.aws_client.list_targets_by_rule(Rule=rule_name)
+        for target in targets.get("Targets"):
+            target_id = target.get("Id")
+            if not target_id:
+                raise KeyError(f"Could not get target ID for {rule_name}")
+            batch_targets[target_id] = target.get("BatchParameters")
+        return batch_targets
+
+    def enable_schedule(self, rule_name: str):
+        self.aws_client.enable_rule(Rule=rule_name)
+
+    def disable_schedule(self, rule_name: str):
+        self.aws_client.disable_rule(Rule=rule_name)
+
+    def add_schedule(self, rule_name: str, description: str, is_enabled: bool):
+        state = "ENABLED" if is_enabled else "DISABLED"
+        self.aws_client.put_rule(
+                Name=rule_name,
+                ScheduleExpression=DEFAULT_SCHEDULE_EXPRESSION,
+                State=state,
+                Description=description
+            )
+
+    def remove_schedule(self, rule_name: str):
+        self.event_bridge_client.delete_rule(rule_name)
+
+    def add_target_to_rule(self, rule_name: str, target_name: str, job_name: str, job_queue: str, role: str):
+        target_id = str(uuid4())
+        target = [{
+            "Id": target_id,
+            "Arn": job_queue,
+            "RoleArn": role,
+            "BatchParameters": {
+                "JobDefinition": target_name,
+                "JobName": job_name
+            }
+        }]
+        self.aws_client.put_targets(Rule=rule_name, Targets=target)
+
+    def remove_targets_from_rule(self, rule_name: str):
+        target_ids = list(self.get_rule_targets(rule_name).keys())
+        self.aws_client.remove_targets(Rule=args.rule_name, Ids=target_ids)
--- a/ingestion/functions/aws_access/EventBridgeClient.py
+++ b/ingestion/functions/aws_access/EventBridgeClient.py
@@ -1,13 +1,17 @@
-import boto3
+import logging
 import re
 from uuid import uuid4
 from datetime import datetime, timedelta
+
+import boto3
 from monthdelta import monthdelta
 
 from . import globaldothealth_configuration as gdoth
 
 # Amazon weeks start on Sunday
 WEEKDAYS = ['SUN', 'MON', 'TUE', 'WED', 'THU', 'FRI', 'SAT']
+
+logger = logging.getLogger(__name__)
 
 def date_of_requested_weekday_in_month(weekday, week_of_month, year, month):
     # find the isoweekday of the first of the month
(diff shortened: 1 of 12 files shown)


ingestion/functions/aws_access/EventBridgeClient.py
```diff
@@ -0,0 +1,173 @@
...
+from datetime import datetime, timedelta
+from monthdelta import monthdelta
+
+from . import globaldothealth_configuration as gdoth
+
+# Amazon weeks start on Sunday
+WEEKDAYS = ['SUN', 'MON', 'TUE', 'WED', 'THU', 'FRI', 'SAT']
+
+# Our file retention policy is twice the schedule period plus 10 days
+GRACE_PERIOD_IN_DAYS = 10
```
abhidg, line 13: Better in globaldothealth_configuration?

abhidg:
GitHub Actions has a service account for AWS setup (that's how we deploy to ECR), so if this script takes <6h it can be that, otherwise a Batch job

ingestion/functions/aws_access/EventBridgeClient.py
```diff
@@ -1,5 +1,114 @@
...
+            assert schedule.startswith('cron') # I'm assuming that if it isn't a rate, it's a cron
+            components = re.match(r'cron\(([0-9]+) ([0-9]+) (\?|\*) (\*) ([0-6]#[1-4]|[A-Z]{3}|[0-6]|\?) (\*)\)', schedule)
+            assert components is not None # Assuming a restricted subset of https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-create-rule-schedule.html rules
+            # I also assume that you never ask for the fifth weekday of the month, because that's a very odd schedule
+            (minute, hour, day_of_month, month, day_of_week, year) = components.groups()
+            # calculate from midnight as we don't want odd seconds etc
+            today = datetime(now.year, now.month, now.day)
+            our_day = today.isoweekday() % 7 # ISO runs 1-7 Mon-Sun, we want 0-6 Sun-Sat
+            # Because of my assumptions above in the regex, I only need the day of the week here
+            # When we are using Python 3.10+, this should be a match
```
abhidg, line 74: Is this using a feature only in 3.10?

ingestion/functions/test_clean_old_ingestion_source_files.py
```diff
@@ -0,0 +1,77 @@
+import boto3
+import os
+import pytest
+import unittest
+from datetime import datetime, timedelta
+from botocore.exceptions import ClientError
```
jim-sheldon, line 6: Nitpicking: third-party imports should live in their own group between standard library imports and local application imports

ingestion/functions/aws_access/EventBridgeClient.py
iamleeg: The intent of the comment is that I didn't use a 3.10-only feature, but it would be better if I could (all those if/elif/else can be a match/case/case which is a nicer way of expressing this).

iamleeg:
> GitHub Actions has a service account for AWS setup (that's how we deploy to ECR), so if this script takes <6h it can be that, otherwise a Batch job

Nice, makes sense to be a daily GitHub action I think.

ingestion/functions/clean_old_ingestion_source_files.py
```diff
@@ -0,0 +1,40 @@
...
+from aws_access.S3Client import S3Client
+import aws_access.globaldothealth_configuration as gdoth
+
+def get_schedule_rules(endpoint_url=None):
+    """Find a list of the schedule rules related to ingestion in G.h"""
+    event_bridge = EventBridgeClient(gdoth.AWS_REGION, endpoint_url)
+    return [ScheduleRule(r) for r in event_bridge.get_rule_descriptions()]
+
+def main(endpoint_url=None):
+    print("Let's do this!")
```
jim-sheldon, line 14: Could this `print` be a `log`?

ingestion/functions/aws_access/EventBridgeClient.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def get_rule_descriptions(self):
+        return self.aws_client.list_rules().get("Rules")
+
+    def get_rule_targets(self, rule_name: str):
+        batch_targets = {}
+        targets = self.aws_client.list_targets_by_rule(Rule=rule_name)
+        for target in targets.get("Targets"):
+            target_id = target.get("Id")
+            if not target_id:
```
jim-sheldon, line 135: Nit: could use walrus

ingestion/functions/test_clean_old_ingestion_source_files.py
```diff
@@ -0,0 +1,77 @@
...
+localstack = 'http://localhost:4566'
+rate_event_id = '5f7796ece78c6866f6f676e0'
+file_content = "blah blah blah" # this is unimportant
+file_time = '1130' # also unimportant
+file_name = 'slartibartfast.json' # "my name is not important"
+
+def s3_object_key_helper(source_id, date, file_name):
+    return f"{source_id}/{date.year}/{date.month}/{date.day}/{file_time}/{file_name}"
+
+
```
jim-sheldon, line 21: Should all methods only run in the dockerized environment?

ingestion/functions/aws_access/globaldothealth_configuration.py
```diff
@@ -0,0 +1,3 @@
+AWS_REGION = "us-east-1"
+INGESTION_SOURCES_BUCKET = "epid-sources-raw"
+DEFAULT_SCHEDULE_EXPRESSION = "rate(1 day)"
```
jim-sheldon, line 3: Might be better to make these values come from environment variables, and keep these values as defaults.

abhidg: +1, config can then be in the GitHub Actions yaml

jim-sheldon:
> > GitHub Actions has a service account for AWS setup (that's how we deploy to ECR), so if this script takes <6h it can be that, otherwise a Batch job
> 
> Nice, makes sense to be a daily GitHub action I think.

I would prefer we keep it in AWS, and thus terraformed, as it would keep infrastructure consolidated and might be part of partner requirements.

ingestion/functions/aws_access/EventBridgeClient.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def date_of_requested_weekday_in_month(self, weekday, week_of_month, year, month):
+        return date_of_requested_weekday_in_month(weekday, week_of_month, year, month)
+
+    def oldest_file_age(self, now):
+        """
+        What is the age in days of the oldest source file for an ingestion rule that we can keep?
+        now: reference time from which to compute file age
+        """
+        assert self.is_ingestion_rule()
```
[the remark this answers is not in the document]
iamleeg: no good point, I designed this by contract so the assertions are there to express what I think should be happening. I'll remove them and replace with logging.

ingestion/functions/test_clean_old_ingestion_source_files.py
iamleeg: As this is called by the test and the test only runs in the docker environment, I think that's already true? Nothing is going to run this function otherwise.

ingestion/functions/aws_access/globaldothealth_configuration.py
iamleeg: OK, will read from env then. Thanks!

PR #2514, opened 2022-02-10
Prepare for EU transition

abhidg:
Changed hardcoded references to use environment variables with
the default set to eu-central-1

### DIFF
--- a/data-serving/scripts/export-data/common.sh
+++ b/data-serving/scripts/export-data/common.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-export ECR=612888738066.dkr.ecr.us-east-1.amazonaws.com
+export ECR=612888738066.dkr.ecr.eu-central-1.amazonaws.com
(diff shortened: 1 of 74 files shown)


data-serving/scripts/export-data/common.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-export ECR=612888738066.dkr.ecr.us-east-1.amazonaws.com
+export ECR=612888738066.dkr.ecr.eu-central-1.amazonaws.com
```
jim-sheldon, line 2: I would expect the number to be different as well. We will know when we create the new repo.

abhidg: I think the number is our AWS account which shouldn't change - but good to parametrize that as well if possible

jim-sheldon:
Good stuff. In general I would expect ARNs to change. We will know more when we create new resources in the eu region.

jim-sheldon (approved):

Issue #2217, opened 2021-11-01
Delete old ADI source data


iamleeg:
**Describe the bug**
Whenever we fetch new data from an upstream source for automated data ingestion, we copy it to S3. It never gets deleted. This adds to our AWS bill (not much, but it is money we don't have to pay) and it means we keep data that is potentially not anonymised to our standards around, which we don't need to do.

**To Reproduce**
Nothing to do, this automatically reproduces.

**Expected behavior**
Source data should be kept no longer than is necessary for diagnosing ADI problems. I'd suggest twice the period between updates plus a bit, so we can retroactively discover a problem and compare the most recent ingested data with the one before that.

iamleeg:
I have put backstop rules in place in S3 on the `epid-sources-raw` and `archive-epid-sources-raw` buckets:

 - source data is marked as noncurrent after 90 days
 - noncurrent files are deleted after 14 days

This is way longer than any of the regular schedules, but means we start clearing up the source data and handles edge cases like where we don't have an automation rule any more so don't have a normal expiration time for the source data.

PR #2521, opened 2022-02-14
Dockerfile for ADI source cleanup script

iamleeg:
Dockerfile, and GitHub action to push to ECR. Needed before we can use it in Batch.

BTW, what's the plan for deploying in Batch? Do I deploy from the console and replicate the setup in the terraform repo, or define in terraform then apply to the prod infrastructure?

#2217

### DIFF
--- a/.github/workflows/ingestion-cleanup-deploy.yml
+++ b/.github/workflows/ingestion-cleanup-deploy.yml
@@ -1,0 +1,60 @@
+name: Ingestion cleanup deploy
+
+on:
+  push:
+    branches: [main, '*-stable']
+    paths:
+    - '.github/workflows/ingestion-cleanup-deploy.yml'
+    - 'ingestion/functions/**'
+    - '!ingestion/functions/README.md'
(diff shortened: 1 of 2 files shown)


.github/workflows/ingestion-cleanup-deploy.yml
```diff
@@ -0,0 +1,60 @@
+name: Ingestion cleanup deploy
+
+on:
+  push:
+    branches: [main, '*-stable']
+    paths:
+    - '.github/workflows/ingestion-cleanup-deploy.yml'
+    - 'ingestion/functions/**'
+    - '!ingestion/functions/README.md'
```
abhidg, line 9: I'd add `!ingestion/functions/parsing/**` and `!ingestion/functions/retrieval/**` to remove unnecessary builds

abhidg:
Looks good, needs a rebase onto main, and I'd add excludes in workflows so that changes to the cleanup script don't trigger builds in ingestion, and vice-versa.

abhidg (approved):

Issue #2319, opened 2021-12-09
Replace Python print statements with logging to StreamHandler


jim-sheldon:
Calling `print` in Python does not automatically write its output to stdout; when a script runs in Docker and hangs on a line, it can result in a timeout with incomplete (even no) output written to Cloudwatch logs.
We should use `logging` instead of `print` statements, and we need to use a `StreamHandler` to ensure each log statement gets flushed from the buffer and made visible in Cloudwatch.

https://docs.python.org/3/library/logging.handlers.html#streamhandler

iamleeg:
Looks like there's still stuff to do in ingestion, I'll take that on.

PR #2525, opened 2022-02-15
Replace print with logging where it makes sense #2319

iamleeg:


ingestion/functions/parsing/USA/USA.py
```diff
@@ -18,6 +19,7 @@
             os.pardir,os.pardir, 'common'))
     import parsing_lib
```
iamleeg, line 20: I think this stanza (found in all of the parsers) is redundant Lambda-era noise. We create the image ourselves from a Dockerfile now, and we put `common` in the right place, so `common.parsing_lib` should be discoverable both in tests and in production. Should we clean this up, or am I wrong?

abhidg: From the comments, this is about getting pytest to run, but that should be possible by using python -m pytest instead of just pytest, so yes, this should be fine to delete :tada:

abhidg (approved):

PR #2533, opened 2022-02-18
ingestion: Revert removal of imports workaround, fixes #2532

### LINKED ISSUE #2532: Ingestion failing with module not found
Ingestion failing with module not found

```


--
  File "./retrieval/retrieval.py", line 47, in <module>
    import common_lib
ModuleNotFoundError: No module named 'common_lib'
During handling of the above exception, another exception occurred:
  File "./retrieval/retrieval.py", line 53, in <module>
    import common_lib
  File "/app/retrieval/../common/common_lib.py", line 16, in <module>
    import common.ingestion_logging as logging
ModuleNotFoundError: No module named 'common'
```

abhidg:
- Revert "Test removal of import workaround. Tests still passed locally... #2319"
- Revert "Remove the imports hack in the rest of the parsers #2319"
- ingestion(logging): addHandler() requires argument
- ingestion: Change path to make pytest pass
- ingestion: Set logging level to INFO

### DIFF
--- a/ingestion/functions/common/common_lib.py
+++ b/ingestion/functions/common/common_lib.py
@@ -9,18 +9,23 @@ object.
 
 import os
 import re
+import sys
 import json
 import tempfile
 import requests
 import functools
-import common.ingestion_logging as logging
+from enum import Enum
+from pathlib import Path
 
 import google
 import google.auth.transport.requests
-
-from enum import Enum
-from pathlib import Path
 from google.oauth2 import service_account
+
+try:
+    import ingestion_logging as logging
+except Exception:
+    sys.path.append(Path(__file__).parent)
+    import common.ingestion_logging as logging
 
 E2E_MOCK_SOURCE_URL = os.environ.get("MOCK_SOURCE_DATA_ADDRESS", "")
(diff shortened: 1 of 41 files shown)


ingestion/functions/common/common_lib.py
```diff
@@ -36,6 +41,7 @@
 MIN_SOURCE_ID_LENGTH, MAX_SOURCE_ID_LENGTH = 24, 24
 
 logger = logging.getLogger(__name__)
+logger.setLevel("INFO")
```
jim-sheldon, line 44: I think that rather than overriding the log level it might be better to set it in `ingestion_logging` based on environment--prod and qa get `INFO`, lower environments get `DEBUG`.

abhidg: Found a bug in setting loglevels in ingestion_logging (they are not followed, and loglevel is always rogging.WARNING), will create an issue for that, and include this task in that issue.

iamleeg:
Thanks for reverting these changes @abhidg!

The ideal solution to finding these packages reliably is to use `setuptools` and install them with `pip`/`poetry` in the target environment. I wish I'd done that now :-(

PR #2536, opened 2022-02-21
2289 curator UI version

iamleeg:
There is a `/version` endpoint on the API that reports the version, which the front-end uses to report the version on the landing page. The version can also be used in scripts for compatibility testing, or in getting problem reports from users on different deployments. This version is set in the following ways:

1. it is set to the value of `$CURATOR_VERSION`.
2. if you run in Docker then this value can come from the env or from a build arg.
3. in the `dev/run_stack.sh` scripts, it's set to the commit SHA
4. in the GitHub action, it's set to the computed version which is the version number in prod 🎉 or the branch name/PR number for other builds. Not super useful in dev but at least reminds you that you _are_ looking at dev (version is `main`) rather than any other instance.


abhidg (approved):
:tada:

Issue #2216, opened 2021-11-01
Move the prune-uploads script to AWS Batch


iamleeg:
It currently takes 6h to run the prune script for the US data set which is right on the runtime limit for GitHub actions.

While here, we also need better sequencing via hooks or queueing for ingestion, pruning, exporting, and aggregation to ensure that downloads and reports are consistent with a stable view of the data.

abhidg:
Bumping to P1 as non-US ingestions are timing out as well

PR #2538, opened 2022-02-21
prune: move to Batch #2216

abhidg:


data-serving/scripts/prune-uploads/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM python:3.10-slim
```
iamleeg, line 1: 3.10 😍

iamleeg (changes requested):
Code changes look good, thanks Abhishek, just concerned that we'd be running `main` code in `prod` and nothing in `dev`.

iamleeg (approved):

PR #2540, opened 2022-02-22
Default sorting by confirmation date #2098

maciekz1996:
Setup default sorting by `confirmationDate` in Line list table and added needed indexes to MongoDB


abhidg (approved):
Thanks!

PR #2546, opened 2022-02-24
completeness: add exponential backoff for downloading files

abhidg:
Also do not return null for countries in completeness.json
when download failed.

### DIFF
--- a/ingestion/monitoring/completeness.py
+++ b/ingestion/monitoring/completeness.py
@@ -1,6 +1,7 @@
 import os
 import sys
 import json
+import time
 import tempfile
 import logging
 import contextlib
@@ -8,6 +9,9 @@ from pathlib import Path
 
 import boto3
 import pandas as pd
+import botocore
+
+MAX_TRIES = 10
 
 
 def data_files(bucket, prefix="csv", suffix=".csv.gz", endpoint_url=None):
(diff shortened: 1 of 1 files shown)


ingestion/monitoring/completeness.py
```diff
@@ -53,14 +56,23 @@ def completeness_s3(s3_object, endpoint_url=None):
...
-            bucket.download_file(s3_object.key, tmp.name)
-        except Exception as e:
-            logging.error(f"Failed to download with exception {e}")
+        for tries in range(MAX_TRIES):
+            try:
+                bucket.download_file(s3_object.key, tmp.name)
+                success = True
+                break
+            except Exception:
+                logging.info(f"Failed to download {s3_object}, retrying in {2**tries} seconds ...")
```
iamleeg, line 68: could probably look at the exception to decide whether it's temporary (worth retrying) or not, but definitely should still log it.

iamleeg (changes requested):

ingestion/monitoring/completeness.py
jim-sheldon: What was the exception that led to adding `sleep`?

jim-sheldon:
Might it be better to get this data from the database? It seems a bit more straightforward that way.

abhidg:
> Better to get this data from the database? Seems more direct and simple that way.

DB would get from S3 anyway for the country export, and doing a .find() would start a collection scan, which I'd like to avoid.

jim-sheldon:
> > Better to get this data from the database? Seems more direct and simple that way.
> 
> DB would get from S3 anyway for the country export, and doing a .find() would start a collection scan, which I'd like to avoid.

But S3 gets its data from the database. This seems like a good place to use scheduled triggers https://docs.mongodb.com/realm/triggers/scheduled-triggers/ ... or maybe making the completeness counting part of the data export.

iamleeg (approved):

PR #2555, opened 2022-02-28
retrieval: remove unnecessary decode-encode step

abhidg:
This should fix MemoryError on attempting encoding


iamleeg (changes requested):
Do we need to read the file in at all? We immediately wrap the in-memory content with a file-like object. I think this could be replaced by opening the file-like object and returning that.

iamleeg (approved):
This can also be done in the other code paths where we construct a `BytesIO` :). In fact in some places we're just copying the file via calls to `read` and `write`, could get the OS to do that for us.

PR #2561, opened 2022-02-28
1564 error logs to slack

iamleeg:
- [x] script doesn't depend on reaching an event from lambda
- [x] copes with being rate limited by Slack
- [x] Documentation
- [x] Dockerfile
- [x] Deploy to ECR
- [ ] Define job in Batch
- [ ] Schedule from EventBridge

### DIFF
--- a/ingestion/monitoring/Dockerfile
+++ b/ingestion/monitoring/Dockerfile
@@ -34,7 +34,7 @@ RUN apt-get update \
         build-essential
 
 # install poetry - respects $POETRY_VERSION & $POETRY_HOME
-RUN curl -sSL https://install.python-poetry.org/ | python3 -
+RUN curl -sSL https://install.python-poetry.org/ | python3 - --version 1.1.13
 
 # copy project requirement files here to ensure they will be cached.
 WORKDIR $PYSETUP_PATH
--- a/ingestion/monitoring/Dockerfile
+++ b/ingestion/monitoring/Dockerfile
@@ -66,6 +66,7 @@ COPY errorLogsToSlack.py poetry.lock pyproject.toml ./
 # quicker install as runtime deps are already installed
 RUN poetry install
 
-# notice I haven't set the environment variables here, these
-# should be configured at invocation time
+# notice I haven't set the environment variables or args here.
+# the slack webhook should be configured in the job definition,
+# and the args should be configured at submission time
 CMD ["poetry", "run", "python3", "./errorLogsToSlack.py"]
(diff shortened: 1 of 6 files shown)


jim-sheldon (approved):

abhidg:
I'm not familiar with boto3 logs module, but will this only return the latest logs? We wouldn't want earlier logs coming in. Also would be good to hook it up to EventBridge event on Batch job failure -- may need some changes to handle the input event which is usually passed as a JSON payload.

iamleeg:
> I'm not familiar with boto3 logs module, but will this only return the latest logs? We wouldn't want earlier logs coming in. 

Do you mean to say that there isn't a separate log stream for each run of a parser? This makes it more difficult, as I need to know what the start and end dates are of the parser run I care about to filter the log stream. Do you know how I can get those?

> Also would be good to hook it up to EventBridge event on Batch job failure -- may need some changes to handle the input event which is usually passed as a JSON payload.

Ah OK, I was hoping I'd be able to define the environment variables in the event but I'll look into doing that.

abhidg:
> Do you mean to say that there isn't a separate log stream for each run of a parser? This makes it more difficult, as I need to know what the start and end dates are of the parser run I care about to filter the log stream. Do you know how I can get those?

Re-reading that bit, it is fine - as there is a separate log stream with the same prefix for each parser (x-x-ingestor-prod/*), which can be obtained from the event.

ingestion/monitoring/Dockerfile
```diff
@@ -0,0 +1,71 @@
+# `python-base` sets up all our shared environment variables
+FROM python:3.9-slim as python-base
```
abhidg, line 2: 3.10?

iamleeg: I had problems installing poetry in the image when I based it on python:3.10-slim so I punted it. I will check again, it may have been an unrelated issue.

iamleeg: Poetry [does not yet support Python 3.10](https://github.com/br3ndonland/inboard/pull/36)

abhidg: This appears to be the new way of installing poetry https://github.com/python-poetry/install.python-poetry.org -- I'm wary of the direct call to install.python-poetry.org, but should be possible to pin to a git tag

iamleeg: Good catch. We were calling get-poetry from master too so we probably need to change this throughout the repo.

PR #2566, opened 2022-03-02
Add data acknowledgment table

maciej-zarzeczny:
**Changes**
- Created Data acknowledgment page that is accessible without logging in
- Added public API endpoint for fetching sources


abhidg (approved):
Looks great, thanks @maciekz1996!

Issue #1564, opened 2021-02-16
Rework error logs to slack function


calremmel:
The error log reporting function that relays ingestion function errors to Slack is currently broken. Presently, adding parser logs as triggers is a time-consuming manual process, and hard to maintain. Additionally, it only covers a few different cases of errors.

Proposed Solution

- [ ] Update SAM template to automatically add triggers
- [ ] Filter logs less stringently - send more to the channel
- [ ] Store separate cache for logs to allow more granular reporting?

calremmel:
This should now wait on AWS Batch migration

abhidg:
Raising to P1 to get more visibility on ingestion failures. Currently there's an error log posted to Slack via a e-mail hook but that is a JSON dump and not very readable

PR #2574, opened 2022-03-08
Decomplect the slack logging by putting it directly into the ingestion job. #1564

iamleeg:
I was facing even more futzing with AWS to get the Slack script working (I needed to create an EventBridge filter to look for failed ingestion job events, and then write a script to interpret the event and extract the log stream to call `errorLogsToSlack.py` when I realised this was all far too complex. I can just make the ingestion jobs themselves log to Slack.


abhidg:
Wouldn't this log everything, even if the job succeeds? Would be better to only log errors and warnings

iamleeg:
It will log anything at log level WARNING or higher, so in principle anything that goes wrong.

abhidg (approved):

PR #2597, opened 2022-03-17
2559 canonical country list

iamleeg:
For _reasons_, I can't make a `distinct` query that uses an index with the mongoose driver. So switch to using mongo directly. This is a draft PR to track CI failures during this migration.


iamleeg:
This has gone on too long without integration and there are a lot of test failures and a lot of work to merge. I am going to abandon this PR and pick smaller amounts of changes onto separate branches to integrate.

PR #2601, opened 2022-03-22
Added db migrations

maciekz1996:
Added two migrations to replace `byConfirmationDateIfListed` index for the one using collation

### DIFF
--- a/data-serving/scripts/setup-db/migrations/20220322114358-drop-old-confirmation-date-index.js
+++ b/data-serving/scripts/setup-db/migrations/20220322114358-drop-old-confirmation-date-index.js
@@ -1,0 +1,15 @@
+module.exports = {
+  async up(db, client) {
+    await db.command({
+      dropIndexes: 'cases',
+      index: 'byConfirmationDateIfListed',
+    });
+  },
+
+  async down(db, client) {
+    await db.command({
+      createIndexes: 'cases',
+      indexes: indexes,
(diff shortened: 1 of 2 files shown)


data-serving/scripts/setup-db/migrations/20220322114358-drop-old-confirmation-date-index.js
```diff
@@ -0,0 +1,15 @@
...
+    await db.command({
+      dropIndexes: 'cases',
+      index: 'byConfirmationDateIfListed',
+    });
+  },
+
+  async down(db, client) {
+    await db.command({
+      createIndexes: 'cases',
+      indexes: indexes,
```
iamleeg, line 12: `indexes` isn't defined here, I think you need to include the value from the previous migration that defined it.

iamleeg (changes requested):

data-serving/scripts/setup-db/migrations/20220322114358-drop-old-confirmation-date-index.js
maciekz1996: Thanks for catching that 👍

iamleeg (approved):

abhidg:
@maciekz1996 Both the indices (the one being dropped and added) are the same. Was one supposed to be without collation?

Issue #2590, opened 2022-03-16
Remove the sidebar for usual user in Data Panel


marlenachroscicka:
Move all links from the sidebar to the footer.

PR #2603, opened 2022-03-23
#2590 - remove sidebar for normal users

maciekz1996:
**Changes**
- Sidebar is not visible for users that are not curators/admins
- Created `Footer` component always visible at the bottom of the page
- Cleaned up `App` component by moving `Sidebar` to it's own component


verification/curator-service/ui/src/components/Sidebar/index.tsx
```diff
@@ -0,0 +1,180 @@
...
+            anchor="left"
+            open={drawerOpen}
+            data-testid="sidebar"
+            classes={{
+                paper: classes.drawerPaper,
+            }}
+        >
+            <div className={classes.drawerContents}>
+                <div className={classes.drawerHeader}></div>
+                <Typography className={classes.covidTitle}>COVID-19</Typography>
```
iamleeg, line 112: I think this needs to be a parameter for "disease title = COVID-19", not hard-coded. One of the goals for G.h is we should be able to quickly start up for a new disease.

iamleeg (changes requested):
Looks great, just want to parameterise the disease name somehow.

iamleeg (approved):
Thanks a lot @maciekz1996!

Issue #2600, opened 2022-03-21
Add version number to applications


maciekz1996:
Ad version number to:

- [x] Map
- [ ] Wordpress
- [x] Curator portal after logging in

PR #2644, opened 2022-03-28
#2600 - version number

maciekz1996:
**Changes**
- Added version number in `Profile menu` when a user is logged in
- Version number links to Github's release notes
- Added release notes url as env variable
- Added `cors` library to the `api` in order to enable version fetching from the map url 

⚠️ In order for those changes to wok properly `REACT_APP_RELEASE_NOTES_URL: 
https://github.com/globaldothealth/list/releases/tag/1.10.1` should be added to `.env`


verification/curator-service/ui/src/components/App/index.tsx
```diff
@@ -286,6 +290,10 @@ function ProfileMenu(props: { user: User }): JSX.Element {
 
     const classes = menuStyles();
 
+    const releaseNotesUrl =
+        process.env.REACT_APP_RELEASE_NOTES_URL ??
+        `https://github.com/globaldothealth/list/releases/tag/${props.version}`;
```
iamleeg, line 295: I think actually the second version here, with the format string, is better than using the environment. `process.env` is a node feature and even if you can use it in the browser, that's the environment _of the browser process_ which we don't control. So either get the release notes URL from the backend as we currently do with the version, or build it knowing the version as you currently do: and I prefer the second approach.

verification/curator-service/ui/src/components/Sidebar/index.tsx
```diff
@@ -52,7 +52,7 @@ const Sidebar = ({ drawerOpen }: SidebarProps): JSX.Element => {
         [user],
     );
 
-    const diseaseName = process.env.REACT_APP_DISEASE_NAME ?? 'COVID-1922222';
+    const diseaseName = process.env.REACT_APP_DISEASE_NAME ?? 'COVID-19';
```
iamleeg, line 55: actually my comment above about `process.env` makes me think this isn't working well either :(

verification/curator-service/ui/src/components/landing-page/LandingPage.tsx
```diff
@@ -114,11 +113,23 @@ const MoreInformationLinks = ({
     version,
     env,
 }: StyleProps & { version: string; env: string }) => {
+    const releaseNotesUrl =
```
iamleeg, line 116: this can be extracted into a utility function along with the duplicate in `App/index.tsx`

iamleeg (changes requested):

iamleeg (approved):

Issue #2559, opened 2022-02-28
Supply canonical list of ISO 3166-1 country codes used


iamleeg:
> [8:06 AM] Lol. Well my next suggestion was going to be to supply a list of available country codes :slightly_smiling_face:

Currently we have a lot of different sources: a JS library, a Python library, and some workarounds for country names that Mapbox uses that neither of those recognise. Would be good to build a list of the country codes we use with maps to the name. Best would be to do this in Python and R as part of our API libraries.

abhidg:
The current documentation links to the Wikipedia entry on ISO 3166 codes which has a comprehensive list.

joe-brilliant:
I think we can closer this @iamleeg @abhidg ?

iamleeg:
No, the wikipedia list is not sufficient.

> Would be good to build a list of the country codes we use with maps to the name.

This won't include all of the weird workarounds we have for Mapbox (e.g. it uses "United States" for country US, which most libraries don't use because it's ambiguous with "United States Minor Outlying Islands" when you're searching), and the names otherwise might not be the same as the ones we use. We actually have to prepare a list of the country codes we use and map them to the names of the countries we use.

joe-brilliant:
Duly noted; thanks for the clarification + explanation @iamleeg !

joe-brilliant:
@abhidg please close if this has been deployed

PR #2651, opened 2022-04-01
Incremental Mongoose destruction #2559

iamleeg:


iamleeg:
Firefox Cypress tests seem flaky: I get different failures locally and in different runs on CI.

iamleeg:
Upgrading to latest Cypress causes different failures in `cypress-run-ff`, at least locally :(.

verification/curator-service/api/src/controllers/auth.ts
```diff
@@ -442,20 +442,22 @@ export class AuthController {
                     }
 
                     // Check if there is a token in DB already for this user
-                    const token = await Token.findOne({ userId: user._id });
+                    const query = { userId: user._id };
+                    const token = await tokens().findOne(query);
                     if (token) {
-                        await token.deleteOne();
+                        await token().deleteOne(query);
```
abhidg, line 448: tokens()

iamleeg:
What I'm seeing is that random tests fail with "expected to find content [whatever] but never did", when the browser screenshot shows the correct UI but not updated to reflect the actions that have been taken in the test (e.g. logged in on the line list, but not showing the filter UI when the test has applied a filter and wants to see the results). So seems like a timing problem but I don't see what has changed to cause it, unless there was recently a new FF version.

jim-sheldon (approved):

PR #2652, opened 2022-04-01
#2637 UI: Disallow future dates in filters

maciekz1996:
**Changes**
- Fixed bug where user could input more than 4 digits as a year in `FiltersDialog`


abhidg (changes requested):
Can the max year be the current year?

maciekz1996:
@

> Can the max year be the current year?

Yeah it actually makes more sense, thanks

PR #2655, opened 2022-04-01
Using updated DNS for dev map

jim-sheldon:


abhidg (commented):
dev-map is back, now in eu-central-1.

PR #2661, opened 2022-04-06
2559 incremental mongoose destruction

iamleeg:


iamleeg:
This deletes mongoose entirely from the model layer, it's only used for managing the database connection. Once the mongo node.js driver is handling that mongoose can be entirely removed, the driver upgraded, and I can write the query I wanted to write a few weeks back :D.

Looks like some of the Cypress tests have been failing so I'll investigate those before merge.

abhidg (approved):

PR #2668, opened 2022-04-13
2559 incremental mongoose destruction

iamleeg:
This is it!


jim-sheldon (approved):

abhidg (approved):
🎉

iamleeg:
Woah this is a mess. The unit tests were running cleanly when I tried locally but that seems to have been based on some stale build artefacts, as I have switched branch and back and now see the same problems that this PR's CI workflow reported.

The underlying cause of this problem seems to be that [the latest version of the nodejs mongodb driver doesn't support typescript 3.9](https://jira.mongodb.org/browse/NODE-4092). I had a go at updating to typescript 4.6 and that went predictably badly. I will see if there's an intermediate version of the mongodb driver to use.

iamleeg:
Hmm, why aren't GitHub workflows running? [status is green](https://www.githubstatus.com)

iamleeg:
Had to make quite a few changes to get this over the line so asking for review again to make sure you're comfortable, it's mostly typescript appeasement but I also drastically changed some versions (including node)

PR #2679, opened 2022-04-25
2670 bucket ages

iamleeg:
These are the changes to schema and data-service to support bucketed age ranges and to use them whenever a case is added or returned via the API, but not to remove the existing age range data or change data exports/UI to work with the ranges.

See documentation on the migration and data-service README for rationale behind this design.

I fully expect some e2e breakages so will wait not merge until the cypress tests are passing.


iamleeg:
Oh would you look at that, the tests work!

PR #2680, opened 2022-04-25
Remove notes field, fixes #2673

### LINKED ISSUE #2673: Remove notes section for all countries, make available upon request
The automatically-ingested notes contain free text that the upstream data providers added to the cases and these can be pretty specific and identifying, as the HDX partners have already discovered. Don't make notes part of the general data set, create a new 'note-viewer' role that has exclusive access and let admins give out (and revoke) that role as required.

abhidg:
This removes the notes field in the UI, API and exports. It doesn't actually delete the notes field in the database yet, which we can do at a later point, or create a separate access control to only give access to notes.


data-serving/data-service/src/controllers/case.ts
```diff
@@ -78,6 +78,7 @@ const dtoFromCase = async (storedCase: LeanDocument<CaseDocument>) => {
         delete (dto as unknown as { demographics: { ageBuckets?: [ObjectId] }}).demographics.ageBuckets;
     }
     delete dto.restrictedNotes;
+    delete dto.notes;
```
iamleeg, line 81: Maybe this is a "not yet" task but we could migrate `restrictedNotes` back over to `notes`, as _all_ notes are now restricted. Then we would be able to build a single approach to exposing notes later.

data-serving/data-service/src/model/fields.json
```diff
@@ -1 +1 @@
...
....administrativeAreaLevel3","location.country","location.geoResolution","location.geometry.latitude","location.geometry.longitude","location.name","location.place","location.query","pathogens","preexistingConditions.hasPreexistingConditions","preexistingConditions.values","revisionMetadata.creationMetadata.curator","revisionMetadata.creationMetadata.date","revisionMetadata.creationMetadata.notes","revisionMetadata.editMetadata.curator","revisionMetadata.editMetadata.date","revisionMetadata.editMetadata.notes","revisionMetadata.revisionNumber", "SGTF", "symptoms.status","symptoms.values","transmission.linkedCaseIds","transmission.places","transmission.routes","travelHistory.travel.dateRange.end","travelHistory.travel.dateRange.start","travelHistory.travel.location.name","travelHistory.travel.methods","travelHistory.travel.purpose","travelHistory.traveledPrior30Days","vaccines.0.name","vaccines.0.batch","vaccines.0.date","vaccines.0.sideEffects","vaccines.1.name","vaccines.1.batch","vaccines.1.date","vaccines.1.sideEffects","vaccines.2.name","vaccines.2.batch","vaccines.2.date","vaccines.2.sideEffects","vaccines.3.name","vaccines.3.batch","vaccines.3.date","vaccines.3.sideEffects",""]
```
iamleeg, line 1: this would be easier to diff if we put each element on its own line :)

data-serving/data-service/test/controllers/case.test.ts
```diff
@@ -200,41 +207,21 @@ describe('GET', () => {
                 .expect(200);
             expect(res.body.cases).toHaveLength(1);
         });
-        it('should query results', async () => {
```
iamleeg, line 203: Is there any text that the query should work over instead of notes, or is there any free-text search facility at all in the platform now? If not we can also drop the text index, which will speed up ingestion and save 13.5G in the prod cluster (the main collection is only 10.1G!)

iamleeg (approved):
Approved with some optional bonus tasks :)

data-serving/data-service/test/controllers/case.test.ts
abhidg: Agreed, in #2687

PR #2686, opened 2022-04-28
ci: Pin Ubuntu release to 20.04 LTS

abhidg:
Now that 22.04 LTS has been released, at some point ubuntu-latest
will point to ubuntu-22.04 which can cause unplanned breakage,
pinning CI to 20.04 until we have a migration plan.

Additionally, drop main-eu builds as EU migration is complete.


jim-sheldon:
Good stuff. We may even want to move to Alpine and/or scratch.

PR #2689, opened 2022-04-29
ci: fix syntax error in workflow_dispatch

abhidg:
workflow_dispatch does not take paths or branches


.github/workflows/geocoding-service-python.yml
```diff
@@ -1,17 +1,3 @@
...
-    branches: [main, '*-stable']
-    paths:
-    - '.github/workflows/geocoding-service-python.yml'
-    - 'geocoding/location-service/**'
-  pull_request:
-    paths:
-    - '.github/workflows/geocoding-service-python.yml'
-    - 'geocoding/location-service/**'
-  workflow_dispatch:
-    paths:
```
jim-sheldon, line 14: delete just this block and leave the above config intact?

abhidg:
> Is the resulting YAML [valid](https://yaml.org/spec/1.2.2/#21-collections)? It looks like declaring a key without a value. If not, we [should be able](https://github.com/github/feedback/discussions/15097) to use `{}`.

GitHub Actions seems to be fine with it, as is pyyaml (`yaml.safe_load("a:\n  b:\n  c:\n")`), which keeps the value at None.
> 
> Out of curiosity, where did you find that `workflow_dispatch` does not take paths or branches? I can't find it in the Github documentation, and I do see [one post](https://ardalis.com/github-actions-on-demand/) about it that links to code that has `branches` removed in a later commit.

Build failures, such as https://github.com/globaldothealth/list/actions/runs/2244984521

.github/workflows/geocoding-service-python.yml
abhidg: typo! thanks

jim-sheldon (approved):

PR #2690, opened 2022-04-29
Import sample data through API

iamleeg:
this means that the sample data is correctly age-bucketed.


data-serving/scripts/setup-db/import-sample-data.py
```diff
@@ -0,0 +1,113 @@
...
+    with open("../../samples/cases.json") as f:
+        sample_cases = json.load(f)
+        converted_cases = [convert_case(c) for c in sample_cases]
+        request_body = {
+            "cases": converted_cases
+        }
+        request_headers = {
+            'X-Api-Key': api_key
+        }
+        response = requests.post(batch_upsert_endpoint, json=request_body, headers=request_headers)
```
jim-sheldon, line 100: try/except?

data-serving/scripts/setup-db/import-sample-data.py
```diff
@@ -0,0 +1,113 @@
...
+
+def api_key_for_generated_curator(base_url: str) -> str:
+    """Create a user and retrieve their API key."""
+    register_user_endpoint = f'{base_url}/auth/register'
+    user = {
+        'name': 'G.h test data importer',
+        'email': f'robot_{time.time()}@global.health',
+        'roles': ['curator'],
+    }
+    response = requests.post(register_user_endpoint, json=user)
```
jim-sheldon, line 68: try/except?

data-serving/scripts/setup-db/import-sample-data.py
```diff
@@ -0,0 +1,113 @@
...
+            maybe_date = date_value
+            if isinstance(maybe_date, dict):
+                date_value = maybe_date.get('$numberLong')
+                date_to_use = datetime.utcfromtimestamp(int(date_value)/1000.0).strftime('%Y-%m-%dT%H:%M:%S.%f')
+            else:
+                date_to_use = maybe_date
+    return date_to_use
+
+
+def convert_case(case):
```
jim-sheldon, line 27: One of the downsides of Python...

jim-sheldon (approved):

abhidg (approved):

iamleeg:
There is one remaining Cypress test failure, I have spent over a day diagnosing it and am no closer, but it is in behaviour we don't use so I am arguing for changing the test expectation to match the new behaviour, merging this PR, then making a new issue to remove the unused functionality.

## Failure analysis

The test does manual CSV upload, then manual CSV upload of the same cases again but with the gender changed in one of two cases. It checks that only one case has been updated (the one with changed gender), but on this branch the test fails because it reports that both tests have been updated.

In fact, they have. The unmodified case has the same values that it did before, but gets attached to the new upload and gets new revision metadata. So there's no visible change, except that the case looks like it got updated (and the `caseRevisions` collection gets bigger).

As far as I can tell, the decision on whether a document needs updating or not is internal to mongodb's `batchWrite` function. The commit that introduced this failure is `d1a192d12f038f9fdfcf70510bbc39814c43d246`, which doesn't change the upsert logic, it introduces the use of bucketed age ranges when saving cases.

## Why I think we should ignore this failure

Firstly, we don't use manual CSV uploads any more, so we won't really observe any issues with incorrect revision metadata. Secondly, we don't use the revision metadata _at all_, and even in our plans to have frozen API requests we will not use it, so there is no incorrect behaviour even where revisions are recorded for unchanged cases.

So I'm actually suggesting not only ignoring this failure for now and merging the PR, but making another issue to completely remove the document revision pattern at a future date. This will greatly increase write performance (i.e. ingestion) without compromising the existing or planned snapshot behaviour (i.e. frozen requests, and daily downloads).

What do you think?

jim-sheldon (approved):

abhidg:
Agreed with @jim-sheldon, we should remove unused functionality. Document revisions could be useful for partner instances, but that’s a if, and the timeline for getting there is unknown, so we shouldn’t keep this around just in case.

PR #2691, opened 2022-05-04
Fix data-service CI failures

abhidg:
Data service CI was failing due to the notes removal change. After this,
setting the 'notes' field of a case document doesn't actually result in
a updated case, as the notes field is deleted by controllers, and it is
not returned from the data service APIs.

To simulate an updated document, this adds a pathogen key to a case.

### DIFF
--- a/data-serving/data-service/src/controllers/case.ts
+++ b/data-serving/data-service/src/controllers/case.ts
@@ -39,31 +39,46 @@ class InvalidParamError extends Error {}
 type BatchValidationErrors = { index: number; message: string }[];
 
 const caseFromDTO = async (receivedCase: CaseDTO) => {
-    const aCase = receivedCase as unknown as LeanDocument<CaseDocument>;
+    const aCase = (receivedCase as unknown) as LeanDocument<CaseDocument>;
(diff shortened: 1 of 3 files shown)


data-serving/data-service/src/controllers/case.ts
```diff
@@ -1099,7 +1118,9 @@ export const casesMatchingSearchQuery = (opts: {
 }): any => {
     // set data limit to 10K by default
     const countLimit = opts.limit ? opts.limit : 10000;
+    console.log(`Search query: ${opts.searchQuery}`);
```
iamleeg, line 1121: switch to `logger.debug`?

data-serving/data-service/test/controllers/case.test.ts
```diff
@@ -779,13 +786,15 @@ describe('POST', () => {
         const newCaseInDb = await Case.findOne({
             'revisionMetadata.revisionNumber': 0,
         });
+        console.log(newCaseInDb);
```
iamleeg, line 789: can be removed

iamleeg (approved):
Looks good, just need to tidy up some `console.log`

PR #2695, opened 2022-05-11
Monitoring fix

jim-sheldon:


iamleeg (approved):
Looks good. I'm wondering about removing the default values entirely and making people define these environment variables in, well, their environment, to aid future portability and to improve documentation of what the platform connects to. No need as a matter of urgency though, so please do merge this.

PR #2707, opened 2022-05-23
curator(api): allow deleting users

abhidg:
This PR adds API endpoints to delete users. It does not have the UI changes or the cleanup process which will assign cases created by them to a nobody or community user

### DIFF
--- a/verification/curator-service/api/openapi/openapi.yaml
+++ b/verification/curator-service/api/openapi/openapi.yaml
@@ -1056,6 +1056,21 @@ paths:
                     $ref: '#/components/responses/422'
                 '500':
                     $ref: '#/components/responses/500'
+        delete:
+            tags: [User]
+            summary: Deletes a user
+            operationId: deleteUser
+            responses:
+                '204':
+                    $ref: '#/components/responses/204'
+                '401':
+                    $ref: '#/components/responses/401'
+                '403':
+                    $ref: '#/components/responses/403'
+                '404':
+                    $ref: '#/components/responses/404'
+                '500':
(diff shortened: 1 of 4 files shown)


verification/curator-service/api/openapi/openapi.yaml
```diff
@@ -1056,6 +1056,17 @@ paths:
                     $ref: '#/components/responses/422'
                 '500':
                     $ref: '#/components/responses/500'
+        delete:
+            tags: [User]
+            summary: Deletes a user
+            operationId: deleteUser
+            responses:
+                '204':
```
iamleeg, line 1064: I've found that if you don't explicitly list 401/403 here then the openapi validator middleware turns that error into a 500 (because you're doing something outside what's documented).

verification/curator-service/api/test/users.test.ts
```diff
@@ -152,3 +152,29 @@ describe('PUT', () => {
...
+        await request
+            .post('/auth/register')
+            .send({ ...baseUser, ...{ roles: ['admin'] } })
+            .expect(200)
+            .expect('Content-Type', /json/);
+        return request
+            .delete('/api/users/5ea86423bae6982635d2e1f8')
+            .expect(404);
+    });
+});
```
iamleeg, line 180: two important tests: can it delete a user that isn't the requesting user, and does it correctly restrict a non-admin user from deleting a user?

iamleeg (changes requested):
A couple of tests need adding and some other clean up, I'm happy to add those changes if you're busy on other things.

iamleeg (approved):

PR #2711, opened 2022-05-30
Add country codes maps to the data dictionaries #2559

iamleeg:
Extra environment variables for this are already in terraform and applied, thanks @jim-sheldon!


data-serving/scripts/export-data/full_export.sh
```diff
@@ -26,6 +28,9 @@ trap 'rm -rf "$DIR"' EXIT  # Cleanup before exit
 
 cp data_dictionary.txt citation.txt "$DIR"
 
+echo "Getting country list from ${GDOTH_API_BASEURL}..."
+curl -H "X-API-Key: ${GDOTH_API_KEY}" "${GDOTH_API_BASEURL}/api/geocode/countryNames" > "${DIR}/countries.json"
```
abhidg, line 32: curl needs to be added to the Dockerfile_full_export as well

abhidg (changes requested):

abhidg (approved):

PR #2724, opened 2022-06-13
2712 api parameterisation

iamleeg:
Replaced the `server` parameter with a `disease` parameter, so instead of saying "the production covid-19 line list or you have to give us a full URL", we say "the production covid-19 line list or you ask for another disease".

I used different approaches in Python and R to accept this parameter: a class constructor in Python and changing the function argument in R. I don't get the impression that R programmers expect or use classes much, and I found three different implementations of classes which made me wonder if I might pick a "wrong" one that users don't expect to see.

I'm not fully satisfied with this solution as you now can't specify self-hosted or a dev/qa instance. But I couldn't work out a way to easily parameterise the disease name (which gives the "marketing" benefit that we advertise flexibility for different diseases) as well as the other fields. Perhaps we still let you set a custom URL in an env var? Would R users expect that as a customisation point?

### DIFF
--- a/api/R/R/cases_cachefile.R
+++ b/api/R/R/cases_cachefile.R
@@ -1,5 +1,5 @@
 cases_cachefile <-
-function(server = GDH_URL, folder="cache", ...) {
+function(server, folder="cache", ...) {
     return(paste0(folder, "/",
                   sha256(paste0(stringify_filters(...), "|", server)), ".csv"))
 }
--- a/api/R/R/constants.R
+++ b/api/R/R/constants.R
@@ -1,3 +1,3 @@
-GDH_URL <- "https://data.covid-19.global.health"
 FILTERS <- c("country")
+DISEASE <- "covid-19"
 downloadAsync <- "/api/cases/downloadAsync"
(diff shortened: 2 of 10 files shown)


abhidg:
Assuming that each disease is a separate instance, we could just ask people to specify the appropriate endpoint? That way they can specify (potentially) self-hosted, as well as dev/qa for monkeypox/covid-19.

Alternatively, to get dev/qa, that can be a separate flag to switch the URL variant

iamleeg:
> Assuming that each disease is a separate instance, we could just ask people to specify the appropriate endpoint?

Yes, though this is a bit harder to use. I suppose we could put that information on the Profile page alongside the API key ("this API key is for use with the following server name: data.covid-19.global.health").

> Alternatively, to get dev/qa, that can be a separate flag to switch the URL variant

Right, so disease=covid-19, env=production would be the main portal. That doesn't support custom URLs for self-hosted, but that's not a scenario we yet have so maybe doesn't need doing for now!

abhidg (approved):

PR #2732, opened 2022-06-17
2714 new data service

iamleeg:
Really early days but here is an end-to-end implementation of one API route to give a sense of the design and get early correction. Some notes:

 - there is exactly one data store at the moment (mongodb), but the service is designed to be flexible and to configure store at launch time
 - the data model has a base class for the day zero schema and a configuration point where additional fields can be supplied: I don't use that yet
 - no OpenAPI docs/conformance yet


.github/workflows/reusable-data-service-ci.yml
```diff
@@ -0,0 +1,39 @@
...
+        pip install flake8
+        # stop the build if there are Python syntax errors or undefined names
+        flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+        # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
+        flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+    - name: Run python tests
+      run: |
+        pip install poetry
+        poetry install
+        poetry update
```
abhidg, line 38: Any reason for update?

data-serving/reusable-data-service/reusable_data_service/model/case.py
```diff
@@ -0,0 +1,46 @@
...
+    information about an individual case for the line list.
+
+    Parameters here are defined to be keyword-only and not set in the
+    initialiser, so that clients can use Builder to populate them. Use
+    the validate() method to determine whether an instance is in a
+    consistent state (this also means we can add custom validation logic
+    to that function)."""
+
+    _: dataclasses.KW_ONLY
+    confirmation_date: datetime.datetime = dataclasses.field(init=False)
```
abhidg, line 19: datetime.date?

data-serving/reusable-data-service/reusable_data_service/model/case.py
```diff
@@ -0,0 +1,46 @@
...
+
+    @classmethod
+    def from_json(cls, obj: str) -> type:
+        """Create an instance of this class from a JSON representation."""
+        case = cls()
+        source = json.loads(obj)
+        for key in source:
+            if key in ["confirmation_date"]:
+                # parse as an ISO 8601 date
+                date = datetime.datetime.strptime(source[key], "%Y-%m-%dT%H:%M:%S.%fZ")
```
abhidg, line 29: confirmation_date should not have time, unless we want to support that

data-serving/reusable-data-service/tests/test_case_end_to_end.py
```diff
@@ -0,0 +1,51 @@
...
+
+from reusable_data_service import app, set_up_controllers
+
+
+@pytest.fixture
+def client_with_patched_mongo(monkeypatch):
+    # configure controllers
+    monkeypatch.setenv("DATA_STORAGE_BACKEND", "mongodb")
+    monkeypatch.setenv(
+        "MONGO_CONNECTION_STRING", "mongodb://localhost/27017"
```
abhidg, line 13: localhost:27017

data-serving/reusable-data-service/tests/test_case_end_to_end.py
```diff
@@ -0,0 +1,51 @@
...
+    monkeypatch.setattr("pymongo.MongoClient", fake_mongo)
+    set_up_controllers()
+    app.config["TESTING"] = True
+    client = app.test_client()
+    yield client
+
+
+def test_get_case_with_known_id(client_with_patched_mongo):
+    # insert a case
+    db = pymongo.MongoClient("mongodb://localhost/27017")
```
abhidg, line 31: localhost:27017

abhidg (commented):
Looks good, other than a typo. Should we use date instead of datetime? We don't have any time-related data in the current schema

data-serving/reusable-data-service/tests/test_case_end_to_end.py
iamleeg: In practice not a problem because the connection string is ignored in tests but I'll definitely correct it so as not to cause any further confusion :).

data-serving/reusable-data-service/reusable_data_service/model/case.py
iamleeg: Yup good point. I parse this as a date time still to handle the timezone correctly, then get a date from it.

data-serving/reusable-data-service/tests/test_case_end_to_end.py
```diff
@@ -0,0 +1,57 @@
...
+    client = app.test_client()
+    yield client
+
+
+def test_get_case_with_known_id(client_with_patched_mongo):
+    # insert a case
+    db = pymongo.MongoClient("mongodb://localhost:27017/outbreak")
+    case_id = (
+        db["outbreak"]["cases"]
+        .insert_one({"confirmation_date": "2021-12-31T01:23:45.678Z"})
```
abhidg, line 34: Will this convert to BSON date or be stored as a string?

iamleeg: Good catch, it can be stored as a date using `datetime.datetime`. I will update the test to ensure that works.

PR #2737, opened 2022-06-22
2714 post new case

iamleeg:
Creating cases. Open questions:

1. do we want case creation metadata on cases?
2. do we want case revision metadata on cases?


abhidg:
We are currently not using any of them, so I'd leave them out for now.

abhidg (approved):

PR #2739, opened 2022-06-27
2714 batch upsert

iamleeg:
This adds the batchUpsert endpoint, which is how bulk uploads and automatic ingestion are supported. As previously discussed with @abhidg this doesn't include curator or revision metadata, because they're unused in the Covid-19 version.

### DIFF
--- a/data-serving/reusable-data-service/reusable_data_service/controller/case_controller.py
+++ b/data-serving/reusable-data-service/reusable_data_service/controller/case_controller.py
@@ -1,6 +1,5 @@
 from flask import jsonify
 from datetime import date
-from hashlib import sha256
 from reusable_data_service.model.case import Case
 from reusable_data_service.model.filter import (
     Anything,
@@ -64,7 +63,7 @@ class CaseController:
         if num_cases <= 0:
             return "Must create a positive number of cases", 400
         try:
-            case = self.create_anonymised_case_if_valid(maybe_case)
+            case = self.create_case_if_valid(maybe_case)
             for i in range(num_cases):
                 self.store.insert_case(case)
             return "", 201
@@ -78,7 +77,7 @@ class CaseController:
     def validate_case_dictionary(self, maybe_case: dict):
         """Check whether a case _could_ be valid, without storing it if it is."""
         try:
-            case = self.create_anonymised_case_if_valid(maybe_case)
+            case = self.create_case_if_valid(maybe_case)
             return "", 204
         except ValueError as ve:
             # ValueError means we can't even turn this into a case
@@ -103,40 +102,27 @@ class CaseController:
         usable_cases = []
         for i, maybe_case in enumerate(cases):
             try:
-                case = self.create_anonymised_case_if_valid(maybe_case)
+                case = self.create_case_if_valid(maybe_case)
                 usable_cases.append(case)
             except Exception as e:
                 errors[str(i)] = e.args[0]
-        (created, updated) = self.store.batch_upsert(usable_cases) if len(usable_cases) > 0 else (0,0)
+        (created, updated) = (
+            self.store.batch_upsert(usable_cases) if len(usable_cases) > 0 else (0, 0)
+        )
         status = 200 if len(errors) == 0 else 207
         response = {"numCreated": created, "numUpdated": updated, "errors": errors}
         return jsonify(response), status
 
-    def create_anonymised_case_if_valid(self, maybe_case: dict):
+    def create_case_if_valid(self, maybe_case: dict):
         """Attempts to create a case from an input dictionary and validate it against
-        the application rules. Implements anonymisation rules so that cases cannot be
-        reidentified against external source data.
-        Raises ValueError or PreconditionError on invalid input."""
+        the application rules. Raises ValueError or PreconditionError on invalid input."""
         case = Case.from_dict(maybe_case)
         self.check_case_preconditions(case)
-        CaseController.anonymise_case(case)
         return case
 
     def check_case_preconditions(self, case: Case):
         if case.confirmationDate < self.outbreak_date:
             raise PreconditionError("Confirmation date is before outbreak began")
-
-    @staticmethod
-    def anonymise_case(case: Case):
-        """Ensure that a stable one-way hash of the source entry ID is stored, and
-        not the source entry ID itself."""
-        if (
-            case.caseReference is not None
-            and case.caseReference.sourceEntryId is not None
-        ):
-            case.caseReference.sourceEntryId = sha256(
-                case.caseReference.sourceEntryId.encode("utf-8")
-            ).hexdigest()
 
     @staticmethod
     def parse_filter(filter: str) -> Filter:
--- a/data-serving/reusable-data-service/reusable_data_service/controller/case_controller.py
+++ b/data-serving/reusable-data-service/reusable_data_service/controller/case_controller.py
@@ -86,6 +86,19 @@ class CaseController:
             # PreconditionError means it's a case, but not one we can use
             return pe.args[0], 422
 
+    def batch_upsert(self, body: dict):
+        """Upsert a collection of cases (updating ones that already exist, inserting
+        new cases). This method can potentially return a 207 mixed status as each case is
+        handled separately. The response will report the number of cases inserted, the
+        number updated, and any validation errors encountered."""
+        if body is None:
+            return "", 415
+        cases = body.get('cases')
+        if cases is None:
+            return "", 400
+        if len(cases) == 0:
+            return "", 400
+
     def create_case_if_valid(self, maybe_case: dict):
         """Attempts to create a case from an input dictionary and validate it against
         the application rules. Raises ValueError or PreconditionError on invalid input."""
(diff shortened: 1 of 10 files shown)


data-serving/reusable-data-service/reusable_data_service/controller/case_controller.py
```diff
@@ -86,17 +87,57 @@ def validate_case_dictionary(self, maybe_case: dict):
...
 
+    @staticmethod
+    def anonymise_case(case: Case):
+        """Ensure that a stable one-way hash of the source entry ID is stored, and
+        not the source entry ID itself."""
+        if (
+            case.caseReference is not None
+            and case.caseReference.sourceEntryId is not None
+        ):
+            case.caseReference.sourceEntryId = sha256(
```
abhidg, line 137: Should we add a salt here? Otherwise someone can sha256 all upstream source Ids and correlate

data-serving/reusable-data-service/reusable_data_service/stores/mongo_store.py
```diff
@@ -51,12 +51,22 @@ def count_cases(self, filter: Filter) -> int:
...
-        to_insert = case.to_dict()
-        for field in Case.date_fields():
-            # BSON works with datetimes, not dates
-            to_insert[field] = date_to_datetime(to_insert[field])
+        to_insert = MongoStore.case_to_bson_compatible_dict(case)
         self.get_case_collection().insert_one(to_insert)
 
+    def batch_upsert(self, cases: List[Case]) -> Tuple[int, int]:
+        dicts = [MongoStore.case_to_bson_compatible_dict(c) for c in cases]
+        needs_insert = lambda d: d['caseReference'] is None or d['caseReference']['sourceId'] is None or d['caseReference']['sourceEntryId'] is None
```
abhidg, line 59: Should sourceId ever be None?

abhidg (commented):
Looks good, other than comment on adding a salt

data-serving/reusable-data-service/reusable_data_service/controller/case_controller.py
iamleeg: yup good point

data-serving/reusable-data-service/reusable_data_service/stores/mongo_store.py
iamleeg: In the original data source the whole caseReference can be optional and the sourceId and sourceEntryId are both optional fields, should I make that stricter here?

data-serving/reusable-data-service/reusable_data_service/controller/case_controller.py
iamleeg: OK thinking about this it adds quite a bit of complexity that I'd like to discuss before committing to an implementation. It absolutely is the right thing to do, and I need to make sure I do it well :).

## Salt description

Each time we create a case, we generate some random salt, compute the hash `H(salt+sourceEntryId)`, then store `salt+H` as the sourceEntryId in the document. (Actually something like `"G.hv1:"+salt+H` so that we can easily identify whether an ID has been hashed, and can change the algorithm later when SHA256 is broken or too cheap.)

## Problem 1: UI

The curator retrieves a case from the line list, edits it, then saves back. The service should not re-hash the source entry ID, because then it will not be seen as the same case.

### Solution 1: `_id`

I think if a case already has an `_id` field we don't modify the `sourceEntryId`, on the basis that the document already came from the database so has been hashed already. If a case doesn't have an `_id` field, then its information is from an external source and should be anonymised.

This adds the constraint that even non-Mongo data stores need to use the `_id` field to store unique identifiers for cases, but that's an easy constraint to document and work with.

## Problem 2: batch upsert

A couple of different use cases rely on being able to upload a whole batch of cases, inserting ones that haven't been seen before and updating ones that have: bulk upload through the portal, and data ingestion. In that case, we need to ensure that no existing cases are duplicated.

### Solution 1: compare `sourceEntryId`

For each case in the input data set, retrieve the salt for the `sourceEntryId` in each document with the same `sourceId` and compute the hash for the incoming case using that salt. If they match, we have the same case, and should update rather than insert.

The problem with this approach is that it's O(n^2) in number of cases for a given sourceId (for every new case, compare against every existing case). The benefit is it implements both the existing upsert behaviour and the source entry ID anonymisation.

### Solution 2: always insert

Cases are always treated as if they will be added to the data set. We add a `dropExisting` parameter to the batch upsert endpoint, which actually becomes the batch _insert_ endpoint. If `dropExisting` is set, then the data service does a version of what we currently have with the `list` flag: it marks all existing cases for that source as to be deleted, inserts the new cases, then deletes the old ones. If it encounters a problem then it rolls back.

The ADI scripts for COVID-19 currently do batched uploads of 250 cases at once, to limit memory size and request size. As such, we couldn't _literally_ have a parameter on a `batchUpsert` request: the first 250 cases would remove everything that's already stored. So we would need to have something stateful (presumably using the database) that mimics the current ADI behaviour: perhaps a "begin replacement upload" and "end replacement upload" pair (with some lifecycle management to deal with ingestion failing in the middle of the process).

The problems with this approach are that it changes the API contract of the data service—there is no more real batch upsert—and that it requires complex state management. The benefits are that it internalises behaviour we already need for ADI, and avoids the expensive case-by-case hash comparisons.

@abhidg @jim-sheldon is there another solution here I'm missing? Preferences for how to deal with this?

data-serving/reusable-data-service/reusable_data_service/stores/mongo_store.py
abhidg: It makes sense for sourceId to not be None, in the current line list I can't recall instances where that is not the case.

data-serving/reusable-data-service/reusable_data_service/controller/case_controller.py
iamleeg: OK so let me write that design up again to make sure that I've got the implications right. Cases are stored across multiple collections, with one collection per source. Presumably named after the sourceId, but for legibility let's give one a name: Spanish cases are in 'ES'. Now when I want to do a bulk upload that will replace all cases:

1. `startReplacement?source=ES` creates a new collection `ES_new` (with all the required indexes) and the data service records the fact that if it sees any `batchInsert`s for ES, it uses the `ES_new` collection. It carries on using the `ES` collection for queries (e.g. list, download).
2. Repeatedly call `batchInsert` until all of the new cases are inserted. Hash the `sourceEntryId`s without reference to any existing data, because we assume that all cases are new.
3. `endReplacement?source=ES` drops the existing ES collection, renames `ES_new` to `ES`, with the effect that now the application sees all the new data.
4. `abortReplacement?source=ES` drops the _new_ collection, as well as the stateful indicator that the data-service should insert into that collection: everything goes back to using the existing collection.
5. We definitely need every case to have a `caseReference` with a `sourceId` because otherwise we don't know where to store it.


Is that what you mean @abhidg?

abhidg: Yes, similar but with slightly less state. Data service batchUpsert always creates a new collection like ES-timestamp and there is a update/refresh?source which picks the appropriate one (latest with more cases that has status == SUCCESS). Idea is no collection is preferred, there is only a current view [ES -> ES-20220623T150319, GB -> …] that can be rolled back.

iamleeg: But then we have to require that all batch upsert operations send a single request containing all the data. Currently ingestion sends smaller batches of 250 cases at a time, and sending the whole US Covid-19 dataset as a single request would be a very large transfer.

abhidg: Batch upsert should have the uploadId like now which it can use to put chunks in the same collection.

iamleeg: There's still something needed to tell the service that you've finished uploading for a particular uploadID, isn't there? So it still has to be stateful.

iamleeg: > Yes, I'm thinking it will be handled by parsing_lib as it is now after finalize_upload() https://github.com/globaldothealth/list/blob/main/ingestion/functions/common/parsing_lib.py#L517

I'm thinking that there should be some internally consistent behaviour for the data service to follow, for situations where there isn't automatic ingestion or where there is still partially manual curation. The day zero behaviour will be more like MPXV today than like COVID-19 today.

iamleeg: And in fact the bulk upload was used in COVID-19 to upload the spreadsheet data, which is a single "source" but contains data from multiple countries. So I think having one collection per upload would make for some very awkward queries.

iamleeg: Thanks for the call @abhidg, we now have a resolution I will document here and then implement.

1. All cases must have a source ID, and there will be no source entry ID (therefore no hashing, and also no correlation with upstream data).
2. Upsert looks for the existence of `_id` (currently the mongo object ID, but generically "an opaque identity token assigned by Global.health") on a case. If `_id` is set, then the case is a replacement; if not set, it is an insertion.
3. The curator service and ingestion library take care of building an atomic-looking data management platform on top of 1 and 2.

iamleeg:
Hi @abhidg I have implemented the changes discussed yesterday: upsert checks whether a case has an `_id` to decide whether to insert or replace, and `sourceEntryId` is no longer part of the `CaseReference` definition. This doesn't change the (individual) case creation endpoint, because a different endpoint is used for updating (creation is `POST /api/cases`; update is `PUT /api/cases/<id>`).

abhidg (approved):

Issue #2714, opened 2022-06-01
Configure form/list for monkeypox data schema


iamleeg:
See Moritz's day zero schema on https://github.com/globaldothealth/monkeypox/issues/21#issuecomment-1143321387

iamleeg:
Here's the plan @abhidg and I discussed yesterday.

![#2714 architecture](https://user-images.githubusercontent.com/237254/174253150-c5f79241-0c57-475e-8df9-6abfc5973c85.png)

So the curator API has a configurable connection to a data service (which it already does), and rather than having a strict OpenAPI validator for cases accepts any object as a case and lets the data service validate its input and report errors: the curator API is responsible for user management, data source/ingestion management and access control.

Existing UI has a lot of connection to new schema, but we can re-use automated source configuration, user profiles/management, overall structure so add new case list/forms for the new flexible schema and choose which to use per deployment.

Geocoding API remains unchanged.

The data API is completely replaced: the existing one has multiple constraints coupling it to the COVID-19 schema, so we'll have a new service for new diseases which encodes the day 0 schema plus custom fields. The covid-19 instances of G.h use the existing data service, and hMPXV/emerging outbreak instances use the new one instead.
New data service can be based on Python, and should not use a schema-defining library like Mongoose so that validation can be done in code to take into account custom fields defined by curators. It should use standard representations of all fields where possible, e.g. ISO country codes, ISO date formats, GeoJSON for locations. OpenAPI documentation should be derived from the code and not a separate, redundant statement of the data schema. Consider whether it is possible/desirable to switch database to postgres or appropriate to continue using mongoDB.

jim-sheldon:
Work on the curator UI captured in this issue: https://github.com/globaldothealth/list/issues/2902

PR #2742, opened 2022-07-05
#2714 download

iamleeg:
Download endpoint. Streams the results to avoid bloating memory; supports CSV, JSON, and TSV. I didn't support the free-text search from the existing data service because we don't expose it in the curator portal, and it is pretty expensive for us to run "at scale".


data-serving/reusable-data-service/reusable_data_service/controller/case_controller.py
```diff
@@ -109,20 +100,63 @@ def batch_upsert(self, body: dict):
...
+        separator_method = f"{format}_separator"
+        footer_method = f"{format}_footer"
+
+        if case_ids is not None:
+            case_iterator = self.store.identified_case_iterator(case_ids)
+            count = len(case_ids)
+        else:
+            predicate = CaseController.parse_filter(filter)
+            if predicate is None:
+                raise ValidationError("cannot understand query")
```
jim-sheldon, line 130: Add the predicate and filter to this exception message?

data-serving/reusable-data-service/reusable_data_service/main.py
```diff
@@ -23,21 +31,70 @@ def list_cases():
...
+        }
+        mime_type = content_types.get(req.get("format"), "text/plain")
+        generator = case_controller.download(
+            req.get("format"), req.get("query"), req.get("caseIds")
+        )
+        return app.response_class(generator(), mimetype=mime_type)
+    except PreconditionUnsatisfiedError as e:
+        return jsonify(""), 400
+    except UnsupportedTypeError as e:
+        return jsonify(""), 415
```
jim-sheldon, line 97: Add the name of the error and/or some context to the returned JSON?

data-serving/reusable-data-service/tests/test_case_controller.py
```diff
@@ -36,180 +43,276 @@ def batch_upsert(self, cases: List[Case]):
...
 def test_list_cases_should_list_the_cases(case_controller):
     with open("./tests/data/case.minimal.json", "r") as minimal_file:
         case = Case.from_json(minimal_file.read())
         case_controller.store.put_case("foo", case)
         case_controller.store.put_case("bar", case)
-    (response, status) = case_controller.list_cases()
-    assert status == 200
-    assert len(response.json["cases"]) == 2
+    cases = case_controller.list_cases()
+    len(cases.cases) == 2
```
jim-sheldon, line 89: assert?

jim-sheldon (approved):

data-serving/reusable-data-service/reusable_data_service/controller/case_controller.py
iamleeg: I didn't do it as the request already contains the query that wasn't understood (and the predicate is `None`), but can do.

PR #2744, opened 2022-07-06
#2714 batch status change

iamleeg:
Would appreciate a fresh pair of eyes on this; the e2e test `test_exclude_selected_cases` fails because the case doesn't actually get updated, and I can't work out why. From what I have seen, the correct case is replaced, the replace operation is acknowledged, and then the case still has its original values: I can't reproduce that in a standalone project so I'm sure the problem is mine rather than mongomock's. But I don't see it.

Thanks!

### DIFF
--- a/data-serving/reusable-data-service/data_service/model/case_exclusion_metadata.py
+++ b/data-serving/reusable-data-service/data_service/model/case_exclusion_metadata.py
@@ -1,6 +1,8 @@
 import dataclasses
 import datetime
 
+from typing import Any
+
 
 @dataclasses.dataclass
 class CaseExclusionMetadata:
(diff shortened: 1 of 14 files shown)


data-serving/reusable-data-service/data_service/model/case_exclusion_metadata.py
```diff
@@ -0,0 +1,34 @@
...
+        return True
+
+    @classmethod
+    def from_dict(cls, dictionary: dict[str, Any]) -> type:
+        """Create a CaseExclusionMetadata from a dictionary representation."""
+        exclusion = CaseExclusionMetadata()
+        exclusion.note = dictionary.get("note")
+        exclusion.date = cls.interpret_date(dictionary.get("date"))
+        if exclusion.date is None:
+            raise ValueError(f"date missing in CaseExclusion document {dict}")
```
abhidg, line 33: typo: dict -> dictionary

data-serving/reusable-data-service/data_service/model/document.py
```diff
@@ -0,0 +1,116 @@
+import dataclasses
+import datetime
+
+from data_service.util.json_encoder import JSONEncoder
+
+from typing import List
```
abhidg, line 6: not needed since 3.9, but can stay as well -- https://peps.python.org/pep-0585/

data-serving/reusable-data-service/data_service/model/document.py
```diff
@@ -0,0 +1,116 @@
...
+        """Create a line listing all of the fields in me and my member dataclasses."""
+        return sep.join(self.field_values()) + "\n"
+
+    def to_tsv(self) -> str:
+        """Generate a row in a CSV file representing myself."""
+        return self.delimiter_separated_values("\t")
+
+    def to_csv(self) -> str:
+        """Generate a row in a CSV file representing myself."""
+        return self.delimiter_separated_values(",")
```
abhidg, line 116: Does this handle commas in values?

abhidg (commented):
looks good, other than a typo

data-serving/reusable-data-service/data_service/model/document.py
iamleeg: Uh, no, not yet…that's a good case to add.

abhidg:
Sounds good to me!

PR #2750, opened 2022-07-12
#2621 changed email on sign in, sign up, request reset password link to low…

OskarKocjan:
Changes 1:
- email field is no longer case-sensitive on a sign in
- every email string changed to lower case in list/verification/curator-service/ui/src/redux/auth/thunk.ts


OskarKocjan:
Changes 3:
- removed earlier obj to additional user indexes
- created a new migration with new additional user indexes for email case-insensitive searching

maciej-zarzeczny (approved):
LGTM 👍

abhidg (approved):

PR #2752, opened 2022-07-12
2714 batch update query

iamleeg:
Just like the existing data service, I do this the easy way by getting the list of cases matching the query and calling through to the other batch update behaviour. It would be possible to make this more efficient using a mongo update_many() command.

### DIFF
--- a/data-serving/reusable-data-service/data_service/controller/case_controller.py
+++ b/data-serving/reusable-data-service/data_service/controller/case_controller.py
@@ -55,10 +55,12 @@ class CaseController:
         if limit <= 0:
             raise PreconditionUnsatisfiedError("limit must be >0")
         predicate = CaseController.parse_filter(filter)
+        print(f"predicate: {predicate}")
         if predicate is None:
             raise ValidationError("cannot understand query")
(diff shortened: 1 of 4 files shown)


data-serving/reusable-data-service/data_service/controller/case_controller.py
```diff
@@ -55,10 +55,12 @@ def list_cases(self, page: int = None, limit: int = None, filter: str = None):
         if limit <= 0:
             raise PreconditionUnsatisfiedError("limit must be >0")
         predicate = CaseController.parse_filter(filter)
+        print(f"predicate: {predicate}")
```
jim-sheldon, line 58: `logging.debug`?

jim-sheldon (approved):

abhidg (approved):

data-serving/reusable-data-service/data_service/controller/case_controller.py
iamleeg: Oops. Not even that: there's no logging anywhere in this yet, and I'll set that up universally later (ideally when I've worked out how to decorate the controller functions)

PR #2757, opened 2022-07-18
2714 custom fields

iamleeg:
This is the first part of extending the model in the data service with additional fields: adding those fields and storing descriptions in the DB. Still to come:

 - replay the stored fields on launch
 - watch the DB for new fields being added (so multiple instances of data-service can stay in sync)
 - integration tests to ensure that the case controller works with the updated fields correctly
 - remove fields
 - add fields that are lists of values
 - number types
 - validation
 - fields that are enumerations (with optional "other" free text entry)
 - indicate that a field is required (with a default value for migrating existing documents)
 - can GET the data dictionary


data-serving/reusable-data-service/data_service/model/case.py
```diff
@@ -86,6 +88,57 @@ def validate(self):
         self.caseReference.validate()
 
 
+observers = set()
```
abhidg, line 91: set() might not work with unhashable types and a generic function would be that

data-serving/reusable-data-service/data_service/main.py
```diff
@@ -150,8 +152,18 @@ def excluded_case_ids():
         return jsonify({"message": e.args[0]}), e.http_code
 
 
+@app.route("/api/schema", methods=["POST"])
+def add_field_to_case_schema():
+    try:
+        req = request.get_json()
+        schema_controller.add_field(req["name"], req["type"], req["description"])
+        return "", 201
```
jim-sheldon, line 160: `except ConflictError`  returns 409, `except PreconditionUnsatisfiedError` returns 400,
and `except KeyError` returns 400?

jim-sheldon (approved):

data-serving/reusable-data-service/data_service/main.py
iamleeg: so `KeyError` shouldn't happen: that will come from `Field.model_type` if you try to use a field type that isn't in the allowed list. So we turn it into `PreconditionUnsatisfiedError` which is the same as when we explicitly check that the type isn't found in the allowed list: i.e. both exception paths mean the same thing so they result in the same error (400 bad request: you've asked for something we aren't going to let you do).

`ConflictError` does indeed result in a 409 Conflict status.

PR #2774, opened 2022-07-20
#2623 Data acknowledgments are sorted case-insensitive

OskarKocjan:
Changes:
- Data acknowledgments are sorted case-insensitive


maciej-zarzeczny (approved):
Everything works well 👍

iamleeg (approved):

PR #2776, opened 2022-07-20
2714 required fields

iamleeg:
Users can indicate that a custom field is required: this will change the validation rules such that a case is invalid if it doesn't provide a value for the specified field. As such, users must also supply an initial default value for the field, which will be applied to existing cases.

### DIFF
--- a/data-serving/reusable-data-service/data_service/controller/schema_controller.py
+++ b/data-serving/reusable-data-service/data_service/controller/schema_controller.py
@@ -64,6 +64,4 @@ class SchemaController:
         field_model = Field(name, type_name, description, required, default)
         add_field_to_case_class(field_model)
         if store_field:
-            # create a storable model of the field and store it
-            # FIXME rewrite the validation logic above to use the data model
             self.store.add_field(field_model)


data-serving/reusable-data-service/data_service/model/case.py
```diff
@@ -94,13 +103,18 @@ def validate(self):
...
 
 
-def make_custom_case_class(name: str, fields=[]) -> type:
-    """Generate a class extending the DayZeroCase class with additional fields."""
+def make_custom_case_class(name: str, fields=[], field_models=[]) -> type:
+    """Generate a class extending the DayZeroCase class with additional fields.
+    fields is a list of dataclass fields that should be added to the generated class.
+    field_models is a list of model objects describing the fields for the data dictionary
+    and for validation."""
+    # FIXME generate the fields list from the field_models
```
iamleeg, line 111: my plan for this is when the day zero schema is committed, to create a list of `Field` objects that contain the day zero fields and get played into the `Case` class; then the logic for `make_custom_case_class` is simply to take a list of all known `Field`s and make data class fields for them. At that point I think the `DayZeroCase` intermediate class goes away.

jim-sheldon (approved):

PR #2786, opened 2022-07-27
#2714 fields with associated free other value

iamleeg:
If one of the enumerated values in a field is 'other', then you get (at no additional cost to the purchaser) a string field named fieldName_other where we can store the free text value.

### DIFF
--- a/data-serving/reusable-data-service/data_service/model/field.py
+++ b/data-serving/reusable-data-service/data_service/model/field.py
@@ -60,14 +60,21 @@ class Field(Document):
     def python_type(self) -> type:
         return self.model_type(self.type)
 
-    def dataclasses_tuple(self) -> (str, type, dataclasses.Field):
+    def dataclasses_tuples(self):
         # Note that the default value here is always None, even if I have a default value!
         # That's because the meaning of "required" in a field model is "a user _is required_ to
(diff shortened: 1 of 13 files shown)


data-serving/reusable-data-service/data_service/model/field.py
```diff
@@ -60,14 +60,21 @@ def from_dict(cls, dictionary):
     def python_type(self) -> type:
         return self.model_type(self.type)
 
-    def dataclasses_tuple(self) -> (str, type, dataclasses.Field):
+    def dataclasses_tuples(self):
```
jim-sheldon, line 63: `-> (str, type, dataclasses.Field)`?

jim-sheldon (approved):

data-serving/reusable-data-service/data_service/model/field.py
iamleeg: should be `-> List[(str, type, dataclasses.Field)]` but I couldn't get the type checker in Python 3.10 to accept that. Maybe a bug in the type checker? Anyway it wouldn't load the module so I removed the type hint.

abhidg: list[tuple[str, type, dataclasses.Field]] should work (from typing import List, Tuple not required anymore)

Issue #2697, opened 2022-05-18
Add Captcha to feedback form submissions


maciej-zarzeczny:
Captcha should be added to applications below:
- [x] Wordpress landing page
- [ ] Curator portal
- [ ] Map

In the curator portal and map we could also add dialog with all the necessary fields that would trigger email sending. Currently it opens external mail app.

joe-brilliant:
Ready to test, TBD if we want to test internally or engage Applause

PR #2792, opened 2022-08-08
#2697 captcha to feedback from submission

OskarKocjan:
Changes:
- added ReCaptcha for sign-in sign-up
- added /feedback endpoint to send emails from the curator portal instead of outside application
- added modal for sending email
- added YAML documentation for /feedback endpoint
- added success/failure alert to feedback


maciej-zarzeczny (changes requested):
Sending emails and recaptcha works well 👍 I left some comments regarding code. It would be good to add some feedback for the user whether the feedback was sent successfully or not. You can use `SnackbarAlert` component for this purpose. Also while testing I noticed that when there are some formik errors displayed in the feedback form after clicking "cancel" and opening the modal again the errors are still present. They should be reseted. 
Screen Shot 2022-08-09 at 12 30 56

verification/curator-service/ui/src/components/FeedbackEmailDialog/index.tsx
```diff
@@ -25,13 +25,27 @@ export interface FeedbackFormValues {
...
-    const classes = useStyles();
     const user = useAppSelector(selectUser);
 
     const maxSizeMessage = 800;
 
+    const [alertInformationOpen, setAlertInformationOpen] =
+        useState<boolean>(false);
+    const [alertInformationMessage, setAlertInformationMessage] =
+        useState<string>('');
+    const [alertInformationType, setAlertInformationType] = useState<
```
maciej-zarzeczny, line 38: In cases like this you can also add all those state variables under one state object. This way you later have to only call `setState` once providing all the values instead of three times

maciej-zarzeczny (approved):
LGTM ✅

verification/curator-service/ui/cypress/integration/components/LandingPage.spec.ts
```diff
@@ -218,7 +218,9 @@ describe('LandingPage', function () {
         cy.get('#password').type('tT$5aaaaak');
         cy.get('#passwordConfirmation').type('tT$5aaaaak');
         cy.get('#isAgreementChecked').check();
-        for (let i = 0; i < 5; i++) {
+        for (let i = 0; i < 7; i++) {
```
abhidg, line 221: why is the Sign-in button being clicked so many times?

verification/curator-service/ui/src/components/landing-page/SignInForm.tsx
```diff
@@ -71,7 +71,9 @@ interface SignInFormProps {
     setRegistrationScreenOn: (active: boolean) => void;
 }
 
-const RECAPTCHA_SITE_KEY = process.env.RECAPTCHA_SITE_KEY as string;
+const RECAPTCHA_SITE_KEY = window.Cypress
+    ? '6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI'
```
abhidg, line 75: is this supposed to be secret? not sure

verification/curator-service/ui/src/components/landing-page/SignUpForm.tsx
```diff
@@ -93,7 +92,9 @@ interface SignUpFormProps {
     setRegistrationScreenOn: (active: boolean) => void;
 }
 
-const RECAPTCHA_SITE_KEY = process.env.RECAPTCHA_SITE_KEY as string;
+const RECAPTCHA_SITE_KEY = window.Cypress
+    ? '6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI'
```
abhidg, line 96: same here

abhidg (commented):
Couple of possible secrets, otherwise looks good

verification/curator-service/ui/src/components/landing-page/SignInForm.tsx
OskarKocjan: Thank you for your feedback! That's a great question. It is a ReCaptchas site-key for testing purposes which is available for all developers: https://developers.google.com/recaptcha/docs/faq#id-like-to-run-automated-tests-with-recaptcha.-what-should-i-do Since every developer have access to it and it is the same for everyone I didn't set it as a secret.

verification/curator-service/ui/cypress/integration/components/LandingPage.spec.ts
OskarKocjan: It is clicked so many times to trigger the login/register limiter that will be used after 4 unsuccessful attempts. I will change the number of clicks to the previous commit which is 5 times.

PR #2801, opened 2022-09-07
Captcha bug fixes

OskarKocjan:
Changes:
- limiters now skip counting attempt when it is successful
- ReCaptcha badge is now invisible
- added information below the sign-in/up button that site is using ReCaptcha
- fixed bug that didn't block the action of a user that failed the ReCaptcha test
- moved part of code responsible for ReCaptcha from SignIn and SignUp form to LandingPage


maciej-zarzeczny (approved):
LGTM 👍

PR #2896, opened 2022-10-04
#2856 - 500 error while filtering fix

maciej-zarzeczny:
**Changes**
- Added missing collations to MongoDB queries (without correct collation no indexes were used which made the queries timeout)
- Added guards for undefined values coming from the DB


abhidg:
LGTM, could you make another PR for the cypress tests? Thanks!

PR #2903, opened 2022-10-12
2901 fix country export

abhidg:
- export(country-export): fix typo, output buckets as jsonArray
- export(transform): support ageRange in addition to ageBuckets


data-serving/scripts/prune-uploads/hooks/aggregate.py
```diff
@@ -26,7 +26,7 @@ def run(sources: list[dict[str, Any]], env: str, dry_run: bool = False):
         )
     except Exception as e:
         logging.exception("Error occurred when fetching job definitions")
-        return
+        return "*Failed* to start aggregation job"
```
jim-sheldon, line 29: `raise` here instead of `return`?

data-serving/scripts/prune-uploads/hooks/country_export.py
```diff
@@ -92,3 +94,4 @@ def run(sources: list[dict[str, Any]], env: str, dry_run: bool = False):
                 )
         except Exception as e:
             logging.exception(f"Error occurred while trying to submit {jobdef}")
+    return errors
```
jim-sheldon, line 97: I would also `raise` here instead of return the list of errors

data-serving/scripts/prune-uploads/prune_uploads.py
```diff
@@ -23,15 +23,16 @@
...
     return [str(x["_id"]) for x in xs]
 
 
-def notify(string, webhook_url):
-    if string:
-        response = requests.post(webhook_url, json={"text": string})
+def notify(string):
+    if string and PRUNE_UPLOADS_WEBHOOK_URL:
+        response = requests.post(PRUNE_UPLOADS_WEBHOOK_URL, json={"text": string})
         return response.status_code == 200
```
jim-sheldon, line 36: Here too, I would do the `try`/`except`/`raise` with `logging.exception`

jim-sheldon:
Made a couple suggestions for cleaner function signatures, LGTM!

jim-sheldon (approved):

Issue #2897, opened 2022-10-08
dev portal is not loading


abhidg:
After deployment from [d39aa37](https://github.com/globaldothealth/list/commit/d39aa37fb282da5ce481532ff9c722012e72ea39), the portal is not loading. Some cypress tests failed after merge, and loading the portal at dev-data gives the following errors in the console

```
Failed to load resource: net::ERR_CONNECTION_REFUSED
/auth/profile:1          Failed to load resource: the server responded with a status of 403 ()
react-dom.production.min.js:189 Error: Missing required parameters: sitekey
    at Qd.<anonymous> (recaptcha__en.js:50:509)
    at new Qd (recaptcha__en.js:470:101)
    at Xb.<anonymous> (recaptcha__en.js:53:189)
    at new Xb (recaptcha__en.js:668:185)
    at recaptcha__en.js:305:391
    at i.explicitRender (recaptcha.js:112:46)
    at i.componentDidUpdate (recaptcha.js:140:10)
    at wl (react-dom.production.min.js:261:316)
    at yl (react-dom.production.min.js:260:446)
    at vl (react-dom.production.min.js:259:431)
ps @ react-dom.production.min.js:189
react-dom.production.min.js:127 Uncaught Error: Missing required parameters: sitekey
    at Qd.<anonymous> (recaptcha__en.js:50:509)
    at new Qd (recaptcha__en.js:470:101)
    at Xb.<anonymous> (recaptcha__en.js:53:189)
    at new Xb (recaptcha__en.js:668:185)
    at recaptcha__en.js:305:391
    at i.explicitRender (recaptcha.js:112:46)
    at i.componentDidUpdate (recaptcha.js:140:10)
    at wl (react-dom.production.min.js:261:316)
    at yl (react-dom.production.min.js:260:446)
    at vl (react-dom.production.min.js:259:431)
hits-i.iubenda.com/write?db=hits1:1          Failed to load resource: net::ERR_CONNECTION_REFUSED
```

PR #2908, opened 2022-10-16
dev: add RECAPTCHA environment variables #2897

abhidg:
This fixes loading the portal when running locally, but does not affect the dev deployment, which is still broken. I checked the EKS deployment and it sets RECAPTCHA_SITE_KEY, which should be read by the curator service, but is not.

edit: Interestingly, if I change the REACT_APP_RECAPTCHA_SITE_KEY to RECAPTCHA_SITE_KEY everywhere, the portal does not work locally. Not sure why that is.


maciej-zarzeczny (approved):
The `REACT_APP` prefix before env variables makes them accessible through `process.env` inside React apps. It basically tells React that those variables should be publicly accessible inside the app

PR #2912, opened 2022-10-18
#2840 - age display bug

maciej-zarzeczny:
**Changes**
- Modified age parsing function
- Fixed other found bug: when user had filters set and refreshed the page there was a race condition on two queries to the DB one with the filters and one without. Because of that sometimes the results were filtered and sometimes not. This is now also fixed


abhidg:
@maciej-zarzeczny there is a failing test case, there needs to be a space on either side of the hyphen in the formatting, otherwise :+1:

maciej-zarzeczny:
@abhidg Thanks for the review, I've missed that while changing the function. It should be fixed now

abhidg (approved):

PR #2961, opened 2023-01-30
NZ governmental dataset changed url and format; parser updated to match.

jsbrittain:
Source dataset URL has also be updated through the curator UI (https://raw.githubusercontent.com/minhealthnz/nz-covid-data/main/cases/covid-case-counts.csv).


jsbrittain:
CI tests fail during flake8 config parsing. Appears to relate to this known issue: https://stackoverflow.com/questions/74558565/flake8-error-code-supplied-to-ignore-option-does-not-match-a-z1-30

